PR #3, opened 2021-11-05
Evolve Loop

ax3l:
Start a template for the evolve loop and push some particles.

Tasks:
- [x] Move `Push` implementation into its own `.cpp` file
- [x] Print out particle positions

### DIFF
--- a/src/particles/Push.cpp
+++ b/src/particles/Push.cpp
@@ -17,7 +17,7 @@ namespace impactx
         using namespace amrex::literals; // for _rt and _prt
 
         // loop over refinement levels
-        int const nLevel = 0;
+        int const nLevel = 1;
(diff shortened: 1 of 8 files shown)


src/particles/Push.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+
+namespace impactx
+{
+    void Push (ImpactXParticleContainer & pc)
+    {
+        using namespace amrex::literals; // for _rt and _prt
+
+        // loop over refinement levels
+        int const nLevel = 0;
```
ax3l:
```suggestion
        int const nLevel = 1;
```

cemitch99:
It looks good to me so far.  Thanks a lot.

ax3l:
Ups, you hit "close" not "merge" @cemitch99 :)

PR #8, opened 2021-11-09
.gitignore: ignore all build_* directories

cemitch99:
Added `build_*/` to the file `.gitignore`.
That change makes sure that our build directories on Cori, e.g., `build_cori/`, and Perlmutter, e.g., `build_perlmutter/`, are not tracked by `git`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 build/
-build_cori/
+build_*/
 Backtrace.*
 *.vth
 *.sw[opqrs]
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build/
+build_cori/
 Backtrace.*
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,4 +1,5 @@
 build/
+build_cori/
```
RobertRyne: I'm leaving a comment about the line that changed. Looks good!

RobertRyne (approved):

.gitignore
ax3l: Hi @cemitch99, let's also add `build_perlmutter/`.

A general way to write this could be: `build_*/`
```suggestion
build_*/
```

ax3l (approved):
That looks great, thanks! :sparkles:

PR #19, opened 2021-11-30
Resize grid based on beam extent

RemiLehe:
This resizes the box by the min and max of the particle container (+ an additional 10%) and then redistributes the particles, in preparation for the space charge calculation.

### DIFF
--- a/src/ImpactX.H
+++ b/src/ImpactX.H
@@ -84,6 +84,9 @@ namespace impactx
         //! Delete level data
         void ClearLevel (int lev) override;
 
+        // Resize the mesh, based on the extent of the bunch of particle
+        void ResizeMesh ();
+
         /** these are the physical/beam particles of the simulation */
         std::unique_ptr<ImpactXParticleContainer> m_particle_container;
 
--- a/src/ImpactX.H
+++ b/src/ImpactX.H
@@ -84,7 +84,7 @@ namespace impactx
         //! Delete level data
         void ClearLevel (int lev) override;
 
-        // Resize the mesh, based on the extent of the bunch of particle
+        //! Resize the mesh, based on the extent of the bunch of particle
         void ResizeMesh ();
 
         /** these are the physical/beam particles of the simulation */
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -101,12 +101,15 @@ namespace impactx
     */
     void ImpactX::ResizeMesh () {
         // Extract the mean and RMS size of the particle positions
-        amrex::ParticleReal x_mean, x_std, y_mean, y_std, z_mean, z_std;
-        mypc->MeanAndStdPositions(x_mean, x_std, y_mean, y_std, z_mean, z_std);
+        amrex::ParticleReal x_min, x_max, y_min, y_max, z_min, z_max;
+        mypc->MinAndMaxPositions(x_min, x_max, y_min, y_max, z_min, z_max);
         // Resize the domain size
+        // The box is expanded slightly beyond the min and max of particles.
+        // This controlled by the variable `frac` below.
+        const amrex::Real frac=0.1;
         amrex::RealBox rb(
-            {x_mean-3*x_std, y_mean-3*y_std, z_mean-3*z_std}, // Low bound
-            {x_mean+3*x_std, y_mean+3*y_std, z_mean+3*z_std}); // High bound
+            {x_min-frac*(x_max-x_min), y_min-frac*(y_max-y_min), z_min-frac*(z_max-z_min)}, // Low bound
+            {x_max+frac*(x_max-x_min), y_max+frac*(y_max-y_min), z_max+frac*(z_max-z_min)}); // High bound
         amrex::Geometry::ResetDefaultProbDomain(rb);
         for (int lev = 0; lev <= max_level; ++lev) {
             amrex::Geometry g = Geom(lev);
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -102,7 +102,7 @@ namespace impactx
     void ImpactX::ResizeMesh () {
         // Extract the mean and RMS size of the particle positions
         amrex::ParticleReal x_min, x_max, y_min, y_max, z_min, z_max;
-        mypc->MinAndMaxPositions(x_min, x_max, y_min, y_max, z_min, z_max);
+        m_particle_container->MinAndMaxPositions(x_min, x_max, y_min, y_max, z_min, z_max);
         // Resize the domain size
         // The box is expanded slightly beyond the min and max of particles.
         // This controlled by the variable `frac` below.
@@ -129,10 +129,10 @@ namespace impactx
 
             // Note: The following operation assume that
             // the particles are in x, y, z coordinates.
-            // Resize the mesh, based on `mypc` extent
+            // Resize the mesh, based on `m_particle_container` extent
             ResizeMesh();
             // Redistribute particles in the new mesh
-            mypc->Redistribute();
+            m_particle_container->Redistribute();
 
             // push all particles
             Push(*m_particle_container, m_lattice);
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -97,12 +97,9 @@ namespace impactx
         amrex::ignore_unused(lev);
     }
 
-    /** Resize the mesh, based on the extent of the bunch of particle
-    */
     void ImpactX::ResizeMesh () {
         // Extract the min and max of the particle positions
-        amrex::ParticleReal x_min, x_max, y_min, y_max, z_min, z_max;
-        m_particle_container->MinAndMaxPositions(x_min, x_max, y_min, y_max, z_min, z_max);
+        auto const [x_min, x_max, y_min, y_max, z_min, z_max] = m_particle_container->MinAndMaxPositions();
         // Resize the domain size
         // The box is expanded slightly beyond the min and max of particles.
         // This controlled by the variable `frac` below.
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -99,9 +99,16 @@ namespace impactx
 
     /** Resize the mesh, based on the extent of the bunch of particle
     */
-    //void ImpactX::ResizeMesh () {
-        // Get the particles' mean and RMS in each direction
-    //}
+    void ImpactX::ResizeMesh () {
+        // Extract the mean and RMS size of the particle positions
+        amrex::ParticleReal x_mean, x_std, y_mean, y_std, z_mean, z_std;
+        mypc->MeanAndStdPositions(x_mean, x_std, y_mean, y_std, z_mean, z_std);
+        // Resize the domain size
+        amrex::RealBox rb(
+            {x_mean-3*x_std, y_mean-3*y_std, z_mean-3*z_std}, // Low bound
+            {x_mean+3*x_std, y_mean+3*y_std, z_mean+3*z_std}); // High bound
+        amrex::Geometry::ResetDefaultProbDomain(rb);
+    }
 
     void ImpactX::evolve (int num_steps)
     {
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -100,7 +100,7 @@ namespace impactx
     /** Resize the mesh, based on the extent of the bunch of particle
     */
     void ImpactX::ResizeMesh () {
-        // Extract the mean and RMS size of the particle positions
+        // Extract the min and max of the particle positions
         amrex::ParticleReal x_min, x_max, y_min, y_max, z_min, z_max;
         m_particle_container->MinAndMaxPositions(x_min, x_max, y_min, y_max, z_min, z_max);
         // Resize the domain size
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -108,6 +108,11 @@ namespace impactx
             {x_mean-3*x_std, y_mean-3*y_std, z_mean-3*z_std}, // Low bound
             {x_mean+3*x_std, y_mean+3*y_std, z_mean+3*z_std}); // High bound
         amrex::Geometry::ResetDefaultProbDomain(rb);
+        for (int lev = 0; lev <= max_level; ++lev) {
+            amrex::Geometry g = Geom(lev);
+            g.ProbDomain(rb);
+            amrex::AmrMesh::SetGeometry(lev, g);
+        }
     }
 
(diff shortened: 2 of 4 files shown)


src/ImpactX.cpp
```diff
@@ -97,6 +97,19 @@ namespace impactx
...
+    /** Resize the mesh, based on the extent of the bunch of particle
+    */
+    void ImpactX::ResizeMesh () {
+        // Extract the mean and RMS size of the particle positions
+        amrex::ParticleReal x_mean, x_std, y_mean, y_std, z_mean, z_std;
+        mypc->MeanAndStdPositions(x_mean, x_std, y_mean, y_std, z_mean, z_std);
+        // Resize the domain size
+        amrex::RealBox rb(
+            {x_mean-3*x_std, y_mean-3*y_std, z_mean-3*z_std}, // Low bound
+            {x_mean+3*x_std, y_mean+3*y_std, z_mean+3*z_std}); // High bound
```
RemiLehe: As discussed with @qianglbl, this will actually be modified so as to use the min/max of the particle positions.

src/ImpactX.cpp
```diff
@@ -106,6 +124,13 @@ namespace impactx
             BL_PROFILE("ImpactX::evolve::step");
             amrex::Print() << " ++++ Starting step=" << step << "\n";
 
+            // Note: The following operation assume that
+            // the particles are in x, y, z coordinates.
+            // Resize the mesh, based on `mypc` extent
+            // ResizeMesh();
```
RemiLehe:
```suggestion
            ResizeMesh();
```

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -82,4 +82,53 @@ namespace impactx
...
+        amrex::ParticleReal& x_mean, amrex::ParticleReal& x_std,
+        amrex::ParticleReal& y_mean, amrex::ParticleReal& y_std,
+        amrex::ParticleReal& z_mean, amrex::ParticleReal& z_std )
+    {
+        amrex::ParticleReal sum_x, sum_x2, sum_y, sum_y2, sum_z, sum_z2, sum_w;
+
+        using PType = ImpactXParticleContainer::SuperParticleType;
+
+        amrex::ReduceOps<amrex::ReduceOpSum, amrex::ReduceOpSum, amrex::ReduceOpSum, amrex::ReduceOpSum, amrex::ReduceOpSum, amrex::ReduceOpSum, amrex::ReduceOpSum> reduce_ops;
+        auto r = amrex::ParticleReduce<amrex::ReduceData<amrex::ParticleReal, amrex::ParticleReal, amrex::ParticleReal, amrex::ParticleReal, amrex::ParticleReal, amrex::ParticleReal, amrex::ParticleReal>>(
```
RemiLehe: @ax3l @atmyers Do you know of a way to make these long lines look nicer?
(I could remove the `amrex::` everywhere, but I think that the line would still be quite long.)

ax3l: I think this location is one of the few places where using a macro is justified.

@atmyers does AMReX define an equivalent of `BOOST_PP_REPEAT` already?
https://www.boost.org/doc/libs/1_75_0/libs/preprocessor/doc/ref/repeat.html
https://stackoverflow.com/a/8556064/2719194

Alternatively, it might be possible to write an `amrex::make_ParticleReduce` factory to get rid of the repetition? That the same way that `std::unique_ptr` -> `std::make_unique` et al. avoid being so verbose.

atmyers: I think one could use template magic to write a function that, given a typename `T` and int `N`, generates an `amrex::ReduceOps` with `T` repeated `N` times. 

It looks like this PR no longer uses the tuple reduce however? Did you encounter a problem with this @RemiLehe?

RemiLehe: Yes, I changed that part of the code, but for physical reasons: last week, @qianglbl commented that, instead of resizing to the RMS, we should resize to the min and max. So the new code computes the min/max instead of the mean/std.

atmyers: I think we could still merge together all those reduction operations into one kernel launch using the tuple interface, however. `amrex::ReduceOps` can be a mixture of `ReduceOpMin` and `ReduceOpMax`.

RemiLehe: Ah, yes, ok. I will update the PR accordingly.

RemiLehe:
@atmyers I just updated the PR. Is that what you had in mind?

src/ImpactX.cpp
```diff
@@ -97,6 +97,27 @@ namespace impactx
         amrex::ignore_unused(lev);
     }
 
+    /** Resize the mesh, based on the extent of the bunch of particle
+    */
+    void ImpactX::ResizeMesh () {
+        // Extract the mean and RMS size of the particle positions
```
atmyers:
```suggestion
        // Extract the min and max of the particle positions
```

atmyers:
> @atmyers I just updated the PR. Is that what you had in mind?

Yes, that's what I meant. We could potentially explore some ways of reducing the verbosity of doing this, but it would require some changes / new features in AMReX.

atmyers (approved):

src/ImpactX.H
```diff
@@ -84,6 +84,9 @@ namespace impactx
         //! Delete level data
         void ClearLevel (int lev) override;
 
+        // Resize the mesh, based on the extent of the bunch of particle
```
ax3l: Make sure Doxygen can pick up the string:
```suggestion
        //! Resize the mesh, based on the extent of the bunch of particle
```

src/ImpactX.cpp
```diff
@@ -97,6 +97,27 @@ namespace impactx
         amrex::ignore_unused(lev);
     }
 
+    /** Resize the mesh, based on the extent of the bunch of particle
+    */
```
ax3l: This belongs in the header (API) docs:
```suggestion
```

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -82,4 +82,40 @@ namespace impactx
...
+            reduce_ops);
+
+        x_min = amrex::get<0>(r);
+        y_min = amrex::get<1>(r);
+        z_min = amrex::get<2>(r);
+        x_max = amrex::get<3>(r);
+        y_max = amrex::get<4>(r);
+        z_max = amrex::get<5>(r);
+
+        amrex::ParallelDescriptor::ReduceRealMin(x_min);
```
ax3l: I think we can do this in a single MPI call instead of a sequential chain of 6 of them.
Similar to:
https://github.com/ECP-WarpX/WarpX/blob/583c78e563f7a74e89fb49ff77c7a93c0baaad78/Source/Diagnostics/ReducedDiags/BeamRelevant.cpp#L324-L340

src/ImpactX.cpp
```diff
@@ -97,6 +97,27 @@ namespace impactx
...
+    void ImpactX::ResizeMesh () {
+        // Extract the min and max of the particle positions
+        amrex::ParticleReal x_min, x_max, y_min, y_max, z_min, z_max;
+        m_particle_container->MinAndMaxPositions(x_min, x_max, y_min, y_max, z_min, z_max);
```
ax3l: Get ready for C++17 nice-ness :heart_eyes: 
```suggestion
        auto const [x_min, x_max, y_min, y_max, z_min, z_max] = m_particle_container->MinAndMaxPositions();
```

Refs.:
- https://godbolt.org/z/G8r7ceKdj
- https://en.cppreference.com/w/cpp/language/structured_binding

src/particles/ImpactXParticleContainer.H
```diff
@@ -83,6 +83,14 @@ namespace impactx
                        amrex::Vector<amrex::ParticleReal> const & y,
                        amrex::Vector<amrex::ParticleReal> const & z);
 
+        /** Compute the min and max of the particle position in each dimension
+        */
+        void
+        MinAndMaxPositions (
+           amrex::ParticleReal& x_min, amrex::ParticleReal& x_max,
+           amrex::ParticleReal& y_min, amrex::ParticleReal& y_max,
+           amrex::ParticleReal& z_min, amrex::ParticleReal& z_max );
```
ax3l: Add `#include <tuple>` at the beginning of the file and then you can:

```suggestion
        *
        * @returns x_min, x_max, y_min, y_max, z_min, z_max
        */
        std::tuple<
            amrex::ParticleReal, amrex::ParticleReal,
            amrex::ParticleReal, amrex::ParticleReal,
            amrex::ParticleReal, amrex::ParticleReal>
        MinAndMaxPositions ();
```

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -82,4 +82,40 @@ namespace impactx
 //        Redistribute();
 
     }
+
+    void
+    ImpactXParticleContainer::MinAndMaxPositions (
+        amrex::ParticleReal& x_min, amrex::ParticleReal& x_max,
+        amrex::ParticleReal& y_min, amrex::ParticleReal& y_max,
+        amrex::ParticleReal& z_min, amrex::ParticleReal& z_max )
+    {
```
ax3l:
```suggestion
    std::tuple<
            amrex::ParticleReal, amrex::ParticleReal,
            amrex::ParticleReal, amrex::ParticleReal,
            amrex::ParticleReal, amrex::ParticleReal>
    ImpactXParticleContainer::MinAndMaxPositions ()
    {
```

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -82,4 +82,40 @@ namespace impactx
...
+        y_max = amrex::get<4>(r);
+        z_max = amrex::get<5>(r);
+
+        amrex::ParallelDescriptor::ReduceRealMin(x_min);
+        amrex::ParallelDescriptor::ReduceRealMin(y_min);
+        amrex::ParallelDescriptor::ReduceRealMin(z_min);
+        amrex::ParallelDescriptor::ReduceRealMax(x_max);
+        amrex::ParallelDescriptor::ReduceRealMax(y_max);
+        amrex::ParallelDescriptor::ReduceRealMax(z_max);
+    }
```
ax3l, line 122:
```suggestion
        return r;
    }
```
or
```suggestion
        return {x_min, x_max, y_min, y_max, z_min, z_max};
    }
```

ax3l (commented):
Thank you!

Two suggestions:
- only a single MPI collective instead of 6
- C++17 nice-ness in APIs :)

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -82,4 +82,41 @@ namespace impactx
...
+            [=] AMREX_GPU_DEVICE(const PType& p) noexcept -> amrex::GpuTuple<amrex::ParticleReal, amrex::ParticleReal, amrex::ParticleReal, amrex::ParticleReal, amrex::ParticleReal, amrex::ParticleReal>
+            {
+                amrex::ParticleReal x = p.pos(0);
+                amrex::ParticleReal y = p.pos(1);
+                amrex::ParticleReal z = p.pos(2);
+                return {x, y, z, x, y, z};
+            },
+            reduce_ops);
+
+        x_min = amrex::get<0>(r);
```
ax3l: Put this directly in a vector, as shown here:
Put those directly in a vector, as shown in the Reduce link below:

https://github.com/ECP-WarpX/WarpX/blob/583c78e563f7a74e89fb49ff77c7a93c0baaad78/Source/Diagnostics/ReducedDiags/BeamRelevant.cpp#L324-L340

src/particles/ImpactXParticleContainer.H
```diff
@@ -83,6 +83,16 @@ namespace impactx
                        amrex::Vector<amrex::ParticleReal> const & y,
                        amrex::Vector<amrex::ParticleReal> const & z);
 
+        /** Compute the min and max of the particle position in each dimension
+        *
+        * @returns x_min, x_max, y_min, y_max, z_min, z_max
+        */
+        std::tuple<
```
ax3l, line 92: `#include <tuple>` missing, see previous review comment :)

src/particles/ImpactXParticleContainer.cpp
ax3l: Still todo from previews review round :)

src/particles/ImpactXParticleContainer.cpp
ax3l: > I think one could use template magic to write a function that, given a typename T and int N, generates an amrex::ReduceOps with T repeated N times.

Yes, that's exactly the `make_...` C++ Factory pattern :)

src/particles/ImpactXParticleContainer.H
ax3l: ab6ce63cc711ecb1cdc4ba0115c374a63d33a6c4

ax3l (approved):
Thanks a lot - wuhu :rocket: :sparkles:

src/particles/ImpactXParticleContainer.H
ax3l: ab6ce63cc711ecb1cdc4ba0115c374a63d33a6c4

src/ImpactX.cpp
```diff
@@ -97,6 +97,24 @@ namespace impactx
...
+        // This controlled by the variable `frac` below.
+        const amrex::Real frac=0.1;
+        amrex::RealBox rb(
+            {x_min-frac*(x_max-x_min), y_min-frac*(y_max-y_min), z_min-frac*(z_max-z_min)}, // Low bound
+            {x_max+frac*(x_max-x_min), y_max+frac*(y_max-y_min), z_max+frac*(z_max-z_min)}); // High bound
+        amrex::Geometry::ResetDefaultProbDomain(rb);
+        for (int lev = 0; lev <= max_level; ++lev) {
+            amrex::Geometry g = Geom(lev);
+            g.ProbDomain(rb);
+            amrex::AmrMesh::SetGeometry(lev, g);
```
ax3l, line 114: @RemiLehe when compiling & running this in `Debug`, I get this error:
```
0::Assertion `hi > lo' failed, file "/home/axel/src/impactx/build/_deps/fetchedamrex-src/Src/Base/AMReX_Algorithm.H", line 107, Msg: "Error - calling bisect but lo and hi don't describe a reasonable interval." !!!
```

ax3l: opened as #44 to track :)

Disabled temporarily in #43 until we can fix it.

PR #21, opened 2021-12-04
Added declaration and initialization of a list of beamline elements.

cemitch99:
This is a first attempt at initializing a list of beamline elements defining the lattice.  For now, the list includes only drift elements.  The main work is done in the function initElements() in ImpactX.cpp.  

Request sanity check that this makes some sense before proceeding further.

### DIFF
--- a/src/ImpactX.H
+++ b/src/ImpactX.H
@@ -7,15 +7,14 @@
 #ifndef IMPACT_X_H
 #define IMPACT_X_H
 
-#include "particles/ImpactXParticleContainer.H"
-
 #include "particles/elements/Drift.H"
+#include "particles/ImpactXParticleContainer.H"
 
 #include <AMReX_AmrCore.H>
 
+#include <list>
 #include <memory>
 
-#include <list>
 
 namespace impactx
 {
--- a/src/ImpactX.H
+++ b/src/ImpactX.H
@@ -9,9 +9,13 @@
 
 #include "particles/ImpactXParticleContainer.H"
 
+#include "particles/elements/Drift.H"
+
 #include <AMReX_AmrCore.H>
 
 #include <memory>
+
+#include <list>
 
 namespace impactx
 {
@@ -49,6 +53,9 @@ namespace impactx
         /** Initialize particle containers and MultiFabs (fields) */
         void initData ();
 
+        /** Initialize the list of lattice elements */
+        void initElements ();
+
         /** Run the main simulation loop for a number of steps
          *
          * @param num_steps number of steps to evolve
@@ -80,6 +87,9 @@ namespace impactx
 
         /** these are the physical/beam particles of the simulation */
         std::unique_ptr<ImpactXParticleContainer> mypc;
+
+        /** these are elements defining the lattice */
+        std::list<Drift> myelements;
     };
 
 } // namespace impactx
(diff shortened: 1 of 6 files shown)


src/ImpactX.cpp
```diff
@@ -117,4 +117,18 @@ namespace impactx
         } // end step loop
     }
 
+    void ImpactX::initElements ()
+    {
+        // initialize element sequence
+        myelements = { };
+
```
ax3l: The list is empty by default: this constructor is called in `ImpactX::ImpactX` implicitly when the class that owns this member variable is constructed.
Thus, we can skip it here :)
```suggestion
```

src/ImpactX.cpp
```diff
@@ -117,4 +117,18 @@ namespace impactx
...
+        // add elements
+        Drift drift1(1.0);
+        Drift drift2(0.5);
+        myelements.push_back(drift1);
+        myelements.push_back(drift2);
```
ax3l: We can even make it more compact and avoid a copy by using:
```suggestion
        myelements.emplace_back(Drift(1.0));
        myelements.emplace_back(Drift(0.5));
```

src/ImpactX.cpp
```diff
@@ -117,4 +117,18 @@ namespace impactx
         } // end step loop
     }
 
+    void ImpactX::initElements ()
```
ax3l, line 120: This new function also needs to be declared in `ImpactX.H` :)

src/ImpactX.cpp
ax3l: Ah, you want to make sure they are cleared. That works :)

src/ImpactX.H
```diff
@@ -9,10 +9,14 @@
 
 #include "particles/ImpactXParticleContainer.H"
 
+#include "particles/elements/Drift.H"
+
 #include <AMReX_AmrCore.H>
 
 #include <memory>
 
+#include <list>
```
ax3l: Just grouping and sorting includes:
```suggestion
#include <list>
#include <memory>

```

src/ImpactX.H
```diff
@@ -9,10 +9,14 @@
 
 #include "particles/ImpactXParticleContainer.H"
 
+#include "particles/elements/Drift.H"
```
ax3l: Grouped and sorted:
```suggestion
#include "particles/elements/Drift.H"
#include "particles/ImpactXParticleContainer.H"
```

src/ImpactX.cpp
```diff
@@ -117,4 +117,18 @@ namespace impactx
...
+    {
+        // initialize element sequence
+        myelements = { };
```
ax3l: The list will be automatically constructed in `ImpactX::ImpactX` as an empty container.
Here, it's good to make sure the list is empty, e.g., if we want to re-initialize it.

`= {}` is semantically equivalent to `.clear()`, but the latter might be a bit more explicit:
```suggestion
        // make sure the element sequence is empty
        myelements.clear();
```

ax3l:
Work-around for CUDA: submitted as Nvidia request for Enhancement with BugID 3458976

Tiny demonstrator in https://cuda.godbolt.org/z/39e4q53Ye

PR #29, opened 2021-12-09
Use ABLASTR: WarpX Routines

ax3l:
Infrastructure:
- [x] w/ and w/o `-DImpactX_ablastr_src=...` works
- [x] add support for `-DImpactX_amrex_src=...`
- [x] ensure `-DImpactX_COMPUTE=CUDA` works
- [x] depends on https://github.com/ECP-WarpX/WarpX/pull/2263

Functionality:
- use charge deposition: https://github.com/ECP-WarpX/WarpX/pull/2652 / #30
- add support for pre-installed ABLASTR library

PR #30, opened 2021-12-10
Charge Deposition

ax3l:
- [x] create a charge density field
- [x] re-use `ablastr::DepositCharge`
- [x] call deposition in the PIC loop

Depends on:
- [x] #29
- [x] https://github.com/ECP-WarpX/WarpX/pull/2652
- [x] fix for resize bug #44: implement a test case with >=2 particles :) #47
- [x] #60
- [x] https://github.com/ECP-WarpX/WarpX/pull/2856 (#64)

### DIFF
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -65,8 +65,26 @@ namespace impactx
     void ImpactX::MakeNewLevelFromScratch (int lev, amrex::Real time, const amrex::BoxArray& ba,
                                           const amrex::DistributionMapping& dm)
     {
-        // todo data_mf.define(ba, dm, 1, 0);
-        amrex::ignore_unused(lev, time, ba, dm);
+        // todo m_rho.define(ba, dm, 1, 0);
+        amrex::ignore_unused(time, ba, dm);
+
+        // set human-readable tag for each MultiFab
+        auto const tag = [lev]( std::string tagname ) {
+            tagname.append("[l=").append(std::to_string(lev)).append("]");
+            return amrex::MFInfo().SetTag(std::move(tagname));
+        };
+
+        // charge (rho) mesh
+        amrex::BoxArray cba = ba;
+        // for MR levels:
+        //cba.coarsen(refRatio(lev - 1));
+
+        auto const rho_nodal_flag = amrex::IntVect::TheNodeVector(); // TODO: clarify if nodal or cell-centered
+        int const ncomp = 1;
+        int const ngRho = 3; // TODO: hard-coded, needs to be resized depending on shape
+
+        m_rho.emplace_back(std::make_unique<amrex::MultiFab>(
+            amrex::convert(cba, rho_nodal_flag), dm, ncomp, ngRho, tag("rho")));
     }
 
     /** Make a new level using provided BoxArray and DistributionMapping and fill
@@ -99,8 +117,7 @@ namespace impactx
      */
     void ImpactX::ClearLevel (int lev)
     {
-        // todo
-        amrex::ignore_unused(lev);
+        m_rho[lev].reset();
     }
 
     void ImpactX::ResizeMesh () {
@@ -150,7 +167,7 @@ namespace impactx
                 //m_particle_container->Redistribute();  // extra overload/arguments?
 
                 // charge deposition on level 0
-                //m_particle_container->DepositCharge(*m_rho.at(0));
+                m_particle_container->DepositCharge(*m_rho.at(0));
 
                 // poisson solve in x,y,z
                 //   TODO
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -79,11 +79,12 @@ namespace impactx
         //cba.coarsen(refRatio(lev - 1));
 
         auto const rho_nodal_flag = amrex::IntVect::TheNodeVector();
-        int const ncomp = 1;
-        int const ngRho = m_particle_container->GetParticleShape() + 1;
+        int const num_components_rho = 1;
+        // as in WarpX: this is likely too large and something like shape/2+1 would do.
+        int const num_guards_rho = m_particle_container->GetParticleShape() + 1;
 
         m_rho.emplace(lev,
-                      amrex::MultiFab{amrex::convert(cba, rho_nodal_flag), dm, ncomp, ngRho, tag("rho")});
+                      amrex::MultiFab{amrex::convert(cba, rho_nodal_flag), dm, num_components_rho, num_guards_rho, tag("rho")});
     }
 
     /** Make a new level using provided BoxArray and DistributionMapping and fill
@@ -165,8 +166,8 @@ namespace impactx
                 // Redistribute particles in the new mesh in x, y, z
                 //m_particle_container->Redistribute();  // extra overload/arguments?
 
-                // charge deposition on level 0
-                m_particle_container->DepositCharge(m_rho.at(0));
+                // charge deposition
+                m_particle_container->DepositCharge(m_rho, this->refRatio());
 
                 // poisson solve in x,y,z
                 //   TODO
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -65,7 +65,6 @@ namespace impactx
     void ImpactX::MakeNewLevelFromScratch (int lev, amrex::Real time, const amrex::BoxArray& ba,
                                           const amrex::DistributionMapping& dm)
     {
-        // todo m_rho.define(ba, dm, 1, 0);
         amrex::ignore_unused(time, ba, dm);
 
         // set human-readable tag for each MultiFab
@@ -76,15 +75,15 @@ namespace impactx
 
         // charge (rho) mesh
         amrex::BoxArray cba = ba;
-        // for MR levels:
+        // for MR levels (TODO):
         //cba.coarsen(refRatio(lev - 1));
 
-        auto const rho_nodal_flag = amrex::IntVect::TheNodeVector(); // TODO: clarify if nodal or cell-centered
+        auto const rho_nodal_flag = amrex::IntVect::TheNodeVector();
         int const ncomp = 1;
-        int const ngRho = 3; // TODO: hard-coded, needs to be resized depending on shape
+        int const ngRho = m_particle_container->GetParticleShape() + 1;
 
-        m_rho.emplace_back(std::make_unique<amrex::MultiFab>(
-            amrex::convert(cba, rho_nodal_flag), dm, ncomp, ngRho, tag("rho")));
+        m_rho.emplace(lev,
+                      amrex::MultiFab{amrex::convert(cba, rho_nodal_flag), dm, ncomp, ngRho, tag("rho")});
     }
 
     /** Make a new level using provided BoxArray and DistributionMapping and fill
@@ -117,7 +116,7 @@ namespace impactx
      */
     void ImpactX::ClearLevel (int lev)
     {
-        m_rho[lev].reset();
+        m_rho.erase(lev);
     }
 
     void ImpactX::ResizeMesh () {
@@ -167,7 +166,7 @@ namespace impactx
                 //m_particle_container->Redistribute();  // extra overload/arguments?
 
                 // charge deposition on level 0
-                m_particle_container->DepositCharge(*m_rho.at(0));
+                m_particle_container->DepositCharge(m_rho.at(0));
 
                 // poisson solve in x,y,z
                 //   TODO
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -80,8 +80,7 @@ namespace impactx
 
         auto const rho_nodal_flag = amrex::IntVect::TheNodeVector();
         int const num_components_rho = 1;
-        // as in WarpX: this is likely too large and something like shape/2+1 would do.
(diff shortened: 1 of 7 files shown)


src/particles/ImpactXParticleContainer.cpp
```diff
@@ -121,4 +125,78 @@ namespace impactx
...
+                //amrex::LayoutData<amrex::Real>* costs = WarpX::getCosts(lev);
+                //amrex::Real* cost = costs ? &((*costs)[pti.index()]) : nullptr;
+                amrex::Real *cost = nullptr;
+
+                // used for MR when we want to deposit for a subset of the particles on the level in the
+                // current box; with offset, we start at a later particle index
+                const long offset = 0;
+                const long np_to_depose = np;
+                int depos_lev = lev;
+
```
atmyers, line 175: If we'd like to support OpenMP threading for CPU execution, then the `#pragma omp parallel` region should go here, including the `local_rho_fab` as a thread private variable.

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -121,4 +125,78 @@ namespace impactx
...
+    void
+    DepositCharge (ImpactXParticleContainer & pc, amrex::MultiFab* rho, int icomp) // ImpactXParticleContainer & pc)
+    {
+        // loop over refinement levels
+        int const nLevel = 1;
+        for (int lev = 0; lev < nLevel; ++lev) {
+            // get simulation geometry information
+            //const amrex::Geometry& gm = this->Geom(lev);
+            //const auto prob_lo = gm.ProbLo();
+
```
atmyers: If we'd like to support OpenMP threading for CPU execution, then the #pragma omp parallel region should go here, including the local_rho_fab as a thread private variable.

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -121,4 +125,78 @@ namespace impactx
...
+                // number of components to deposit
+                const int nc = 1;
+
+                const int n_rz_azimuthal_modes = 0;
+                const long load_balance_costs_update_algo = 0;
+
+                // call amrex::Gpu::synchronize() for tiny profiler regions
+                const bool do_device_synchronize = true;
+
+                ablastr::DepositCharge<ImpactXParticleContainer>
```
ax3l: Now called:
```suggestion
                ablastr::particles::deposit_charge<ImpactXParticleContainer>
```

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -6,11 +6,15 @@
  */
 #include "ImpactXParticleContainer.H"
 
+#include <ABLASTR/DepositCharge.H>
```
ax3l: Now named:
```suggestion
#include <ablastr/particles/DepositCharge.H>
```

src/particles/ImpactXParticleContainer.cpp
ax3l: added :)

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -118,4 +121,85 @@ namespace impactx
...
+                    long const offset = 0;
+                    long const np_to_depose = np;
+                    int const depos_lev = lev;
+
+                    amrex::ParticleReal const charge = 1.0; // fixme
+
+                    auto const[nox, noy, noz] = std::array<int, 3>{2, 2, 2};
+                    auto const ng_rho = amrex::IntVect{nox + 1, noy + 1, noz + 1};
+
+                    std::array<amrex::Real, 3> const &dx = {1., 1., 1.};
```
ax3l: well, that's it :D

RemiLehe: Maybe this
```suggestion
                    std::array<amrex::Real, 3> const &dx = gm.CellSize();
```

RemiLehe: or 
```
std::array<amrex::Real, 3> const &dx = { gm.CellSize(0), gm.CellSize(1), gm.CellSize(2) };
```

ax3l: Thank you! Yes, the latter ! :)

src/ImpactX.cpp
```diff
@@ -65,8 +65,25 @@ namespace impactx
...
+            tagname.append("[l=").append(std::to_string(lev)).append("]");
+            return amrex::MFInfo().SetTag(std::move(tagname));
+        };
+
+        // charge (rho) mesh
+        amrex::BoxArray cba = ba;
+        // for MR levels (TODO):
+        //cba.coarsen(refRatio(lev - 1));
+
+        auto const rho_nodal_flag = amrex::IntVect::TheNodeVector();
```
ax3l, line 82: Reviewer: double check what we want for now as discretization:
[nodal or cell-centered](https://amrex-codes.github.io/amrex/docs_html/Basics.html#box)

RemiLehe: In order to re-use the WarpX/ablastr code for the Poisson solver, we need to have it nodal for now.

src/particles/ChargeDeposition.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+
+namespace impactx
+{
+    void
+    ImpactXParticleContainer::DepositCharge (amrex::MultiFab & rho, int icomp)
+    {
+        // loop over refinement levels
+        int const nLevel = this->maxLevel();
+        for (int lev = 0; lev <= nLevel; ++lev) {
```
RemiLehe, line 28: Here it seems that the particles from all levels deposit to the same `MultiFab` `rho`.

This is a bit different than the way we do things in WarpX, in which we:
- Deposit the particles from level 0 into `rho[lev=0]`, particles from level 1 into `rho[lev=1]`, etc.
- Then we use coarsening and `ParallelCopy` in order to propagate the charge deposited in `rho[lev=1]` down to `rho[lev=0]`
The reason we do it this way in WarpX is that the distribution mapping of `rho[lev=0]` and particles in level 0 is related, and the distribution mapping of `rho[lev=1]` and particles in level 1 is related too. This way we avoid particles trying to deposit into cells that are not owned by the local MPI rank.

src/particles/ImpactXParticleContainer.H
```diff
@@ -155,6 +160,17 @@ namespace impactx
...
+        /** Deposit the charge of the particles onto a grid
+         *
+         * This resets the values in rho to zero and then deposits the particle
+         * charge.
+         *
+         * @param rho charge grid to deposit on
+         * @param icomp component in rho to store into
+         */
+        void
+        DepositCharge (amrex::MultiFab & rho, int icomp = 0);
```
RemiLehe: I think that we can get rid of `icomp` here. In the context of ImpactZ, this will always be `0`, so no need to expose it as an option.

cemitch99 (approved):
This looks reasonable to me.  I'm not familiar with how this is implemented in WARP-X, so I can't follow-up on Remi's comment.

src/ImpactX.cpp
```diff
@@ -65,8 +65,25 @@ namespace impactx
...
+        int const ncomp = 1;
+        int const ngRho = m_particle_container->GetParticleShape() + 1;
```
ax3l: This might be too large but I think it might then also be too large in WarpX.

ax3l: Rename: `num_guards_rho`

src/particles/ChargeDeposition.cpp
```diff
@@ -0,0 +1,99 @@
...
+#endif
+            {
+                amrex::FArrayBox local_rho_fab;
+
+                using ParIt = ImpactXParticleContainer::iterator;
+                for (ParIt pti(*this, lev); pti.isValid(); ++pti) {
+                    const int np = pti.numParticles();
+
+                    // preparing access to particle data: SoA of Reals
+                    auto &soa_real = pti.GetStructOfArrays().GetRealData();
```
ax3l: restrict

src/particles/ChargeDeposition.cpp
```diff
@@ -0,0 +1,99 @@
...
+                    // after https://github.com/ECP-WarpX/WarpX/pull/2838 add const:
+                    auto const wp = soa_real[RealSoA::w];
+                    int const * const AMREX_RESTRICT ion_lev = nullptr;
+
+                    // mesh refinement ratio between lev and depos_lev (currently none)
+                    auto const ref_ratio = amrex::IntVect(AMREX_D_DECL(1, 1, 1));
```
ax3l: we can get this from the `ImpactX` (AMRCore) class.

src/particles/ChargeDeposition.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+                    // physical lower corner of the current box
+                    amrex::RealBox const grid_box{tilebox, gm.CellSize(), gm.ProbLo()};
+                    amrex::Real const * const xyzmin_ptr = grid_box.lo();
+                    std::array<amrex::Real, 3> const xyzmin = {xyzmin_ptr[0], xyzmin_ptr[1], xyzmin_ptr[2]};
+
+                    // pointer to costs data (currently unused)
+                    //amrex::LayoutData<amrex::Real>* costs = WarpX::getCosts(lev);
+                    //amrex::Real* cost = costs ? &((*costs)[pti.index()]) : nullptr;
+                    amrex::Real * const cost = nullptr;
```
ax3l: restrict

src/particles/ChargeDeposition.cpp
```diff
@@ -0,0 +1,99 @@
...
+                    // current box; with offset, we start at a later particle index
+                    long const offset = 0;
+                    long const np_to_depose = np;
+                    int const depos_lev = lev;
+
+                    amrex::ParticleReal const charge = 1.0; // TODO once we implement charge
+
+                    // particle shapes and needed deposition guards
+                    auto const[nox, noy, noz] = std::array<int, 3>{m_particle_shape, m_particle_shape, m_particle_shape};
+                    auto const ng_rho = amrex::IntVect{nox + 1, noy + 1, noz + 1};
```
ax3l: number of guards: see above (might be too large). Can be taken from m_rho as well

src/particles/ChargeDeposition.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+                    // cell size of the mesh to deposit to
+                    std::array<amrex::Real, 3> const &dx = {gm.CellSize(0), gm.CellSize(1), gm.CellSize(2)};
+
+                    // number of components to deposit
```
ax3l, line 79: actually the component number in m_rho

src/particles/ChargeDeposition.cpp
```diff
@@ -0,0 +1,99 @@
...
+                    int const nc = 1;
+
+                    // RZ modes (unused)
+                    int const n_rz_azimuthal_modes = 0;
+
+                    // call amrex::Gpu::synchronize() for tiny profiler regions
+                    bool const do_device_synchronize = true;
+
+                    ablastr::particles::deposit_charge<ImpactXParticleContainer>
```
ax3l, line 74: Overload in ablastr for sensible defaults/simplifications.

ax3l: https://github.com/ECP-WarpX/WarpX/pull/2856

src/ImpactX.cpp
```diff
@@ -65,8 +65,26 @@ namespace impactx
...
+
+        // charge (rho) mesh
+        amrex::BoxArray cba = ba;
+        // for MR levels (TODO):
+        //cba.coarsen(refRatio(lev - 1));
+
+        auto const rho_nodal_flag = amrex::IntVect::TheNodeVector();
+        int const num_components_rho = 1;
+        // as in WarpX: this is likely too large and something like shape/2+1 would do.
+        int const num_guards_rho = m_particle_container->GetParticleShape() + 1;
```
ax3l: Oh, forgot this one.

ax3l: I tried `shape/2 + 1` and even `10`.
Whatever I do, the lower bound of `algo.particle_shape = 3` seems to get out-of-bounds for `rho` at the moment.

ax3l: fixed xyzmin

src/particles/ChargeDeposition.cpp
```diff
@@ -0,0 +1,82 @@
...
+                    std::array<amrex::Real, 3> const xyzmin = {xyzmin_ptr[0], xyzmin_ptr[1], xyzmin_ptr[2]};
+
+                    // mesh-refinement: for when we do not deposit on the same level
+                    // note: would need to communicate the deposited-to boxes afterwards
+                    //int const depos_lev = lev;
+                    // mesh refinement ratio between lev and depos_lev
+                    //auto const rel_ref_ratio = ref_ratio.at(depos_lev) / ref_ratio.at(lev);
+                    amrex::ignore_unused(ref_ratio);
+
+                    amrex::ParticleReal const charge = 1.0; // TODO once we implement charge
```
ax3l, line 66: To Do for a follow-up PR:
We will set this when https://github.com/ECP-WarpX/impactx/pull/58 is merged and the unit system for charge is figured out.

We can merge the PR already, no need to wait for this constant atm (and we will likely normalize it to `q_e` anyway).

RemiLehe (approved):

PR #31, opened 2021-12-13
Limit the scope of profiler

WeiqunZhang:
In main.cpp, the destructor of the profiler was called after
amrex::Finalize.  This might cause an error in the future in SYCL if there
is a device synchronization call in the dtor, because the SYCL queues in
amrex have been deleted.  In this commit, we limit the scope of the profiler
so that its destructor is called before the queues are deleted.  Note that
it is never an issue for CUDA/HIP, because the device synchronization calls
in those backends do not need any amrex objects.  Furthermore, this is not
an issue currently for SYCL either.  But in the future, we might adopt the
WarpX approach of wrapping AMReX profilers and calling device
synchronization in the destructor.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -27,8 +27,9 @@ int main(int argc, char* argv[])
         impactx::overwrite_amrex_parser_defaults
     );
 
-    BL_PROFILE_VAR("main()", pmain);
     {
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Thanks a lot! :tada:

PR #32, opened 2021-12-14
Add function to compute mean and std

RemiLehe:
It looks like this function may be useful in the future, if we want to emulate an open-boundary Poisson solver by using AMReX Poisson solver with boundaries located ~10 sigmas away from the beam.


ax3l:
LGTM!

I think we can even share this in the ABLASTR lib directly, since it's universally useful:
https://github.com/ECP-WarpX/WarpX/pull/2675 / #33

ax3l:
Thank you @RemiLehe! :sparkles: 

Integrated upstream in ABLASTR: https://github.com/ECP-WarpX/WarpX/pull/2675
Merged into ImpactX via #33

PR #35, opened 2021-12-21
Max Level: From AMRinfo

ax3l:
The AMRcore (AMRinfo) and particle containers store the maximum level. Remove some hard-coding to the maximum level being zero for now.

This is just cleaning, no new functionality added.

### DIFF
--- a/src/particles/Push.cpp
+++ b/src/particles/Push.cpp
@@ -89,7 +89,7 @@ namespace detail
         using namespace amrex::literals; // for _rt and _prt
 
         // loop over refinement levels
-        int const nLevel = 1;
+        int const nLevel = pc.maxLevel();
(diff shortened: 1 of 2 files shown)


src/particles/Push.cpp
```diff
@@ -89,7 +89,7 @@ namespace detail
         using namespace amrex::literals; // for _rt and _prt
 
         // loop over refinement levels
-        int const nLevel = 1;
+        int const nLevel = pc.maxLevel();
```
ax3l, line 92: Ups, should have been `+1` or `<=`

ax3l: -> 20d1fec646d263fa8fe7373e0849e540690a4742

PR #45, opened 2022-01-05
Turn-off Poisson solver if there is only 1 macroparticle (or less)

### LINKED ISSUE #44: Handling of Single Particles / Perfectly Flat / Line Beams
While running in debug (#43), I realized the `ResizeMesh` function does assert the simulation:
https://github.com/ECP-WarpX/impactx/pull/19/files#r778820245
```
0::Assertion `hi > lo' failed, file "/home/axel/src/impactx/build/_deps/fetchedamrex-src/Src/Base/AMReX_Algorithm.H", line 107, Msg: "Error - calling bisect but lo and hi don't describe a reasonable interval." !!!
```

I temporarily disabled calling it and enabled debug builds in CI (#43). The function probably needs to be fixed.

RemiLehe:
~~We should not resize the mesh when no particle is present.~~

This PR disables the Poisson solver when only one particle is present.

Fix #44

### DIFF
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -139,8 +139,9 @@ namespace impactx
             // Note: The following operation assume that
             // the particles are in x, y, z coordinates.
             // Resize the mesh, based on `m_particle_container` extent
-            //   TODO: this is broken (fails in Debug mode)
-            //ResizeMesh();
+            if (m_particle_container.TotalNumberOfParticles(false,false) > 0) {
+                ResizeMesh();
+            }
 
             // Redistribute particles in the new mesh in x, y, z
             //m_particle_container->Redistribute();  // extra overload/arguments?
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -136,21 +136,24 @@ namespace impactx
                                                      transformation::Direction::T2Z,
                                                      pzd);
 
-            // Note: The following operation assume that
-            // the particles are in x, y, z coordinates.
-            // Resize the mesh, based on `m_particle_container` extent
-            if (m_particle_container->TotalNumberOfParticles(false,false) > 0) {
+            // Space-charge calculation: turn off if there is only 1 particle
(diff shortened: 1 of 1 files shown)


src/ImpactX.cpp
```diff
@@ -139,8 +139,9 @@ namespace impactx
             // Note: The following operation assume that
             // the particles are in x, y, z coordinates.
             // Resize the mesh, based on `m_particle_container` extent
-            //   TODO: this is broken (fails in Debug mode)
-            //ResizeMesh();
+            if (m_particle_container.TotalNumberOfParticles(false,false) > 0) {
```
RemiLehe:
```suggestion
            if (m_particle_container->TotalNumberOfParticles(false,false) > 0) {
```

RemiLehe:
Hm, the problem is actually that we have only 1 particle in this simulation.
In this case, the min and max position is equal, and therefore the resize fails.
@qianglbl In Impact, how do you resize the box when there is only 1 macroparticle in the beam? (granted: this is not a common case for production simulations, but it may be a useful case for testing.)

src/ImpactX.cpp
```diff
@@ -139,8 +139,9 @@ namespace impactx
             // Note: The following operation assume that
             // the particles are in x, y, z coordinates.
             // Resize the mesh, based on `m_particle_container` extent
-            //   TODO: this is broken (fails in Debug mode)
-            //ResizeMesh();
+            if (m_particle_container->TotalNumberOfParticles(false,false) > 0) {
```
```suggestion
            if (m_particle_container->TotalNumberOfParticles(false,false) > 1) {
```

ax3l:
@RemiLehe I think as you suggested, it makes sense to dynamically disable:
- grid scaling from `ResizeMesh`
- deposition
- Poisson solve
- space charge push

as soon as our particle container drops <2 particles.

All those components are not needed for a single particle sim :)

ax3l:
@RemiLehe just a small hint: can you please fork the repo for future PRs? :)

ax3l (approved):
Thank you :)

PR #48, opened 2022-01-18
Fix order in which min/max is extracted from ABLASTR function

RemiLehe:
The order in which the values are extracted need to match this line:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/ablastr/particles/ParticleMoments.H#L74

Seen in #47

### DIFF
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -104,7 +104,7 @@ namespace impactx
 
     void ImpactX::ResizeMesh () {
         // Extract the min and max of the particle positions
-        auto const [x_min, x_max, y_min, y_max, z_min, z_max] = m_particle_container->MinAndMaxPositions();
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Wuhu, thank you! :tada:

ax3l:
@RemiLehe I just missed this: can you please fix the documentation here?
https://github.com/ECP-WarpX/impactx/blob/fdfd115363e0c757c044bd7aca9ce4afb130b2db/src/particles/ImpactXParticleContainer.H#L100
and here,
https://github.com/ECP-WarpX/WarpX/blob/d3e17e7f7402ddc41272784cd5dd202eb6f7e4cd/Source/ablastr/particles/ParticleMoments.H#L28
too?

ax3l:
Update: fixes pushed to #52 and https://github.com/ECP-WarpX/WarpX/pull/2787

PR #58, opened 2022-02-08
Charge over mass

qianglbl:


src/particles/ImpactXParticleContainer.H
```diff
@@ -28,7 +28,6 @@ namespace impactx
         {
             ux,  ///< momentum in x, normalized to proper velocity
             uy,  ///< momentum in y, normalized to proper velocity
-            t,   ///< phase in arrival time * 2pi * reference frequency, radians
             pt,  ///< kinetic energy deviation, (E0-E)/mc^2
             q_m, ///< charge to mass ratio, in q_e/m_e (?)
```
ax3l, line 52: Let's update the unit that we use now:
```suggestion
            q_m, ///< charge to mass ratio, in q_e/eV
```

src/ImpactX.cpp
```diff
@@ -232,15 +232,25 @@ namespace impactx
...
         pp_dist.get("energy", energy);
 
-        amrex::Real bunch_charge = 0.0;  // Bunch charge (C)
+        amrex::ParticleReal bunch_charge = 0.0;  // Bunch charge (C)
         pp_dist.get("charge", bunch_charge);
 
         std::string particle_type;  // Particle type
         pp_dist.get("particle", particle_type);
 
+        amrex::ParticleReal qm = 0.0; //charge/mass ratio
```
ax3l, line 278:
```suggestion
        amrex::ParticleReal qm = 0.0;  // charge/mass ratio (q_e/eV)
```

src/particles/ImpactXParticleContainer.H
ax3l: Feel also free to rename this to `qm` if you prefer. You used `qm` as naming in all other locations :)

src/particles/ImpactXParticleContainer.H
```diff
@@ -94,6 +93,8 @@ namespace impactx
          * @param px momentum in x
          * @param py momentum in y
          * @param pz momentum in z
+         * @param qm charge over mass in 1/eV
+         * @param bchchg total charge within a bunch in C
```
ax3l, line 97: Isn't the unit in `q_e`? If it's C then I think we need to add a scaling factor in the weighting below?
```suggestion
         * @param bchchg total charge within a bunch in q_e
```

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -75,11 +77,8 @@ namespace impactx
         pinned_tile.push_back_real(RealSoA::ux, *px.cbegin(), *px.cend());
         pinned_tile.push_back_real(RealSoA::uy, *py.cbegin(), *py.cend());
         pinned_tile.push_back_real(RealSoA::pt, *pz.cbegin(), *pz.cend());
-
-        //the following should be updated
-        pinned_tile.push_back_real(RealSoA::t, np, 0.0);
-        pinned_tile.push_back_real(RealSoA::q_m, np, 0.0);
-        pinned_tile.push_back_real(RealSoA::w, np, 1.0/np);
+        pinned_tile.push_back_real(RealSoA::q_m, np, qm);
+        pinned_tile.push_back_real(RealSoA::w, np, bchchg/np);
```
ax3l: I think we might need to add a scaling factor from `C` to `q_e` here (since your `q_m` is in `q_e/eV`)?
```suggestion
        amrex::ParticleReal const q_e = 1.60217662e-19;
        pinned_tile.push_back_real(RealSoA::w, np, bchchg/q_e/np);
```

cemitch99: More basic question:  Where is a particle's q_m value actually needed?  In the existing implementation of elements, this information is not used.  Will this be used by the AMReX Poisson solver?  Also, we have available the particle type and the bunch charge.

ax3l: At least the charge will be used for deposition #30 :)
But there it is constant for the whole species (weighting or ionization state can vary for individual particles).

The AMReX solver should not need the q/m, we can scale before/after calling it.

cemitch99 (approved):
Ok, it looks good to me.

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -86,11 +88,9 @@ namespace impactx
...
         pinned_tile.push_back_real(RealSoA::uy, &(*py.cbegin()), &(*py.cend()));
         pinned_tile.push_back_real(RealSoA::pt, &(*pz.cbegin()), &(*pz.cend()));
-
-        //the following should be updated
-        pinned_tile.push_back_real(RealSoA::t, np, 0.0);
-        pinned_tile.push_back_real(RealSoA::q_m, np, 0.0);
-        pinned_tile.push_back_real(RealSoA::w, np, 1.0/np);
+        pinned_tile.push_back_real(RealSoA::q_m, np, qm);
+        amrex::ParticleReal const q_e = 1.60217662e-19;
+        pinned_tile.push_back_real(RealSoA::w, np, bchchg/q_e/np);
```
cemitch99, line 94: It looks like the weight w is now the number of physical particles per macroparticle, as we discussed during the meeting.  Looks ok to me.

cemitch99 (approved):
Looks good to me.

ax3l:
@qianglbl @cemitch99 I think there are a few unresolved comments of mine in this PR. Can you please address them in a follow-up PR?

ax3l:
I think it's these two:
- https://github.com/ECP-WarpX/impactx/pull/58#discussion_r808621781
- https://github.com/ECP-WarpX/impactx/pull/58#discussion_r808621647

PR #56, opened 2022-02-04
Modified Push and PushSingleParticle functions to access reference

cemitch99:
Modified the Push and PushSingleParticle functions to access the reference particle data.  

Next: reference particle will be added as an argument to each element "Drift.H", etc.

PR #59, opened 2022-02-08
Elements: Restrict Pointers & References

ax3l:
Explicitly declare that AoS and SoA arguments do not alias each other when passed into beamline elements. This allows the compiler to optimize more, as we guarantee that these are unique memory addresses.

- [x] rebase after #56 was merged

Minimal example:
- https://godbolt.org/z/vofejGsK8


cemitch99 (approved):
Looks good to me.

PR #62, opened 2022-02-11
Added function in subfolder "diagnostics" for printing particle data.

cemitch99:
Added a call to print final particle data in "evolve" loop after tracking, for benchmarking purposes.

### DIFF
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -188,7 +188,7 @@ namespace impactx
 
         // print final particle distribution to file
         diagnostics::DiagnosticOutput(*m_particle_container,
-                      diagnostics::OutputType::PrintParticles);
+                                      diagnostics::OutputType::PrintParticles);
 
     }
 
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -285,6 +285,9 @@ namespace impactx
               py.reserve(npart);
               pt.reserve(npart);
 
+              // write file header
+              amrex::PrintToFile("initial_beam.txt") << "#x y t px py pt\n";
+
               for(amrex::Long i = 0; i < npart; ++i) {
 
                   waterbag(ix, iy, it, ipx, ipy, ipt, rng);
@@ -296,7 +299,7 @@ namespace impactx
                   pt.push_back(ipt);
                   amrex::PrintToFile("initial_beam.txt") << ix << " " << iy << " ";
                   amrex::PrintToFile("initial_beam.txt") << it << " " << ipx << " ";
-                  amrex::PrintToFile("initial_beam.txt") << ipy << " " << ipt << " " << std::endl;
+                  amrex::PrintToFile("initial_beam.txt") << ipy << " " << ipt << "\n";
               }
           }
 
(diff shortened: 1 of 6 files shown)


src/particles/diagnostics/DiagnosticOutput.cpp
```diff
@@ -0,0 +1,63 @@
...
+                        amrex::ParticleReal const y = p.pos(1);
+                        amrex::ParticleReal const t = p.pos(2);
+
+                        // access SoA Real data
+                        amrex::ParticleReal &px = part_px[i];
+                        amrex::ParticleReal &py = part_py[i];
+                        amrex::ParticleReal &pt = part_pt[i];
+
+                        // write particle data to file
+                        amrex::PrintToFile("output_beam.txt") << x << " " << y << " ";
```
ax3l, line 67: Managed memory: copy from GPU to CPU
https://github.com/ECP-WarpX/WarpX/blob/fe971e82a49a1c5101e5ef48cce849460dcb598d/Source/Diagnostics/ReducedDiags/FieldProbe.cpp#L515-L531

The latter array can then be printed to file after that loop.

We can also do this, that might be compact:
https://github.com/ECP-WarpX/impactx/blob/e509f2a6d26e8596f5c6ea5b41286b67fc87cbae/src/particles/Push.cpp#L141-L155

ax3l: Later we do explicit copies to avoid managed memory: https://github.com/AMReX-Codes/amrex/blob/6e27136f76280e565d30ff654031d1b1ac683171/Src/Particle/AMReX_ParticleIO.H#L608

ax3l: if we want to do this properly, there is an easy way that I forgot about!
```C++
// create a host-side particle buffer
auto tmp = amrex::ParticleBuffer::getTmpPC<amrex::PinnedArenaAllocator>(pc);

// copy device-to-host
using SrcData = WarpXParticleContainer::ParticleTileType::ConstParticleTileDataType;
bool const local = true;
tmp.copyParticles(*pc, local);
```

The object in `tmp` behave just like our particle container `pc` and we can plot it with a for loop on the host as above.

src/ImpactX.cpp
```diff
@@ -279,6 +285,9 @@ namespace impactx
               py.reserve(npart);
               pt.reserve(npart);
 
+              // write file header
+              amrex::PrintToFile("initial_beam.txt") << "#x y t px py pt\n";
```
ax3l, line 289: @WeiqunZhang how do people handle multiple outputs between runs. I think I saw a standard renaming scheme that we use to move old files out?

Currently, this will append to a previous run's output.

WeiqunZhang: Old files can be removed with `amrex::FileSystem::Remove(std::string const& filename)`.

cemitch99 (commented):
The suggested changes look good to me.

src/ImpactX.cpp
ax3l: Thx! Ok that looks a bit tricky, because we suffix the files in `amrex::PrintToFile` automatically with MPI rank and OMP thread. Removing the current MPI+OMP thread in a follow-up can be confusing, if one restarts with a different MPI size or thread width.

ax3l: Found a solution: #65 :)

PR #65, opened 2022-02-17
Move Diags Between Runs

ax3l:
Create diagnostics in `diags/`.
Move when starting a new run to a backup name.

This avoids that we simply append to a previous run, which would be confusing during debugging and benchmarking.

Follow-up to #62

### DIFF
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -301,9 +301,9 @@ namespace impactx
                   px.push_back(ipx);
                   py.push_back(ipy);
                   pt.push_back(ipt);
-                  amrex::PrintToFile("diags/initial_beam.txt") << ix << " " << iy << " ";
-                  amrex::PrintToFile("diags/initial_beam.txt") << it << " " << ipx << " ";
-                  amrex::PrintToFile("diags/initial_beam.txt") << ipy << " " << ipt << "\n";
+                  amrex::PrintToFile("diags/initial_beam.txt")
+                      << ix << " " << iy << " " << it << " "
+                      << ipx << " " << ipy << " " << ipt << "\n";
               }
           }
 
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -15,6 +15,7 @@
 #include <AMReX_REAL.H>
 #include <AMReX_ParmParse.H>
 #include <AMReX_Print.H>
+#include <AMReX_Utility.H>
 
 #include <string>
 #include <vector>
@@ -42,6 +43,9 @@ namespace impactx
     {
         AmrCore::InitFromScratch(0.0);
         amrex::Print() << "boxArray(0) " << boxArray(0) << std::endl;
+
+        // move old diagnostics out of the way
+        amrex::UtilCreateCleanDirectory("diags", true);
 
         this->initDist();
         amrex::Print() << "# of particles: " << m_particle_container->TotalNumberOfParticles() << std::endl;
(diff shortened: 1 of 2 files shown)


src/ImpactX.cpp
```diff
@@ -297,9 +301,9 @@ namespace impactx
                   px.push_back(ipx);
                   py.push_back(ipy);
                   pt.push_back(ipt);
-                  amrex::PrintToFile("initial_beam.txt") << ix << " " << iy << " ";
-                  amrex::PrintToFile("initial_beam.txt") << it << " " << ipx << " ";
-                  amrex::PrintToFile("initial_beam.txt") << ipy << " " << ipt << "\n";
+                  amrex::PrintToFile("diags/initial_beam.txt") << ix << " " << iy << " ";
```
WeiqunZhang: It would be more efficient to merge the three calls to PrintToFile into one.

src/particles/diagnostics/DiagnosticOutput.cpp
```diff
@@ -64,9 +64,9 @@ namespace impactx::diagnostics
                         amrex::ParticleReal const pt = part_pt[i];
 
                         // write particle data to file
-                        amrex::PrintToFile("output_beam.txt") << x << " " << y << " ";
-                        amrex::PrintToFile("output_beam.txt") << t << " " << px << " ";
-                        amrex::PrintToFile("output_beam.txt") << py << " " << pt << "\n";
+                        amrex::PrintToFile("diags/output_beam.txt") << x << " " << y << " ";
```
WeiqunZhang: It would be more efficient to merge the three calls to PrintToFile into one.

src/ImpactX.cpp
cemitch99: Yes, that makes sense.  Line 304 of course needs to be continued to include it, ipx, ipy, and ipt.

src/particles/diagnostics/DiagnosticOutput.cpp
cemitch99: es, that makes sense.  Line 67 of course needs to be continued to include t, px, py, and pt.

src/ImpactX.cpp
```diff
@@ -297,9 +301,9 @@ namespace impactx
...
+                  amrex::PrintToFile("diags/initial_beam.txt") << it << " " << ipx << " ";
+                  amrex::PrintToFile("diags/initial_beam.txt") << ipy << " " << ipt << "\n";
```
```suggestion
                  amrex::PrintToFile("diags/initial_beam.txt")
                      << ix << " " << iy << " " << it
                      << " " << ipx << " " << ipy << " " << ipt << "\n";
```

src/particles/diagnostics/DiagnosticOutput.cpp
```diff
@@ -64,9 +64,9 @@ namespace impactx::diagnostics
...
+                        amrex::PrintToFile("diags/output_beam.txt") << t << " " << px << " ";
+                        amrex::PrintToFile("diags/output_beam.txt") << py << " " << pt << "\n";
```
ax3l:
```suggestion
                        amrex::PrintToFile("diags/output_beam.txt")
                            << x << " " << y << " "<< t << " "
                            << px << " " << py << " " << pt << "\n";
```

ax3l (commented):
Oh, good catch!

src/ImpactX.cpp
ax3l: Thanks, good catch! :)

cemitch99 (approved):
Looks good to me.

PR #70, opened 2022-02-26
Moved reference particle access inside "CoordinateTransformation".

cemitch99:
This removes hard-coded values for ptd, pzd that were previously used.


ax3l (approved):
Thank you, this looks great to me :+1: :)

PR #73, opened 2022-03-01
Parallelize particle initialization

### LINKED ISSUE #68: Parallel Particle Init
We need to improve the particle initialization: to allow having a first few boxes initially, initialize in parallel & redistribute.

https://github.com/ECP-WarpX/impactx/blob/4b7fa17b1404fb6d21ab38636114befd9722bbf6/src/particles/ImpactXParticleContainer.cpp#L103

Currently, we will be limited to initializing as many particles as a single GPU can handle (even if we run on multiple GPUs).

atmyers:
Each takes a portion of the `npart` particles, Redistribute is called after adding them to the container.

- [x] depends on https://github.com/AMReX-Codes/amrex/pull/2695

Close #68


src/particles/ImpactXParticleContainer.cpp
```diff
@@ -100,8 +102,8 @@ namespace impactx
         particle_tile.resize(new_np);
         amrex::copyParticles(
                 particle_tile, pinned_tile, 0, old_np, pinned_tile.numParticles());
-//        Redistribute(); // TODO
 
+        Redistribute();
```
ax3l: One thing why we commented this out here for now:

I think we have not set up our box yet to a reasonable extent. The box extent is determined from the positions of the min/max particle positions in 3D.

If we redistribute now, we probably have no good map to distribute into yet and might even remove particles that are out of bounds?

Maybe we can discuss with @qianglbl @cemitch99 @RemiLehe how to solve this chicken-egg problem in a clever way.

ax3l:
We discussed this in our developer meeting last week.

It would be good to solve the "box extent" chicken-egg problem by:
- optionally asking the user for an initial bounding box (if we initialize in parallel from arbitrary distributions)

Later on, in case we load beams from a data file, we can ask the data file about its min/max values.

src/particles/ImpactXParticleContainer.cpp
ax3l: We discussed this in our developer meeting last week.

It would be good to solve the "box extent" chicken-egg problem by:
- optionally asking the user for an initial bounding box (if we initialize in parallel from arbitrary distributions)

Later on, in case we load beams from a data file, we can ask the data file about its min/max values.

atmyers: In the case that the user does not provide a bounding box, should we track the min / max of the particle positions as they are generated, and reduce over the ranks?

ax3l: We currently do that after they are created, as part of the PIC loop:
https://github.com/ECP-WarpX/impactx/blob/570abd95e167f03479826a0bd21ecbaf509fb555/src/ImpactX.cpp#L172-L173
https://github.com/ECP-WarpX/impactx/blob/570abd95e167f03479826a0bd21ecbaf509fb555/src/ImpactX.cpp#L135-L151

Do you suggest we could also do it *while* we initialize them, then resize, then do the first redistribute? :)

ax3l: We currently do that after they are created, as part of the PIC loop:
https://github.com/ECP-WarpX/impactx/blob/570abd95e167f03479826a0bd21ecbaf509fb555/src/ImpactX.cpp#L172-L173
https://github.com/ECP-WarpX/impactx/blob/570abd95e167f03479826a0bd21ecbaf509fb555/src/ImpactX.cpp#L135-L151

Do you suggest we could also do it *while* we initialize them?

atmyers: Yes, that's what I mean - I'll update this PR to do so.

src/ImpactX.cpp
```diff
@@ -309,30 +309,34 @@ namespace impactx
...
-                  pt.push_back(ipt);
-                  amrex::PrintToFile("diags/initial_beam.txt")
-                      << ix << " " << iy << " " << it << " "
-                      << ipx << " " << ipy << " " << ipt << "\n";
-              }
+          int myproc = amrex::ParallelDescriptor::MyProc();
+          int nprocs = amrex::ParallelDescriptor::NProcs();
+          int navg = npart / nprocs;
+          int nleft = npart - navg * nprocs;
+          int npart_this_proc = (myproc < nleft) ? navg+1 : navg;
```
ax3l: We move this part in #107 in its own file.

ax3l: Logic simplified in #108, so there is now only one location to add this parallelism to :)

atmyers: I think I've successfully integrated those changes onto this branch.

atmyers:
It looks like the parallel initialization is working correctly when I check it locally, however, the particle diagnostics currently only print the particles assigned to rank 0. We could use `PC::WriteAsciiFile` in amrex to handle this for the parallel case.

ax3l:
Good point, added in #109

src/ImpactX.cpp
```diff
@@ -102,6 +102,5 @@ namespace impactx
         diagnostics::DiagnosticOutput(*m_particle_container,
                                       diagnostics::OutputType::PrintParticles,
                                       "diags/output_beam.txt");
```
ax3l: Removing unrelated change
```suggestion
                                      "diags/output_beam.txt");

```

src/initialization/InitDistribution.cpp
```diff
@@ -45,23 +45,27 @@ namespace
...
-            for(amrex::Long i = 0; i < npart; ++i) {
-                distr(ix, iy, it, ipx, ipy, ipt, rng);
-                x.push_back(ix);
-                y.push_back(iy);
-                t.push_back(it);
-                px.push_back(ipx);
-                py.push_back(ipy);
-                pt.push_back(ipt);
-            }
+        int myproc = amrex::ParallelDescriptor::MyProc();
```
ax3l, line 52:
```suggestion
        // Logic: We initialize 1/Nth of particles, independent of their
        // position, per MPI rank. We then measure the distribution's spatial
        // extent, create a grid, resize it to fit the beam, and then
        // redistribute particles so that they reside on the correct MPI rank.
        int myproc = amrex::ParallelDescriptor::MyProc();
```

src/initialization/InitDistribution.cpp
```diff
@@ -191,6 +195,8 @@ namespace impactx
             amrex::Abort("Unknown distribution: " + distribution_type);
         }
 
+        this->ResizeMesh();
```
ax3l, line 205:
```suggestion
        // Resize the mesh to fit the spatial extent of the beam and then
        // redistribute particles, so they reside on the MPI rank that is
        // responsible for the respective spatial particle position.
        this->ResizeMesh();
```

ax3l (commented):
LGTM, I am just adding a few comments :)

ax3l (approved):
Thanks a lot :rocket: :sparkles:

PR #84, opened 2022-03-15
Theory: Start Assumptions

ax3l:
Start to document assumptions


cemitch99 (approved):
I assume that this PR is just to create intro.rst, to which we will gradually add over time.  It looks good to me.

PR #85, opened 2022-03-16
Added a linear constant focusing element.

cemitch99:
This will be useful for future benchmarking with space charge.

### DIFF
--- a/examples/cfchannel/README.rst
+++ b/examples/cfchannel/README.rst
@@ -1,0 +1,28 @@
+.. _examples-cfchannel:
+
+Constant Focusing Channel
+==========================
+
+Stationary beam in a constant focusing channel (without space charge).
+
+The matched Twiss parameters at entry are:
+
+* :math:`\beta_\mathrm{x} = 1.0` m
+* :math:`\alpha_\mathrm{x} = 0.0`
+* :math:`\beta_\mathrm{y} = 1.0` m
+* :math:`\alpha_\mathrm{y} = 0.0`
+
+We use a 2 GeV proton beam with initial unnormalized rms emittance of 1 um. 
+The longitudinal beam parameters are chosen so that the bunch has radial
+symmetry when viewed in the beam rest frame.
+
+The particle distribution should remain unchanged, to within the level expected due to numerical particle noise.
+This fact is independent of the length of the channel.  This is tested using the second moments of the distribution.
+
+In this test, the initial and final values of :math:`\sigma_x`, :math:`\sigma_y`, :math:`\sigma_t`, :math:`\epsilon_x`, :math:`\epsilon_y`, and :math:`\epsilon_t` must agree with nominal values.
+
+
+.. literalinclude:: input_cfchannel.in
+   :language: ini
+   :caption: You can copy this file from
+``examples/cfchannel/input_cfchannel.in``.
--- a/examples/cfchannel/README.rst
+++ b/examples/cfchannel/README.rst
@@ -1,7 +1,7 @@
 .. _examples-cfchannel:
 
 Constant Focusing Channel
-==========================
+=========================
 
 Stationary beam in a constant focusing channel (without space charge).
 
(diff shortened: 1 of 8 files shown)


src/particles/elements/ConstF.H
```diff
@@ -0,0 +1,96 @@
+/* Copyright 2021 Chad Mitchell, Axel Huebl
```
ax3l:
```suggestion
/* Copyright 2022 Chad Mitchell, Axel Huebl
```

src/ImpactX.cpp
```diff
@@ -252,6 +252,13 @@ namespace impactx
                 pp_element.get("g", g);
                 pp_element.get("K2", K2);
                 m_lattice.emplace_back( DipEdge(psi, rc, g, K2) );
+            } else if (element_type == "constf") {
```
ax3l, line 255: That looks good.

Do you want to add a minimal input set that uses (aka: "covers") this code path already in continuous integration (CI)? Does not need to be super meaningful yet, but would make sure the code does not segfault as we continue to add code to the repo.

src/particles/elements/ConstF.H
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        /** This is a constf functor, so that a variable of this type can be used like a
+         *  constf function.
+         *
+         * @param p Particle AoS data for positions and cpu/id
+         * @param px particle momentum in x
+         * @param py particle momentum in y
+         * @param pt particle momentum in t
+         * @param refpart reference particle
```
ax3l:
```suggestion
         * @param refpart reference particle (unused)
```

ax3l (commented):
Looks great! :+1:

src/ImpactX.cpp
```diff
@@ -252,6 +252,13 @@ namespace impactx
...
+                amrex::Real ds, kx, ky, kt;
+                pp_element.get("ds", ds);
```
ax3l, line 257: Please don't forget to add those to `docs/source/usage/parameters.rst`

src/ImpactX.cpp
cemitch99: Yes, I can create another benchmark input file to test the consf element without space charge, which should work as-is.  Will work on this today.

src/ImpactX.cpp
cemitch99: Will do this; thanks for the reminder.

src/ImpactX.cpp
cemitch99: Just added a commit.  (See above.)

src/ImpactX.cpp
cemitch99: Done.  (See above.)

examples/cfchannel/README.rst
```diff
@@ -0,0 +1,28 @@
...
+The particle distribution should remain unchanged, to within the level expected due to numerical particle noise.
+This fact is independent of the length of the channel.  This is tested using the second moments of the distribution.
+
+In this test, the initial and final values of :math:`\sigma_x`, :math:`\sigma_y`, :math:`\sigma_t`, :math:`\epsilon_x`, :math:`\epsilon_y`, and :math:`\epsilon_t` must agree with nominal values.
+
+
+.. literalinclude:: input_cfchannel.in
+   :language: ini
+   :caption: You can copy this file from
+``examples/cfchannel/input_cfchannel.in``.
```
ax3l:
```suggestion
   :caption: You can copy this file from ``examples/cfchannel/input_cfchannel.in``.
```

examples/cfchannel/README.rst
```diff
@@ -0,0 +1,28 @@
+.. _examples-cfchannel:
+
+Constant Focusing Channel
+==========================
```
ax3l: `.rst` is a bit picky with the alignment here:
```suggestion
Constant Focusing Channel
=========================
```

ax3l (approved):
Thank you 👍

PR #87, opened 2022-03-17
Doc: Major Cori Update

ax3l:
Update Cori instructions after the major upgrade yesterday.

Before using the new modules, remove the old, pre-build software:
```bash
rm -rf $HOME/sw/knl $HOME/sw/haswell $HOME/sw/gpu $HOME/sw/cori_gpu
```


cemitch99 (approved):
This also works for me.

PR #88, opened 2022-03-22
Docs: Update Assumptions

ax3l:
Add update to assumptions based on review in our weekly meeting.

Follow-up to #84

### DIFF
--- a/docs/source/theory/intro.rst
+++ b/docs/source/theory/intro.rst
@@ -17,7 +17,9 @@ This is a work-in-progress list of physical assumptions implemented in the numer
 Tracking and Lattice Optics
 """""""""""""""""""""""""""
 
-* **velocity spread:** when solving space-charge effects, we assume that the relative spread of velocities of particles in the beam is negligible compared to the velocity of the reference particle
+* **tracking through lattice optics:** is treated through linear order with respect to the reference particle
+
+  * **velocity spread:** the above linearization implies that, when solving space-charge effects, we assume that the relative spread of velocities of particles in the beam is negligible compared to the velocity of the reference particle
 
 
 Space Charge (Poisson Solver)
--- a/docs/source/theory/intro.rst
+++ b/docs/source/theory/intro.rst
@@ -12,6 +12,6 @@ Assumptions
 
 This is a work-in-progress list of physical assumptions implemented in the numerics of ImpactX:
 
-* **quasi-electrostatic:** we assume there are retardation effects
+* **electrostatic in the bunch frame:** we assume there are no retardation effects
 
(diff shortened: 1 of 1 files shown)


cemitch99 (commented):
I think it may help to break the assumptions into two sections:  one section for the Poisson solver, and one section for the tracking and lattice optics.

The first bullet (electostatic on the bunch frame) would go into the first section.

The second bullet could be replaced by saying that the dynamics is treated through linear order  (for now).

ax3l:
Thank you Chad, those are great suggestions. Updated accordingly.

cemitch99 (commented):
Under "Tracking and Lattice Optics", what about the statement that the tracking through the lattice optics is treated through linear order with respect to the reference particle?  This is precise, and is more accurate that commenting about the velocity spread only.

docs/source/theory/intro.rst
```diff
@@ -7,11 +7,20 @@ Introduction
...
 -----------
 
-This is a work-in-progress list of physical assumptions implemented in the numerics of ImpactX:
+This is a work-in-progress list of physical assumptions implemented in the numerics of ImpactX.
+
+
+Tracking and Lattice Optics
+"""""""""""""""""""""""""""
+
+* **velocity spread:** when solving space-charge effects, we assume that the relative spread of velocities of particles in the beam is negligible compared to the velocity of the reference particle
```
ax3l: Suggested by @cemitch99:
```suggestion
* **tracking through lattice optics:** is treated through linear order with respect to the reference particle

  * **velocity spread:** the above linearization implies that, when solving space-charge effects, we assume that the relative spread of velocities of particles in the beam is negligible compared to the velocity of the reference particle
```

docs/source/theory/intro.rst
```diff
@@ -7,11 +7,22 @@ Introduction
...
+
+  * **velocity spread:** the above linearization implies that, when solving space-charge effects, we assume that the relative spread of velocities of particles in the beam is negligible compared to the velocity of the reference particle
+
 
-* **quasi-electrostatic:** we assume there are retardation effects
+Space Charge (Poisson Solver)
+"""""""""""""""""""""""""""""
 
-* **longitudinal velocity spread:** when solving space-charge effects, we assume that the relative spread of velocities of particles in the beam is negligible compared to the velocity of the reference particle
+* **electrostatic in the bunch frame:** we assume there are no retardation effects
```
ax3l:
```suggestion
* **electrostatic in the bunch frame:** we assume there are no retardation effects and we solve the Poisson equation in the bunch frame
```

PR #91, opened 2022-03-28
Fix MR Level Loops: `finestLevel()`

ax3l:
Only iterate up to `finestLevel` of the simulation.

- `maxLevel`: the maximum allowed refinement level in a simulation
- `finestLevel`: the currently used, finest refinement level (can change over steps between `0` and `maxLevel` with AMR)


src/particles/transformation/CoordinateTransformation.cpp
```diff
@@ -27,8 +27,8 @@ namespace transformation {
         amrex::ParticleReal const pd = ref_part.pt;  // Design value of pt/mc2 = -gamma
 
         // loop over refinement levels
-        int const nLevel = pc.maxLevel();
```
ax3l, line 30: bugfix: before this PR, we forgot a +1 here

PR #107, opened 2022-04-07
Initialization: Split Files

ax3l:
The `ImpactX.cpp` files got quite long and spaghetti-ish, so it is time to introduce some structure.

Split out initialization logic, so that we can compile individual `.cpp` files in parallel.

Also deduplicates some of the diagnostics logic to print out the initial beam distribution.


cemitch99 (approved):
This structure looks good to me.  I absolutely agree about 'ImpactX.cpp'.

src/initialization/InitDistribution.cpp
```diff
@@ -0,0 +1,275 @@
...
+        if (unit_type == "static") {
+            amrex::Print() << "Static units" << std::endl;
+        } else if (unit_type == "dynamic") {
+            amrex::Print() << "Dynamic units" << std::endl;
+        } else {
+            amrex::Abort("Unknown units (static/dynamic): " + unit_type);
+        }
+
+        amrex::Print() << "Initialized beam distribution parameters" << std::endl;
+    }
```
cemitch99, line 274: In Impact-Z, the same parameter input set is used for each element (although certain parameters may be interpreted differently in a few cases).  If we follow the same strategy here, much of the code duplication above can be reduced.  However, allowing different types to have different inputs does add some flexibility.  Maybe this could be useful at some future point.  Is it worth it?

cemitch99 (commented):
This structure looks good to me.  I absolutely agree about the need to split 'ImpactX.cpp'.

src/initialization/InitDistribution.cpp
ax3l: I would probably just add a helper function for the block
```C++
          if (amrex::ParallelDescriptor::IOProcessor()) {
              x.reserve(npart);
              y.reserve(npart);
              t.reserve(npart);
              px.reserve(npart);
              py.reserve(npart);
              pt.reserve(npart);

              for(amrex::Long i = 0; i < npart; ++i) {
                  KVdist(ix, iy, it, ipx, ipy, ipt, rng);
                  x.push_back(ix);
                  y.push_back(iy);
                  t.push_back(it);
                  px.push_back(ipx);
                  py.push_back(ipy);
                  pt.push_back(ipt);
              }
          }
```

for now. All we need is a template for the exact function we call.

The rest is then specific to the function.

ax3l: ->  #108 :)

Issue #103, opened 2022-03-30
Add Multi-Node Test


ax3l:
Add a multi-node (MPI) test.

Dependencies:
- #68: parallel init
- #109: parallel diagnostics

ax3l:
I added a first MPI-driven test in #109:
https://github.com/ECP-WarpX/impactx/pull/109#issuecomment-1095421586

```
3: Test command: /home/axel/src/spack/var/spack/environments/warpx-omp-dev/.spack-env/view/bin/mpiexec "-n" "2" "/home/axel/src/impactx/build/bin/impactx.MPI.OMP.DP" "/home/axel/src/impactx/examples/fodo/input_fodo.in"
3: Environment variables: 
3:  OMP_NUM_THREADS=1
3: Test timeout computed to be: 1500
3: MPI initialized with 2 MPI processes
3: MPI initialized with thread support level 0
3: OMP initialized with 1 OMP threads
3: AMReX (22.04-9-g84a1f7d0715b) initialized
3: boxArray(0) (BoxArray maxbox(2)
3:        m_ref->m_hash_sig(0)
3:        ((0,0,0) (7,7,7) (0,0,0)) ((8,0,0) (15,7,7) (0,0,0)) )
3: 
3: Beam kinetic energy (MeV): 2000
3: Bunch charge (C): 0
3: Particle type: electron
3: Number of particles: 10000
3: Beam distribution type: waterbag
3: Static units
3: Initialized beam distribution parameters
3: # of particles: 10000
3: Initialized element list
3:  ++++ Starting step=0
```

I think we need to remove some hacks in `main.cpp` and make sure it uses multiple boxes (currently: 2?), in a follow-up.

ax3l:
Further improved via #127

ax3l:
Improved in #161 and #165

MPI Python test in #153 (CI: #182)

PR #109, opened 2022-04-11
IO: Parallel Particle ASCII Output

ax3l:
Write one file per MPI rank (and pot. OpenMP thread).
Update post-processing logic to process this.

Needed for #103

To Do:
- [x] add a parallel test

X-ref:
- depends on #73
- https://github.com/ECP-WarpX/impactx/pull/73#issuecomment-1093075788

### DIFF
--- a/src/particles/diagnostics/DiagnosticOutput.cpp
+++ b/src/particles/diagnostics/DiagnosticOutput.cpp
@@ -7,6 +7,7 @@
 #include "DiagnosticOutput.H"
 
 #include <AMReX_Extension.H>  // for AMREX_RESTRICT
+#include <AMReX_ParallelDescriptor.H>  // for ParallelDescriptor
 #include <AMReX_REAL.H>       // for ParticleReal
 #include <AMReX_Print.H>      // for PrintToFile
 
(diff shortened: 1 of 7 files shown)


src/particles/diagnostics/DiagnosticOutput.cpp
```diff
@@ -26,52 +27,66 @@ namespace impactx::diagnostics
...
-                        // write particle data to file
-                        amrex::PrintToFile(file_name)
-                            << x << " " << y << " "<< t << " "
-                            << px << " " << py << " " << pt << "\n";
-                    } // i=0...np
-                } // if( otype == OutputType::PrintParticles)
-            } // end loop over all particle boxes
-        } // env mesh-refinement level loop
+                                // write particle data to file
+                                amrex::PrintToFile(file_name)
```
atmyers:
```suggestion
                                amrex::AllPrintToFile(file_name)
```

Edit - actually, this still won't work because each rank will write separate files, I think you need to duplicate the file open / write / flush logic in the amrex function here.

src/particles/diagnostics/DiagnosticOutput.cpp
```diff
@@ -26,52 +27,66 @@ namespace impactx::diagnostics
...
-            for (ParIt pti(tmp, lev); pti.isValid(); ++pti) {
-                const int np = pti.numParticles();
+            amrex::PrintToFile(file_name) << "x y t px py pt\n";
+        }
+
+        const int MyProc = amrex::ParallelDescriptor::MyProc();
+
+        for (int proc = 0; proc < amrex::ParallelDescriptor::NProcs(); proc++) {
+            // note: this is the same logic as in amrex::ParticleContainer::WriteAsciiFile
+            //   this will obviously not scale and also has the assumption
```
atmyers: Yes, the ascii format is meant for small tests and should not be used at scale. Also I'm not sure about the flush thing, in the AMReX function each rank flushes the file before the next rank is allowed to open it...

ax3l: Correct, also here. Maybe I should add a note in the API interface of the function that this is temporary.

src/particles/diagnostics/DiagnosticOutput.cpp
ax3l: oh, ok. let me check.

ax3l: We could also generate N files for now via `AllPrintToFile` and adjust the post-processing scripts for now to use a glob.

ax3l:
I also added a first MPI-parallel test (#103) now:
```
3: Test command: /home/axel/src/spack/var/spack/environments/warpx-omp-dev/.spack-env/view/bin/mpiexec "-n" "2" "/home/axel/src/impactx/build/bin/impactx.MPI.OMP.DP" "/home/axel/src/impactx/examples/fodo/input_fodo.in"
3: Environment variables: 
3:  OMP_NUM_THREADS=1
3: Test timeout computed to be: 1500
3: MPI initialized with 2 MPI processes
3: MPI initialized with thread support level 0
3: OMP initialized with 1 OMP threads
3: AMReX (22.04-9-g84a1f7d0715b) initialized
3: boxArray(0) (BoxArray maxbox(2)
3:        m_ref->m_hash_sig(0)
3:        ((0,0,0) (7,7,7) (0,0,0)) ((8,0,0) (15,7,7) (0,0,0)) )
3: 
3: Beam kinetic energy (MeV): 2000
3: Bunch charge (C): 0
3: Particle type: electron
3: Number of particles: 10000
3: Beam distribution type: waterbag
3: Static units
3: Initialized beam distribution parameters
3: # of particles: 10000
3: Initialized element list
3:  ++++ Starting step=0
```

I think we need to remove some hacks in `main.cpp` and make sure it uses multiple boxes (currently: 2?), in a follow-up.

src/particles/diagnostics/DiagnosticOutput.cpp
ax3l: Done that way now, pandas is powerful enough to handle this :)

atmyers (approved):

PR #111, opened 2022-04-12
Added benchmark examples to test Gaussian and KV distributions.

cemitch99:
These and further examples of distribution benchmarks will appear together in the subfolder:

examples/distgen/


ax3l (approved):
Thank you :)

Issue #104, opened 2022-03-31
MAD-X input reader: constants, elements, lines, segments


ax3l:
We want to support reading accelerator lattices directly from MAD-X input files.
For this, we will need to support at least the following syntax elements:
- constants
- elements
- lines
- segments
- simple math

from MAD-X definitions.

For that, we could explore if @janfschmidt's `palattice` could be used as a library :)
https://github.com/janfschmidt/palattice

References on MAD-X:
- https://mad.web.cern.ch/mad/webguide/manual.html
- https://github.com/MethodicalAcceleratorDesign/MAD-X/blob/master/doc/latexuguide/format.tex
- MAD websites
  - MAD-X      http://cern.ch/madx   (current version, https://github.com/MethodicalAcceleratorDesign/MAD-X/blob/master/src/mad_parse.c)
  - MAD-9      http://cern.ch/mad9
  - MAD-8      http://cern.ch/mad8
  - MAD-NG   [http://cern.ch/madng/dev](http://cern.ch/mad/releases/madng) (next generation under development)

Example files for MAD-X lattices:
- LBNL/ImpactX/ABLASTR team: https://drive.google.com/drive/u/0/folders/1-BZysj84iEjhDOw-JsEDuu7IytGdl8YN

ax3l:
@qianglbl and @cemitch99 mentioned:
MAD-X also has an "optics" command that can output a simple CSV file with one element per row. IMPACT-Z can read this file in.

We might want to add support to:
- write the same file format for the lattice from ImpactX
- read in the same file format to initialize our lattice.

### Optics CSV file structure

Columns with names (column 1), type (column 2) and values: meaning of the column depends on the element type (column 2).

Values: note that parameters are not in that CSV file, e.g., fringe-field details.

ax3l:
With the new Python layer, we could also implement the first MAD(-X) reader on the Python side.

Warp has some routines we can get logic from (@jlvay):
- https://bitbucket.org/berkeleylab/warp/src/master/scripts/lattice/lattice.py
- https://warp.lbl.gov/how-tos/lattice/adding-lattice-elements

n01r:
This is the link to the Python MADX parser that Andreas Adelmann wrote and pushed to a repo
https://gitlab.psi.ch/adelmann/pymadxparser.git

Implemented via #214

np-eazy:
[...]

Issue #93, opened 2022-03-30
Python Bindings


ax3l:
Use https://github.com/AMReX-Codes/pyamrex to establish Python bindings.

PR #118, opened 2022-04-29
Added a thin multipole element

cemitch99:
This includes addition of a new element type "multipole", which accepts as input an arbitrary multipole index, together with a normal and skew integrated strength, and produces the kick associated with a thin multipole element.

An example is included, which checks the propagation of the beam moments.  

Possible future changes:
1) Allow the possibility to input an arbitrary number of multipole coefficients, and take the superposition of these
1a) for the special case of a thin multipoles, we can also stack them in a line, e.g., via Python inputs #93 & Mad-X #104 
2) Modify the example to compute the Hamiltonian and check its preservation under the multipole kick

### DIFF
--- a/docs/source/usage/examples.rst
+++ b/docs/source/usage/examples.rst
@@ -13,6 +13,7 @@ This section allows you to **download input files** that correspond to different
    examples/cfchannel/README.rst
    examples/kurth/README.rst
    examples/fodo_rf/README.rst
+   examples/multipole/README.rst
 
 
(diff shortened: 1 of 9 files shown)


ax3l:
@cemitch99 I think this branch still needs the parameter read logic added to `src/initialization/InitElement.cpp` :)

ax3l (approved):
:tada:

qianglbl (approved):

PR #126, opened 2022-05-03
Improve Thin Multipole Performance

ax3l:
Follow-up to #118:

- complex type: avoid type mixing (default for `amrex::ComplexType<T>` is `T=double` not `amrex::ParticleReal`)
- improve performance: compute factorial of the multipole index only once and on CPU
- add `const` where appropriate


cemitch99 (approved):
The changes look good to me.

PR #129, opened 2022-05-13
Rename ImpactX:: Public Methods

ax3l:
Rename some common methods to be more descriptive, avoiding abbreviations and readying them to be replaced by non-inputs-file control from Python.


cemitch99 (approved):
This looks fine to me.  I agree that this could be helpful for debugging/ease of reading because the function names are more descriptive.

PR #139, opened 2022-05-31
Analysis: Id as Pandas Dataframe Index

ax3l:
Use the unique particle ID (#135) as pandas column index. This is more useful instead of adding an auto-generated, zero-based index in the pandas dataframe: joins of multiple time steps can be done on the default index (#137) and there is no confusion by having an extra column.


cemitch99 (approved):
This looks good to me, thanks.

PR #140, opened 2022-06-07
Modified the initial beam size in the IOTA lens benchmark example.

cemitch99:
The previous beam size was too large, resulting in a few particles experiencing large changes (>10%) in the quantities that should be invariants of motion.  The maximum change is now ~10^(-3).


ax3l:
Restarting CI (had some hick up)

ax3l:
Awesome, thanks!
@cemitch99 do you also want to edit the Python test script to make the tolerance more narrow? :)

ax3l (approved):
Awesome, thank you! :sparkles:

examples/iota_lens/analysis_iotalens.py
```diff
@@ -96,20 +96,20 @@ def read_all_files(file_pattern):
     rsuffix='_initial'
 )
 # add new columns: dH and dI
-beam_joined['dH'] = beam_joined["H_initial"] - beam_joined["H_final"]
-beam_joined['dI'] = beam_joined["I_initial"] - beam_joined["I_final"]
+beam_joined['dH'] = abs(beam_joined["H_initial"] - beam_joined["H_final"])
+beam_joined['dI'] = abs(beam_joined["I_initial"] - beam_joined["I_final"])
```
ax3l: If we use the `.abs()` method of the resulting dataframe, then pandas might be able to compute this in delayed fashion (when we access it later - aka lazy computing).
```suggestion
beam_joined['dH'] = (beam_joined["H_initial"] - beam_joined["H_final"]).abs()
beam_joined['dI'] = (beam_joined["I_initial"] - beam_joined["I_final"]).abs()
```

PR #146, opened 2022-06-22
Prepare for global reference particle push

cemitch99:
Promoted the reference particle to an 8-tuple on the extended phase space, with coordinates:

(x,px,y,py,z,pz,t,pt)

and initialized these coordinates.  This is done in preparation for pushing the reference particle in global coordinates.

Next steps:
1) add a global reference particle push to each element struct
2) add a call to the global reference particle push in "Push.cpp".


cemitch99:
Superseded by PR#147.

PR #149, opened 2022-06-24
Added operator to push reference particle in global coordinates to all elements

cemitch99:
Modified structs for "Drift.H", "Quad.H", etc. to include operator to update the reference particle in global laboratory coordinates (x,px,y,py,z,pz,t,pt).

Next step:

- [x] Add a call to this operator within `Push.cpp`.


ax3l (approved):
Thanks a lot :)

I added a few commits to finalize the API and push the reference particle, too.

PR #150, opened 2022-06-27
Added (optional) input per element - number of space charge slices.

cemitch99:
For each thick element, an optional parameter "nslice" was added at input.

To be used as the number of slices (here = the number of space charge kicks) per element.

### To Do

- [x] add `nslice` to each element's constructor
- [x] use it in our evolve loop: https://github.com/ECP-WarpX/impactx/blob/81684134057d43b9f4ecf8b6dcf9221b84e4b9f8/src/ImpactX.cpp#L77
```c++
int num_steps = 1;
std::visit([&num_steps](auto & element){ num_steps = element.nslice; }, element_variant);
```
- [x] add diagnostics in each sub-step for now so we can plot distribution evolutions

### DIFF
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -175,7 +175,7 @@ Lattice Elements
 * ``<element_name>.type`` (``string``)
     Indicates the element type for this lattice element. This should be one of:
 
-        * ``drift`` for free drift. This requires this additional parameters:
+        * ``drift`` for free drift. This requires these additional parameters:
 
             * ``<element_name>.ds`` (``float``, in meters) the segment length
 
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -364,6 +364,13 @@ Numerics and algorithms
 Diagnostics and output
 ----------------------
 
+* ``diag.slice_step_diagnostics`` (``boolean``, default: ``false``)
+  By default, diagnostics is performed at the beginning and end of the simulation.
+  Enabling this flag will write diagnostics every step and slice step
+
+* ``diag.file_min_digits`` (``integer``, optional, default: ``6``)
+    The minimum number of digits used for the iteration number appended to the diagnostic file names.
+
 Diagnostics related to integrable optics in the IOTA nonlinear magnetic insert element:
 
 * ``diag.alpha`` (``float``, unitless) Twiss alpha of the bare linear lattice at the location of output for the nonlinear
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -175,9 +175,12 @@ Lattice Elements
 * ``<element_name>.type`` (``string``)
     Indicates the element type for this lattice element. This should be one of:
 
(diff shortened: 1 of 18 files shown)


docs/source/usage/parameters.rst
```diff
@@ -175,10 +175,13 @@ Lattice Elements
 * ``<element_name>.type`` (``string``)
     Indicates the element type for this lattice element. This should be one of:
 
-        * ``drift`` for free drift. This requires this additional parameter:
+        * ``drift`` for free drift. This requires this additional parameters:
```
```suggestion
        * ``drift`` for free drift. This requires these additional parameters:
```

ax3l (approved):

ax3l (changes requested):
Let's not merge this one yet, because it would add if merged now add a parameter that is silently ignored.

We need to:
- [x] add this to the element APIs
- [ ] use it once space charge is used.

By not yet merging this, users get a warning instead of a silently ignored parameter as we still develop this. I'll keep this open for a bit more.

src/initialization/InitElement.cpp
```diff
@@ -40,17 +40,23 @@ namespace impactx
             // Initialize the corresponding element according to its type
             if (element_type == "quad") {
                 amrex::Real ds, k;
+                int nslice = 1;
```
ax3l: We could consider adding a simulation parameter that can be used as the default value for all elements (unless otherwise specified). Would that make sense/be user friendly or is it better to force users to give us explicitly the substeps per element in any case?

cemitch99: I suggest that we discuss this with Ji/others next week.  One issue is that elements have different lengths, and also different dynamics, so the optimal choice of step size may vary significantly from one element to the next.  This is why both IMPACT-Z and MaryLie/IMPACT input the number of substeps per element.  However, having a universal default could be useful for convergence studies.  One option is to replace the initialization "nslice = 1" above with "nslice = nsteps".

ax3l: yes, that's what I was thinking. Just a global non-`1` default that can be overwritten per element.

cemitch99: I just added "num_steps" as an input, and this is now used as the default value of "nslice" for each element.

ax3l (approved):

docs/source/usage/parameters.rst
```diff
@@ -172,13 +172,20 @@ Lattice Elements
     A list of names (one name per lattice element), in the order that they
     appear in the lattice.
 
+* ``lattice.nslice`` (``integer``) optional (default: ``1`` unless thin element)
```
```suggestion
* ``lattice.nslice`` (``integer``) optional (default: ``1``)
```

docs/source/usage/parameters.rst
```diff
@@ -348,6 +364,13 @@ Numerics and algorithms
 Diagnostics and output
 ----------------------
 
+* ``diag.slice_step_diagnostics`` (``boolean``, default: ``false``)
```
```suggestion
* ``diag.slice_step_diagnostics`` (``boolean``, optional, default: ``false``)
```

cemitch99 (commented):
This looks good.

PR #152, opened 2022-06-27
Added output of reference particle information

cemitch99:
Added an additional diagnostic function in DiagnosticOutput.cpp to print the global reference particle coordinates to file.

- [x] This required access to "GetRefPart()".  To do:  Check const correctness.

### DIFF
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -57,6 +57,11 @@ namespace impactx
                                       diagnostics::OutputType::PrintParticles,
                                       "diags/initial_beam.txt");
 
+        // print initial reference particle to file
+        diagnostics::DiagnosticOutput(*m_particle_container,
+                                      diagnostics::OutputType::PrintRefParticle,
+                                      "diags/ref_particle.txt");
+
         // print the initial values of the two invariants H and I
         diagnostics::DiagnosticOutput(*m_particle_container,
                                       diagnostics::OutputType::PrintNonlinearLensInvariants,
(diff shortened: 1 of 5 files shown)


src/particles/diagnostics/DiagnosticOutput.H
```diff
@@ -22,7 +22,8 @@ namespace impactx::diagnostics
     enum class OutputType
     {
         PrintParticles, ///< ASCII diagnostics, for small tests only
-        PrintNonlinearLensInvariants ///< ASCII diagnostics for the IOTA nonlinear lens, for small tests only
+        PrintNonlinearLensInvariants, ///< ASCII diagnostics for the IOTA nonlinear lens, for small tests only
+        PrintRefParticle ///< ASCII diagnostics, for small tests only
```
ax3l, line 26:
```suggestion
        PrintRefParticle ///< ASCII diagnostics for the reference particle
```

ax3l (approved):
Looks great!
I added the additionally needed API access for the reference particle, so we can keep the container constant.

ax3l:
Will rebase #150 out, since it's unrelated and will need to stay open for a bit more.

src/ImpactX.cpp
```diff
@@ -57,6 +57,11 @@ namespace impactx
                                       diagnostics::OutputType::PrintParticles,
                                       "diags/initial_beam.txt");
 
+        // print initial reference particle to file
+        diagnostics::DiagnosticOutput(*m_particle_container,
+                                      diagnostics::OutputType::PrintRefParticle,
+                                      "diags/ref_particle.txt");
```
```suggestion
                                      "diags/initial_ref_particle.txt");
```

src/ImpactX.cpp
```diff
@@ -121,6 +126,11 @@ namespace impactx
                                       diagnostics::OutputType::PrintParticles,
                                       "diags/output_beam.txt");
 
+        // print final reference particle to file
+        diagnostics::DiagnosticOutput(*m_particle_container,
+                                      diagnostics::OutputType::PrintRefParticle,
+                                      "diags/ref_particle.txt");
```
```suggestion
                                      "diags/output_ref_particle.txt");
```

ax3l (approved):

ax3l:
@cemitch99 do you like to add this in a python analysis script in this PR? That would make the PR self-consistent (preferred), otherwise we can do a follow-up :)

PR #123, opened 2022-05-03
Python Bindings

ax3l:
Implements ImpactX as a Python module.
Related to #93

### To Do

- [x] needs https://github.com/ECP-WarpX/WarpX/pull/3081 / #128
- [x] needs #145
- [x] needs https://github.com/AMReX-Codes/amrex/pull/2842
- [x] needs #148
- [x] third party: pull in pyAMReX
- [x] Python unit tests: fixtures & tests
- [x] build: build AMReX with position independent code
- [x] add example inputs file(s): add a line of elements in a `for` loop

Follow-up:
- [ ] add a particle beam: #153

PR #153, opened 2022-06-27
Python: Init Beam Distributions

### LINKED ISSUE #93: Python Bindings

ax3l:
Close #93
Follow up to #123

- [x] depends on https://github.com/AMReX-Codes/pyamrex/pull/40
- [x] depends on https://github.com/AMReX-Codes/pyamrex/pull/51
- [x] add particle beam distributions
- [x] expose `AddParticles`

### DIFF
--- a/src/initialization/InitDistribution.cpp
+++ b/src/initialization/InitDistribution.cpp
@@ -7,6 +7,7 @@
  * Authors: Axel Huebl, Chad Mitchell, Ji Qiang
  * License: BSD-3-Clause-LBNL
  */
+#include "InitDistribution.H"
 #include "ImpactX.H"
 #include "particles/ImpactXParticleContainer.H"
 #include "particles/distribution/Waterbag.H"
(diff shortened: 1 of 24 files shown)


src/initialization/InitDistribution.cpp
```diff
@@ -267,6 +209,7 @@ namespace impactx
         refPart.z = 0.0;
         refPart.px = 0.0;
         refPart.py = 0.0;
+        // make the next two lines a helper function?
```
ax3l, line 212: @cemitch99 we could make this a helper member function in `RefPart`, what do you think?

ax3l: Discussed on Slack: same for "set_energy_MeV", "get_beta", "get_gamma", etc.?
> Yes, I agree this sounds good.

-> will be a separate PR

cemitch99 (commented):
Looks good to me.

PR #155, opened 2022-07-02
Add More Profiling

ax3l:
Add more profiling information for our tiny profiler summary as well as for profiling tools.


cemitch99 (approved):
This looks good to me.

PR #160, opened 2022-07-15
Modified ordering of evolve/push loop.

cemitch99:
Modified the ordering of the tracking loop as follows:

Moved the loop over elements from the innermost loop (in function Push) to the outermost loop (in function "evolve").  This is needed in order to apply space charge between or within elements.


ax3l (approved):
Thank you :sparkles:

PR #167, opened 2022-07-20
Starting Plotting of Benchmarks

ax3l:
Start checking in plots for benchmarks.

## FODO
fodo_sigma
fodo_scatter

# Chicane
chicane_sigma
chicane_scatter
note: last row should be labeled `ct [mm]` in the horizontal axis


examples/fodo/plot_fodo.py
```diff
@@ -0,0 +1,119 @@
...
+print(particle_42)
+
+# beam transversal size over steps
+moments = list(map(
+    lambda step: (step, get_moments(beam[beam["step"] == step])),
+    steps
+))
+#print(moments)
+sigx = list(map(
+    lambda step_val: step_val[1][0] * micron,
```
cemitch99: For most of the examples here, I suggest using units of mm for x, y.

examples/fodo/plot_fodo.py
```diff
@@ -0,0 +1,123 @@
...
+        read_file(file_pattern),
+        axis=0,
+        ignore_index=True,
+    ) #.set_index('id')
+
+
+# initial/final beam on rank zero
+beam = read_time_series("build/examples/diags/beam_*")
+#print(beam)
+micron = 1.e6
```
cemitch99: Suggest using mm = 1.e3 instead.

examples/fodo/plot_fodo.py
ax3l: good point, updated :)

cemitch99 (commented):
Usually we view the output of (at least) sigx and sigy at a large number of z-locations, to see their continuous evolution, so I think it works to output once per slice, as you have it here.  We can then make nslice as large as needed.

However, we usually don't output the phase space projections as frequently, so I recommend producing output only once per element.

cemitch99 (commented):
Also, we more often show plots of the projections (x [mm], px [mrad]) or (y [mm], py [mrad]).

cemitch99 (commented):
Looking great!  We will need to adjust spacing between the scatterplots so that the horizontal axes are visible.

examples/fodo/plot_fodo.py
```diff
@@ -0,0 +1,184 @@
...
+# beam transversal size & emittance over steps
+moments = list(map(
+    lambda step: (step, get_moments(beam[beam["step"] == step])),
+    steps
+))
+#print(moments)
+sigx = list(map(lambda step_val: step_val[1][0] * millimeter, moments))
+sigy = list(map(lambda step_val: step_val[1][1] * millimeter, moments))
+emittance_x = list(map(lambda step_val: step_val[1][3] * mm_mrad, moments))
+emittance_y = list(map(lambda step_val: step_val[1][4] * mm_mrad, moments))
```
ax3l: @cemitch99 commented:
```
We could multiply by 1e9 instead, and print the value in (nm).
(In this case the dimensionless factor of "rad" is usually suppressed.)
```

cemitch99 (approved):
Looks great.

PR #168, opened 2022-07-20
Elements: slice `ds`

ax3l:
Slice the segment length `ds` per push according to the number of slices we use between space charge pushes.

Follow-up to #160


ax3l:
I did a quick test of all benchmarks with `lattice.nslice = 10` and everything still passed.

cemitch99 (approved):
Perfect.

PR #169, opened 2022-07-21
Diagnostics: Zero-Based

ax3l:
Reorder the diagnostics output:
- step zero: initial state before start
- all other steps: one-based at the end of each step

### DIFF
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -55,23 +55,33 @@ namespace impactx
     {
         BL_PROFILE("ImpactX::evolve");
 
-        // print initial particle distribution to file
-        diagnostics::DiagnosticOutput(*m_particle_container,
-                                      diagnostics::OutputType::PrintParticles,
-                                      "diags/initial_beam.txt");
-
-        // print initial reference particle to file
-        diagnostics::DiagnosticOutput(*m_particle_container,
-                                      diagnostics::OutputType::PrintRefParticle,
-                                      "diags/initial_ref_particle.txt");
-
-        // print the initial values of the two invariants H and I
-        diagnostics::DiagnosticOutput(*m_particle_container,
-                                      diagnostics::OutputType::PrintNonlinearLensInvariants,
-                                      "diags/initial_nonlinear_lens_invariants.txt");
-
         // a global step for diagnostics including space charge slice steps in elements
-        int global_step = -1;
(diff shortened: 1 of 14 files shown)


src/ImpactX.cpp
```diff
@@ -55,23 +55,33 @@ namespace impactx
...
+        {
+            pp_diag.queryAdd("file_min_digits", file_min_digits);
+
+            // print initial particle distribution to file
+            std::string diag_name = amrex::Concatenate("diags/beam_", global_step, file_min_digits);
+            diagnostics::DiagnosticOutput(*m_particle_container,
+                                          diagnostics::OutputType::PrintParticles,
+                                          diag_name);
+
+            // print initial reference particle to file
```
cemitch99, line 73: Eventually, it will be good to write the reference particle coordinates at each step as (s, x, px, y, py, z, pz, t, pt), where s denotes the path length (sum of ds over all element slices).  We can put all of this data into a single file, which can be used for plotting the reference orbit if/when desired.

cemitch99 (approved):
This looks good to me.

src/ImpactX.cpp
ax3l: Good point. So this PR already writes the reference particle into a single file - but we don't add `s` yet in the table (this PR adds the step at least)

cemitch99: This will only be important if there are bending elements, so that the reference path length and the global z-coordinate do not coincide.

ax3l: Do you think we should add an integrated `s` attribute to the reference particle for this?

cemitch99: I considered this, and I decided against it at the time, because I think of s as an independent variable parametrizing the reference orbit (an alternative to t), and because it can be obtained directly from the element values of ds.

However, nothing is lost by including it as an attribute of the reference particle.  One could then update s along with the the global coordinates within each element slice, and then access this value as needed (instead of computing s elsewhere).

cemitch99: I can include this change in a follow-up PR if needed.

ax3l: I agree, question is basically convenience (and number of errors we do numerically adding a lot of `ds` values if they are quite different over many orbits). I am ok to add it to alleviate some post-processing work.

PR #170, opened 2022-07-22
Added integrated path length s as a reference particle attribute.

cemitch99:
The integrated path length "s" (m) is used in IMPACT-Z to identify the beam's location within the accelerator lattice.  

1) This has been added as an attribute of the reference particle.

2) The path length s is updated during each reference particle push.

The value of s is accessed in "DiagnosticOutput"~~, but it is not yet printed, in order to maintain compatibility with existing postprocessing scripts.  These can be modified as needed to use the value of "s".~~

- [x] add in Python bindings, too

### DIFF
--- a/src/particles/diagnostics/DiagnosticOutput.cpp
+++ b/src/particles/diagnostics/DiagnosticOutput.cpp
@@ -35,7 +35,7 @@ namespace impactx::diagnostics
             } else if (otype == OutputType::PrintNonlinearLensInvariants) {
                 amrex::AllPrintToFile(file_name) << "id H I\n";
             } else if (otype == OutputType::PrintRefParticle) {
-                amrex::AllPrintToFile(file_name) << "step x y z t px py pz pt\n";
+                amrex::AllPrintToFile(file_name) << "step s x y z t px py pz pt\n";
             }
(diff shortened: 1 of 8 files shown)


src/particles/diagnostics/DiagnosticOutput.cpp
```diff
@@ -141,6 +141,7 @@ namespace impactx::diagnostics
                     // preparing to access reference particle data: RefPart
                     RefPart const ref_part = pc.GetRefParticle();
 
+                    amrex::ParticleReal const s = ref_part.s;
```
ax3l, line 144: this needs to be added in two locations now:
- printed in the header (line 45)
- printed in the data table (line 156+)

ax3l:
> The value of s is accessed in "DiagnosticOutput", but it is not yet printed, in order to maintain compatibility with existing postprocessing scripts. These can be modified as needed to use the value of "s".

Our postprocessing scripts via Pandas read in new columns automatically, including identification of the new header, and do not fail if we add new columns :) Thus, added accordingly :)

ax3l (approved):
Thanks, this looks great! :)

PR #172, opened 2022-07-23
Python: Test Catch User Errors

ax3l:
Test multiple possible usage errors from Python. We want to make sure that these errors throw exceptions and do not abort/terminate the Python process. That way, the user-experience in IPython/Jupyter notebooks will be better (no session restart needed).

- [x] depends on https://github.com/AMReX-Codes/amrex/pull/2936 (via #238)
- [x] depends on https://github.com/AMReX-Codes/amrex/pull/2944 (via ...)

### DIFF
--- a/tests/python/conftest.py
+++ b/tests/python/conftest.py
@@ -13,7 +13,7 @@ else:
     print("NO mpi4py load")
 
 
-@pytest.fixture(autouse=True, scope="session")
+@pytest.fixture(autouse=True, scope="function")
 def amrex_init():
     amrex.initialize(
         [
--- a/src/ImpactX.H
+++ b/src/ImpactX.H
@@ -82,6 +82,10 @@ namespace impactx
             int npart
         );
 
+        /** Validate the simulation is ready to run via @see evolve
+         */
+        void validate ();
+
         /** Run the main simulation loop for a number of steps
          */
         void evolve ();
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -55,23 +55,11 @@ namespace impactx
     {
         BL_PROFILE("ImpactX::evolve");
 
+        validate();
+
         // a global step for diagnostics including space charge slice steps in elements
         //   before we start the evolve loop, we are in "step 0" (initial state)
         int global_step = 0;
-
-        // count particles - if no particles are found in our particle container, then a lot of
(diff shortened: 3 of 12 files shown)


tests/python/conftest.py
```diff
@@ -10,7 +10,7 @@
 else:
     print("NO mpi4py load")
 
-@pytest.fixture(autouse=True, scope='session')
+@pytest.fixture(autouse=True, scope='function')
```
ax3l: Looks like I have to work more on these scopes:
```
../../../tests/python/test_impactx.py::test_impactx_nofile ERROR
../../../tests/python/test_impactx.py::test_impactx_noparticles amrex::Error::0::ParmParse::Initialize(): already initialized! !!!
SIGABRT
```

ax3l: I needed to change to `session` because I want to reset the `ParmParse` parameters, so that things like `algo.particle_shape` are truly unset for the next test.

cemitch99 (approved):
This looks good to me.

tests/python/conftest.py
ax3l: (Problem does not show up if a test is run individually.)

RTSandberg (approved):
I don't see why this wouldn't pass but it looks good to me

tests/python/conftest.py
ax3l: Found it finally :tada: 
https://github.com/AMReX-Codes/amrex/pull/2936

src/particles/transformation/T2Z.H
```diff
@@ -62,8 +62,8 @@ namespace transformation
             amrex::ParticleReal const arg = 1.0_prt + pow(m_pzd+pt, 2) + pow(px, 2) + pow(py, 2);
 
             // check for validity of arguments
-            AMREX_ALWAYS_ASSERT_WITH_MESSAGE(argd > 0.0_prt, "invalid ptd arg (<=0)");
-            AMREX_ALWAYS_ASSERT_WITH_MESSAGE(arg > 0.0_prt, "invalid pt arg (<=0)");
+            AMREX_ASSERT_WITH_MESSAGE(argd > 0.0_prt, "invalid ptd arg (<=0)");
```
ax3l: Moved to debug-only because this is in a kernel as part of a hot loop.

ax3l:
ah, got ya: https://github.com/AMReX-Codes/amrex/pull/2944

PR #179, opened 2022-07-26
Fix Test Directories: Unique

ax3l:
Give each test a unique working directory. That way, we do not need to clean up the `diags` output between runs and can run tests in parallel if we want to with `ctest --test-dir build --output-on-failure -j 4`.


cemitch99 (approved):
This looks good to me.  As the number and complexity of the tests increases, I think running tests in parallel will become essential.

PR #180, opened 2022-07-26
Added reference particle helper functions.

### LINKED ISSUE #178: Reference Particle Helpers
Let's add helper functions to the reference particle to:
- set energy in MeV
- get energy in MeV, gamma, gamma-1
- get momentum in normalized units, beta*gamma
- get beta

cemitch99:
Added the following functions associated with the reference particle in ImpactXParticleContainer.cpp:

Set_Energy_MeV - initialize reference energy
GetGamma - return reference particle gamma
GetBeta - return reference particle beta
GetBetaGamma - return beta*gamma
Get_Energy_MeV - return reference particle kinetic energy

Close #178

### DIFF
--- a/src/particles/ImpactXParticleContainer.H
+++ b/src/particles/ImpactXParticleContainer.H
@@ -141,6 +141,17 @@ namespace impactx
         void
         SetRefParticle (RefPart const refpart);
 
+
+        /** Initialize reference particle energy
+         *
+         * @param energy initial kinetic energy (MeV)
+         * @param massE particle rest energy (MeV)
+         */
+        void
+        Set_Energy_MeV (amrex::ParticleReal const energy,
+            amrex::ParticleReal const massE);
+
+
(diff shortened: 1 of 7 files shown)


src/particles/ImpactXParticleContainer.H
```diff
@@ -155,6 +166,43 @@ namespace impactx
...
         GetRefParticle () const;
 
+
+        /** Get reference particle relativistic gamma
+         *
+         * @param refpart reference particle
+         * @returns relativistic gamma
+         */
+        amrex::ParticleReal
+        GetGamma (RefPart const refpart);
```
ax3l: Since we have access to the reference particle, we could make this an empty argument.

Alternatively, we could implement all setters and getters directly in the `RefPart` class? That is what I originally thought and would make it more explicit which energy/gamma/etc. we mean.

PR #183, opened 2022-07-27
Add a benchmark example using the IOTA linear lattice.

cemitch99:
This example contains the "bare" linear lattice of the IOTA storage ring (v. 8.4), and propagates a matched beam over a single turn.


ax3l (approved):
Awesome, thank you! :rocket: :sparkles:

cemitch99 (commented):
The Python implementation looks good to me so far.

ax3l:
The IOTA test case runs 5min, so I will speed it up a bit (~2x) by making it an MPI test.

We could consider creating output less often, it's definitely the ASCII test diagnostics that drags this at the moment.

qianglbl (approved):

PR #189, opened 2022-08-01
Add `algo.space_charge`

ax3l:
Add the `algo.space_charge` option (default: true). This controls if we do additional operations for space charge.

PR #191, opened 2022-08-01
Diagnostics: Benchmark Control

ax3l:
Since our diagnostics are currently slow ASCII outputs, we want to disable them during benchmarks. This adds this option via `diag.enable = 0` / `sim.set_diagnostics(False)`.

Also simplifies the existing option `sim.set_diags_slice_step_diagnostics` to `sim.set_slice_step_diagnostics`.

- [x] rebase after #189 was merged


RTSandberg (approved):
Looks good to me!

PR #200, opened 2022-08-08
Docs: Python API in the Manual

ax3l:
Document the Python API in the user-facing manual.

Sphinx Ref.:
- https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html

To do:
- [x] add each user-facing class, property & method
- [x] add docstrings and arguments

Follow-up:
- add a Python version of each example (and show py version first in docs) #202

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -68,8 +68,8 @@ Usage
    :hidden:
 
    usage/how_to_run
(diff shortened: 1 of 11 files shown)


RTSandberg (approved):
This looks really nice!

PR #203, opened 2022-08-11
Default 1D Domain Decomposition: z

ax3l:
Change the default domain decomposition from x to z.

## To Do

- [ ] check if that is smarter
- [ ] check if that is faster for most of our long-ish bunches


n01r:
Picking this up again after a very long time:

When I compare the current suite of CTests there is no difference.

For a case like `cyclotron` which seems to have a rather long bunch duration, the z-decomposed case is actually about 100ms slower. But this runs within a second.

[All_CTest_decomp_comparison.zip](https://github.com/ECP-WarpX/impactx/files/14058701/All_CTest_decomp_comparison.zip)

Perhaps I would need to run something larger, and with more particles?

PR #209, opened 2022-08-22
CMake: Simplify Example Tests

ax3l:
Add a helper function to simplify the registration of new tests.


cemitch99 (approved):
Looks great to me.

PR #225, opened 2022-08-30
Plot Scripts: Fix `--save-png` Call

ax3l:
The idea is to not pop up matplotlib windows in ctest calls.

This was lost when I refactored #209 and the current PR restores that.


n01r (approved):
Ah, yea that's better, thank you! :)

ax3l:
There is one more location in `pytest.ImpactX` that we need to guard for non-interactive runs: #227

PR #227, opened 2022-08-30
pytest Plot Scripts: Non-interactive

ax3l:
Same as #225 but for `pytest`: avoid popping up matplotlib windows in ctest calls.


n01r (commented):
Looks good!

PR #234, opened 2022-09-01
Docs: Update SoA Unit Descriptions

ax3l:
Update the description of particle SoA, mainly momentum, descriptions with respect to unit systems used.

Seen today in #162

### DIFF
--- a/src/particles/ImpactXParticleContainer.H
+++ b/src/particles/ImpactXParticleContainer.H
@@ -52,11 +52,11 @@ namespace impactx
     {
         enum
         {
-            ux,  ///< momentum in x, normalized to proper velocity
-            uy,  ///< momentum in y, normalized to proper velocity
-            pt,  ///< kinetic energy deviation, (E0-E)/mc^2
-            m_qm, ///< charge to mass ratio, in q_e/m_e (q_e/eV)
-            w,   ///< particle weight, unitless
+            ux,  ///< momentum in x, normalized to reference longitudinal momentum (unitless)
+            uy,  ///< momentum in y, normalized to reference longitudinal momentum (unitless)
+            pt,  ///< kinetic energy deviation, normalized to (E0-E)/mc^2 (unitless)
+            m_qm, ///< charge to mass ratio, normalized to q_e/m_e (q_e/eV)
+            w,   ///< particle weight, (unitless)
             nattribs ///< the number of attributes above (always last)
         };
     };


src/particles/ImpactXParticleContainer.H
```diff
@@ -52,11 +52,11 @@ namespace impactx
...
         enum
         {
-            ux,  ///< momentum in x, normalized to proper velocity
-            uy,  ///< momentum in y, normalized to proper velocity
-            pt,  ///< kinetic energy deviation, (E0-E)/mc^2
-            m_qm, ///< charge to mass ratio, in q_e/m_e (q_e/eV)
-            w,   ///< particle weight, unitless
+            ux,  ///< momentum in x, normalized to reference longitudinal momentum (unitless)
+            uy,  ///< momentum in y, normalized to reference longitudinal momentum (unitless)
+            pt,  ///< kinetic energy deviation, normalized to (E0-E)/mc^2 (unitless)
```
ax3l, line 57: @cemitch99 in `t`, where `pt` is `pz` - does `pt` still have the same normalization (energy deviation) or is actually a momentum deviation, also normalized to reference longitudinal momentum (unitless)?

cemitch99: Great question.  After the Z-T transformation, when the particle coordinates are obtained at fixed t, the component denoted "pt" is used to store the longitudinal momentum deviation pz.  (Likewise "t" is used to store the deviation in longitudinal position z.)  The stored value of pz is unitless.  I will double-check whether it is in static units (normalized by the reference momentum) or dynamic units (normalized by mc).

ax3l: Thanks!

We can probably use for these SoA attributes a similar doc syntax as we use for the SoA attributes:
https://impactx.readthedocs.io/en/latest/_static/doxyhtml/structimpactx_1_1_real_ao_s.html

Something like:
```
momentum in z, normalized to reference longitudinal momentum [unitless] (at fixed t) OR kinetic energy deviation, normalized to (E0-E)/mc^2 [unitless] (at fixed z) 
```

ax3l: Another random thought on nomenclature: why do we say "at fixed z"? Don't we mean "at fixed s"?

cemitch99: Yes, right now "at fixed s" is more appropriate.  However, in the future some soft-edged elements may use a local "z" coordinate, so this is not so straightforward.  For the documentation, for now "at fixed s" works.

ax3l: Thanks! just to clarify,
> in the future some soft-edged elements may use a local "z" coordinate

does this influence the reference particle argument interpretation or just the "local" math we will do in that element?

cemitch99: Just the "local" math.  The reference particle "z" argument has a fixed, global interpretation.

cemitch99: > @cemitch99 in `t`, where `pt` is `pz` - does `pt` still have the same normalization (energy deviation) or is actually a momentum deviation, also normalized to reference longitudinal momentum (unitless)?

The current implementation of the Z-T (and T-Z) transformations use momenta normalized by mc.  It will be better if I modify this to work with either normalization via an input flag.  Then the interpretation will be clear.

cemitch99: This is a minor change; I'll submit a PR soon.

ax3l: Does this proposed change look correct for now?
```suggestion
            ux,  ///< momentum in x, normalized to reference longitudinal momentum (unitless) (at fixed t or s)
            uy,  ///< momentum in y, normalized to reference longitudinal momentum (unitless) (at fixed t or s)
            pt,  ///< momentum in z, normalized to reference longitudinal momentum [unitless] (at fixed t) OR kinetic energy deviation, normalized to (E0-E)/mc^2 [unitless] (at fixed s)
```
or

ax3l: update: pt/pz are currently both normalized to (E0-E)/mc^2

cemitch99: Line 57 needs to be modified.  pz is normalized to the magnitude of the reference momentum (call it, say, p0), in the same way as px and py.  However, pt is normalized to p0*c.  Explicitly, pt = (E0-E)/(p0*c).

ax3l:
will be included in #242

PR #127, opened 2022-05-10
ImpactX: Move Constructor Logic

ax3l:
- [x] needs AMReX bugfix: https://github.com/AMReX-Codes/amrex/pull/2773 / https://github.com/ECP-WarpX/impactx/pull/132

### Follow-Up

- [ ] add `amr.n_cell` options and initialize another `AmrCore` object: #240

PR #240, opened 2022-09-09
Expose Control over Domain Resolution & MR Gridding

ax3l:
## Objective

Expose resolution, box and domain size control to the user.

`InitAmrCore.cpp` and `InitMeshRefinement.cpp` are the core files of this PR: Each contains about 50 new lines of logic, the rest was in `InitOneBoxPerRank` before. Everything else is docs, python bindings or testing.

## AMReX-related low-level changes

Additional constructor to our wrapper class `AmrCoreData`. This class was introduced in #127 to work around limitations of purely virtual members of `AmrCore`/`AmrMesh`. Is just transports constructor arguments and now exposes one more `AmrCore` constructor.

## Box Init strategies

The old default, `InitOneBoxPerRank` is still present and will be used when the user does not specify any cell resolution. We use this to simply get a box per MPI rank so we can init particles.

If the user does specify `amr.n_cells`, we use the regular AMReX init logic now, as in WarpX.

This code is in `InitAmrCore`.

## Tests

There is currently no documented AMReX way to change the resolution of a simulation (number of cells) after `AmrCore` was already initialized. No problem if you work with inputs files, but if you work with Python that becomes tricky.

We hack this quickly by using:
```py
    pp_amr = amrex.ParmParse("amr")
    pp_amr.addarr("n_cell", [16, 24, 32])

    sim = impactx.ImpactX()
```
before running `sim = ImpactX();`. Later on, we will either with @WeiqunZhang's help find a way to change resolution after `AmrCore` was initialized or we work-around the AMReX limitation and wrap the ImpactX object away to delay construction up to evolve (or so).

- [x] Python bindings: as good as we can for now (using AMReX ParmParse to inject prior to sim object)
- [x] Tests


src/python/ImpactX.cpp
```diff
@@ -53,6 +54,85 @@ void init_ImpactX(py::module& m)
...
+                /*
+                amrex::ParmParse pp_amr("amr");
+                amrex::Vector<int> const n_cell_v(n_cell.begin(), n_cell.end());
+                pp_amr.addarr("n_cell", n_cell_v);
+
+                int const max_level = ix.maxLevel();
+                for (int lev=0; lev<=max_level; lev++) {
+                    ix.ClearLevel(lev);
+                    // TODO: more...
+                }
```
ax3l, line 75: Question for @WeiqunZhang how to change number of cells in an `AmrCore` object at runtime :)

docs/source/usage/parameters.rst
```diff
@@ -33,36 +33,38 @@ Overall simulation parameters
...
-    Note: in development; currently allowed value: ``2 2 2``.
+* ``geometry.dynamic_size`` (``boolean``) optional (default: ``true`` for dynamic)
+    Use dynamic (``true``) resizing of the field mesh, via ``geometry.prob_relative``, or static sizing (``false``), via ``geometry.prob_lo``/``geometry.prob_hi``.
+
+* ``geometry.prob_relative`` (``float``, unitless) optional (default: ``0.1``)
+    By default, we dynamically extract the minimum and maximum of the particle positions in the beam.
+    The field mesh is expanded beyond the physical extent of particles by this factor.
+
+* ``geometry.prob_lo`` and ``geometry.prob_hi`` (3 floats, in meters) optional (required if ``geometry.dynamic_size`` is ``false``)
+    The extent of the full simulation domain.
```
```suggestion
    The extent of the full simulation domain relative to the reference particle position.
```

docs/source/usage/python.rst
```diff
@@ -23,6 +23,25 @@ General
 
       :param int order: B-spline order ``1``, ``2``, or ``3``
 
+   .. py:property:: n_cell
+
+      The number of grid points along each direction on the coarsest level.
+
+   .. py:property:: domain
+
+      The physical extent of the full simulation domain in meters.
```
```suggestion
      The physical extent of the full simulation domain, relative to the reference particle position, in meters.
```

src/initialization/InitAmrCore.H
```diff
@@ -0,0 +1,56 @@
...
+     * we do not need space charge or load balancing.
+     *
+     * @return simulation_geometry the geometry (topology) of the simulation;
+     *         amr_info contains information on mesh-refinement and box/grid blocks
+     */
+    AmrCoreData
+    one_box_per_rank ();
+
+    //! the fake domain size (lower end), used to indicate that the user made no choice for it
+    constexpr amrex::RealVect undefined_geometry_prob_lo(0.0);
```
ax3l: Seems more logical as an fake position relative to ref particle (even if overwritten later)
```suggestion
    constexpr amrex::RealVect undefined_geometry_prob_lo(-1.0);
```

src/python/ImpactX.cpp
```diff
@@ -53,6 +54,85 @@ void init_ImpactX(py::module& m)
...
+                amrex::Vector<amrex::Real> const prob_lo_v({rb.lo()[0], rb.lo()[1], rb.lo()[2]});
+                amrex::Vector<amrex::Real> const prob_hi_v({rb.hi()[0], rb.hi()[1], rb.hi()[2]});
+                pp_geometry.addarr("prob_lo", prob_lo_v);
+                pp_geometry.addarr("prob_hi", prob_hi_v);
+
+                pp_geometry.add("dynamic_size", false);
+
+                ix.ResizeMesh();
+            },
+            "The physical extent of the full simulation domain in meters."
```
ax3l:
```suggestion
            "The physical extent of the full simulation domain, relative to the reference particle position, in meters."
```

tests/python/test_charge_deposition.py
```diff
@@ -44,14 +66,21 @@ def test_charge_deposition(save_png=True):
         rbx = amrex.RealBox(bx, dr, gm.ProbLo())
 
         arr = rho.array(mfi)
-        arr_np = np.array(arr, copy=False)
+        arr_np = np.array(arr, copy=False)  # indices: comp, z, y, x
```
ax3l, line 69: should add some assert here to ensure cell res change worked

tests/python/test_charge_deposition.py
```diff
@@ -14,17 +21,30 @@ def test_charge_deposition(save_png=True):
...
     sim.init_beam_distribution_from_inputs()
     sim.init_lattice_elements_from_inputs()
+    print(f"sim.n_cell={sim.n_cell}")
+    print(f"sim.domain={sim.domain}")
 
     sim.evolve()
+    print(f"sim.n_cell={sim.n_cell}")
+    print(f"sim.domain={sim.domain}")
 
     rho = sim.rho(lev=0)
```
ax3l: should add some assert on the size of rho here to ensure cell number change worked

src/initialization/InitAmrCore.cpp
```diff
@@ -0,0 +1,135 @@
...
+} // namespace details
+
+    AmrCoreData
+    init_amr_core ()
+    {
+        if (!amrex::Initialized())
+        {
+            default_init_AMReX();
+            // note: due to global state, it would be too early to use amrex::Abort() here
+            //throw std::runtime_error("AMReX must be initialized before ImpactX simulation can be constructed.");
```
RemiLehe: Did you want to implement this as part of this PR?

src/initialization/InitAmrCore.cpp
```diff
@@ -0,0 +1,135 @@
...
+    init_physical_domain ()
+    {
+        amrex::ParmParse pp_geometry("geometry");
+        bool dynamic_size = true;
+        pp_geometry.queryAdd("dynamic_size", dynamic_size);
+        amrex::Vector<amrex::Real> prob_lo(undefined_geometry_prob_lo.begin(), undefined_geometry_prob_lo.end());
+        amrex::Vector<amrex::Real> prob_hi(undefined_geometry_prob_hi.begin(), undefined_geometry_prob_hi.end());
+        if (dynamic_size)
+        {
+            pp_geometry.queryAdd("prob_lo", prob_lo);
```
RemiLehe, line 54: This could be a bit misleading, no?
It is seems that, in the `dynamic_size` mode, it is reading the `prob_lo` and `prob_hi`, but this actually has no influence on the simulation, because `prob_lo` and `prob_hi` will be redefined anyway by `ResizeMesh`.

src/initialization/InitMeshRefinement.cpp
```diff
@@ -116,13 +113,42 @@ namespace impactx
...
+            rb = {
+                {x_min - frac * (x_max - x_min), y_min - frac * (y_max - y_min),
+                 z_min - frac * (z_max - z_min)}, // Low bound
+                {x_max + frac * (x_max - x_min), y_max + frac * (y_max - y_min),
+                 z_max + frac * (z_max - z_min)}  // High bound
+            };
+        }
+        else
+        {
+            amrex::Vector<amrex::Real> prob_lo;
```
RemiLehe, line 139: Is this needed here?
It seems that, when `dynamic_size` is `false`, the code will still call `amrex::Geometry::ResetDefaultProbDomain` and `amrex::AmrMesh::SetGeometry` (even though `prob_lo` and `prob_hi` are not changing.)

Would it make sense to instead call `return` early in this function if `dynamic_size` is `false`?

RemiLehe (commented):
I had a few questions in the code here. Would you have time answer them?

src/initialization/InitAmrCore.cpp
ax3l: I mean to add a default here in case someone reads the value before `evolve()` is started (I think some internal routines for particle init do). Maybe `addarr` is less confusing?

Update: Actually, I used `queryAdd` because in Python a user will usually decide to switch to non-dynamic _after_ `ImpactX` object creation, thus I did not want to overwrite their inputs.

src/initialization/InitAmrCore.cpp
ax3l: Oh, that's old code from `InitOneBoxPerRank.cpp` and should be removed.

ax3l:
```suggestion
```

src/initialization/InitAmrCore.cpp
```diff
@@ -0,0 +1,135 @@
...
+        }
+        amrex::RealBox rb(prob_lo.data(), prob_hi.data());
+        return rb;
+    }
+} // namespace details
+
+    AmrCoreData
+    init_amr_core ()
+    {
+        if (!amrex::Initialized())
```
ax3l, line 71:
```suggestion
        // note: due to global state, it would be too early to use amrex::Abort() in these functions
        if (!amrex::Initialized())
```

src/initialization/InitMeshRefinement.cpp
ax3l: Yes, that is intentional in case someone changed from dynamic to static size after part of a simulation. or if someone changed the RealBox of the sim domain from Python, see `src/python/ImpactX.cpp` line 103.

`amrex::Geometry::ResetDefaultProbDomain` and `amrex::AmrMesh::SetGeometry` are cheap functions.

src/initialization/InitAmrCore.cpp
ax3l:
```suggestion
            // read and conditionally overwrite problem domain size:
            //   users might switch to !dynamic_size after init and we don't
            //   want to overwrite their inputs
            pp_geometry.queryAdd("prob_lo", prob_lo);
```

src/initialization/InitMeshRefinement.cpp
ax3l:
```suggestion
            // note: we read and set the size again because an interactive /
            //       Python user might have changed it between steps
            amrex::Vector<amrex::Real> prob_lo;
```

cemitch99 (approved):
This looks okay to me.

RemiLehe (approved):

src/python/ImpactX.cpp
ax3l: From what we chatted to far on Slack, we need to update the data members of AmrCore, AmrMesh, and AmrParGDB.
All MultiFabs need to be updated separately and are not automatically changed. Not sure what we need to update in particle containers, e.g., `parGDB` related.

PR #241, opened 2022-09-09
Enable OpenMP in particle push and coordinate transformation routines.

### LINKED ISSUE #195: OpenMP Support
When compiling with the OpenMP backend, we do not yet use particle tiling it seems.
With our current distribution strategy, we start with one block per device.

atmyers:
Close #195 

## To Do

- [x] turn tiling on
- [x] set a reasonable default tile size when OMP is selected
- [x] rebase and also OpenMP-accelerate new kernels from #162

### DIFF
--- a/src/particles/ImpactXParticleContainer.cpp
+++ b/src/particles/ImpactXParticleContainer.cpp
@@ -25,7 +25,7 @@
 namespace impactx
 {
     bool do_omp_dynamic () {
-        bool do_dynamic = false;
+        bool do_dynamic = true;
         amrex::ParmParse pp_impactx("impactx");
         pp_impactx.query("do_dynamic_scheduling", do_dynamic);
         return do_dynamic;
--- a/src/particles/ImpactXParticleContainer.cpp
+++ b/src/particles/ImpactXParticleContainer.cpp
@@ -24,6 +24,42 @@
 
 namespace impactx
 {
+    ParIter::ParIter (ContainerType& pc, int level)
+        : amrex::ParIter<0, 0, RealSoA::nattribs, IntSoA::nattribs>(pc, level,
+              amrex::MFItInfo().SetDynamic( []() -> bool {
+                  bool do_dynamic = false;
+                  amrex::ParmParse pp_impactx("impactx");
+                  pp_impactx.query("do_dynamic_scheduling", do_dynamic);
+                  return do_dynamic; }()))
+    {}
+
+    ParIter::ParIter (ContainerType& pc, int level, amrex::MFItInfo& info)
+        : amrex::ParIter<0, 0, RealSoA::nattribs, IntSoA::nattribs>(pc, level,
+              info.SetDynamic( []() -> bool {
+                  bool do_dynamic = false;
+                  amrex::ParmParse pp_impactx("impactx");
+                  pp_impactx.query("do_dynamic_scheduling", do_dynamic);
+                  return do_dynamic; }()))
+    {}
+
+    ParConstIter::ParConstIter (ContainerType& pc, int level)
+        : amrex::ParConstIter<0, 0, RealSoA::nattribs, IntSoA::nattribs>(pc, level,
+              amrex::MFItInfo().SetDynamic( []() -> bool {
+                  bool do_dynamic = false;
+                  amrex::ParmParse pp_impactx("impactx");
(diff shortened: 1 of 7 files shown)


ax3l:
Thanks! Do we need to change some defaults, like the dynamic scheduling default we do in WarpX?

atmyers:
Yes, at minimum we should set turn tiling on and set a reasonable default tile size when OMP is selected - I will update.

ax3l:
@atmyers thanks! We also merged #162 now. Can you please rebase and OpenMP-enable those routines as well? :)

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -23,6 +23,29 @@
 
 namespace impactx
 {
+    bool do_omp_dynamic () {
+        bool do_dynamic = false;
+        amrex::ParmParse pp_impactx("impactx");
+        pp_impactx.query("do_dynamic_scheduling", do_dynamic);
```
ax3l, line 30: Status: as of Andrew's benchmarks yesterday, using dynamic scheduling was was slower than w/o
Update: oops, that compared Debug & Release at the same time.

atmyers:
Here are the current performance results on the expanding beam test:

```
dynamic_1.out:TinyProfiler total time across processes [min...avg...max]: 1.35 ... 1.35 ... 1.35
dynamic_2.out:TinyProfiler total time across processes [min...avg...max]: 0.8144 ... 0.8144 ... 0.8144
dynamic_4.out:TinyProfiler total time across processes [min...avg...max]: 0.5541 ... 0.5541 ... 0.5541
static_1.out:TinyProfiler total time across processes [min...avg...max]: 1.353 ... 1.353 ... 1.353
static_2.out:TinyProfiler total time across processes [min...avg...max]: 0.8157 ... 0.8157 ... 0.8157
static_4.out:TinyProfiler total time across processes [min...avg...max]: 0.5614 ... 0.5614 ... 0.5614
```

src/particles/ImpactXParticleContainer.cpp
atmyers: See current results below.

ax3l:
Thank you! Can you perform a little test on a single CPU package comparing MPI w/ 1 OMP thread vs. all OMP threads?

ax3l:
Can you please also update MFIter loops to be OpenMP accelerated?

atmyers:
Here is a comparison between pure MPI and pure OMP on the expanding beam test, with diags disabled:

```
MPI:
1 ranks, 1 threads: 1.357
2 ranks, 1 threads: 0.9569
4 ranks, 1 threads: 0.6122

OMP:
1 ranks, 1 threads: 1.353
1 ranks, 2 threads: 0.8035
1 ranks, 4 threads: 0.6009
```

ax3l:
Thank you so much, that's awesome :tada: 
Quickly `grep`-ing the code, I think there is one more `MFIter` loop in `src/initialization/InitMeshRefinement.cpp` - can that one be accelerated as well?

ax3l:
(I'll merge this real quick because I want to rebase #245, but pls feel free to post a follow-up l8r)

ax3l:
Oh, must have been an old version. Only one MFIter in the code as of `development`.

src/particles/spacecharge/ForceFromSelfFields.cpp
```diff
@@ -40,6 +40,9 @@ namespace impactx::spacecharge
             space_charge_field.at(lev).at("y").setVal(0.);
             space_charge_field.at(lev).at("z").setVal(0.);
 
+#ifdef AMREX_USE_OMP
+#pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
+#endif
             for (amrex::MFIter mfi(phi.at(lev)); mfi.isValid(); ++mfi) {
```
ax3l, line 46: @WeiqunZhang commented:
we can also do tiling on CPU for MFIter loops (not yet done here).

Let's investigate if this helps us here and potentially make it a user option.

Example:
https://github.com/ECP-WarpX/WarpX/blob/3fe406c9701f61e07b23f7123cf0a7bad492c6dc/Source/Diagnostics/ComputeDiagFunctors/BackTransformFunctor.cpp#L110

PR #243, opened 2022-09-10
Change tasks-per-node to tasks-per-gpu

n01r:
On Perlmutter, the previous option led to warnings and crashes as described in https://github.com/ECP-WarpX/WarpX/issues/3356. In [WarpX PR#3375](https://github.com/ECP-WarpX/WarpX/pull/3375) this fix was first proposed.


ax3l (approved):
Thanks a lot :)

PR #248, opened 2022-09-16
Examples: Fix Absolute Tolerance in Analysis (Part 2)

ax3l:
In `np.allclose`, the absolute and relative tolerances are added. We meant to disable the term with the absolute one, so this should be zero and not one.

Ref.: https://numpy.org/doc/stable/reference/generated/numpy.allclose.html

PR #250, opened 2022-09-16
Fix relative tolerance and targets for a few benchmark problems.

cemitch99:
This is a replacement for:  "Examples: Fix Absolute Tolerance in Analysis (Part 2) #248".  Changes are made to the analysis scripts for the following benchmark problems:

1) cfchannel
2) kurth
3) kvdist
4) iotalens
5) iotalattice

The relative tolerance was increased from 1% to 1.5%.  In some cases, target values were updated.


ax3l (approved):
:tada:

PR #251, opened 2022-09-16
Charge Deposition: Fix Sign

ax3l:
This fixes the charge deposition. Electrons should deposit negative charge :-)

### DIFF
--- a/src/particles/ChargeDeposition.cpp
+++ b/src/particles/ChargeDeposition.cpp
@@ -66,8 +66,8 @@ namespace impactx
                     //auto const rel_ref_ratio = ref_ratio.at(depos_lev) / ref_ratio.at(lev);
                     amrex::ignore_unused(ref_ratio);
 
-                    amrex::ParticleReal const q_e = 1.60217662e-19;  // TODO move out
-                    amrex::ParticleReal const charge = q_e;
+                    // in SI [C]
+                    amrex::ParticleReal const charge = m_refpart.charge;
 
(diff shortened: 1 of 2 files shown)


cemitch99 (approved):
Great!

RemiLehe (approved):

RemiLehe:
@ax3l It seems that you will need to modify the test to replace:
```
assert math.isclose(beam_charge, 1.0e-9)
``` 
with 
```
assert math.isclose(beam_charge, -1.0e-9)
```

ax3l:
correct, wanted to quickly see if this fails now as intended.

It does, but I also need to increase the tolerance a little now: `beam_charge=-1.0000000087381137e-09`

cemitch99:
> @ax3l It seems that you will need to modify the test to replace:
> 
> ```
> assert math.isclose(beam_charge, 1.0e-9)
> ```
> 
> with
> 
> ```
> assert math.isclose(beam_charge, -1.0e-9)
> ```

This is a good point.  In Impact-Z, the sign problem is avoided by treating all charges as positive.  (So electrons are treated as positrons.) This makes life a bit easier, and it does not affect tracking in the external fields because the field strength is given in a normalized form (scaled by signed magnetic rigidity p/q).  Either option is fine, as long as the sign of the pushed single-particle charge is the same as the sign of the deposited charge of the bunch.

ax3l:
I like that we have both, would allow us to do some (short beamline) tests with partly neutral beams as come out from some plasma sources.

PR #252, opened 2022-09-16
Replace Hard-Coded Constants with ABLASTR

ax3l:
Replace hard-coded constants with constants from ABLASTR. This provides us with a single source of truth and avoids nifty issues with slightly-off numerical values depending on the precision used in multiple locations.

- [x] depends on https://github.com/ECP-WarpX/WarpX/pull/3405

### DIFF
--- a/cmake/dependencies/ABLASTR.cmake
+++ b/cmake/dependencies/ABLASTR.cmake
@@ -134,7 +134,7 @@ set(ImpactX_ablastr_src ""
 set(ImpactX_ablastr_repo "https://github.com/ECP-WarpX/WarpX.git"
     CACHE STRING
     "Repository URI to pull and build ABLASTR from if(ImpactX_ablastr_internal)")
-set(ImpactX_ablastr_branch "425d22a5b596eb1ceb8e5a07d1ed3766870f9d9e"
+set(ImpactX_ablastr_branch "5deed41b5788a7f59da1cdc55cb30b2cea781441"
     CACHE STRING
     "Repository branch for ImpactX_ablastr_repo if(ImpactX_ablastr_internal)")
 
--- a/src/particles/ReferenceParticle.H
+++ b/src/particles/ReferenceParticle.H
@@ -91,7 +91,7 @@ namespace impactx
         {
             using namespace amrex::literals;
 
-            constexpr amrex::ParticleReal inv_MeV_invc2 = ablastr::constant::SI::MeV_invc2;
+            constexpr amrex::ParticleReal inv_MeV_invc2 = 1.0_prt /  ablastr::constant::SI::MeV_invc2;
             return amrex::ParticleReal(mass * inv_MeV_invc2);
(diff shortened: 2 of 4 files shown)


src/particles/ReferenceParticle.H
```diff
@@ -89,8 +91,8 @@ namespace impactx
         {
             using namespace amrex::literals;
 
-            constexpr double MeVc2_kg = 1.78266192e-30;
-            return amrex::ParticleReal(mass / MeVc2_kg);
+            constexpr amrex::ParticleReal inv_MeV_invc2 = ablastr::constant::SI::MeV_invc2;
```
ax3l:
```suggestion
            constexpr amrex::ParticleReal inv_MeV_invc2 = 1.0_prt /  ablastr::constant::SI::MeV_invc2;
```

src/particles/ReferenceParticle.H
```diff
@@ -165,8 +166,8 @@ namespace impactx
         {
             using namespace amrex::literals;
 
-            constexpr double qe = 1.602176634e-19;
-            return amrex::ParticleReal(charge / qe);
+            constexpr amrex::ParticleReal inv_qe = ablastr::constant::SI::q_e;
```
ax3l:
```suggestion
            constexpr amrex::ParticleReal inv_qe = 1.0_prt / ablastr::constant::SI::q_e;
```

cmake/dependencies/ABLASTR.cmake
```diff
@@ -134,7 +134,7 @@ set(ImpactX_ablastr_src ""
 set(ImpactX_ablastr_repo "https://github.com/ECP-WarpX/WarpX.git"
     CACHE STRING
     "Repository URI to pull and build ABLASTR from if(ImpactX_ablastr_internal)")
-set(ImpactX_ablastr_branch "425d22a5b596eb1ceb8e5a07d1ed3766870f9d9e"
+set(ImpactX_ablastr_branch "5deed41b5788a7f59da1cdc55cb30b2cea781441"
```
n01r, line 137: Is this something that always has to be set manually?

n01r (approved):

cmake/dependencies/ABLASTR.cmake
ax3l: yes, this is the location where we maintain the default downloaded version of ABLASTR, AMReX et al.
Users can rely on those defaults.

In Package managers, we overwrite those. For that reason, these variables are [CMake `CACHE` variables](https://cliutils.gitlab.io/modern-cmake/chapters/basics/variables.html) (options).

ax3l: For such maintenance workflows, see here:
- https://warpx.readthedocs.io/en/latest/maintenance/release.html
- https://impactx.readthedocs.io/en/latest/maintenance/release.html (to be updated)

PR #258, opened 2022-09-20
Fix: Docs for Sigmas of Distributions

ax3l:
Fix the parameter docs for sigma of positions and momentum for beam distributions.


cemitch99 (approved):
This looks good.

PR #260, opened 2022-09-21
Python: All Sim Options As Properties

ax3l:
Refactor `ImpactX` setters to be properties.
This is more Pythonic and also allows us to query values of options instead of only setting them.

### DIFF
--- a/examples/cfchannel/run_cfchannel.py
+++ b/examples/cfchannel/run_cfchannel.py
@@ -12,10 +12,10 @@ from impactx import ImpactX, RefPart, distribution, elements
 sim = ImpactX()
 
 # set numerical parameters and IO control
-sim.set_particle_shape(2)  # B-spline order
-sim.set_space_charge(False)
-# sim.set_diagnostics(False)  # benchmarking
-sim.set_slice_step_diagnostics(True)
+sim.particle_shape = 2  # B-spline order
+sim.space_charg = False
+# sim.diagnostics = False  # benchmarking
+sim.slice_step_diagnostics = True
 
 # domain decomposition & space charge mesh
 sim.init_grids()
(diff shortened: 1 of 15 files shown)


examples/cfchannel/run_cfchannel.py
```diff
@@ -12,10 +12,10 @@
 sim = ImpactX()
 
 # set numerical parameters and IO control
-sim.set_particle_shape(2)  # B-spline order
-sim.set_space_charge(False)
-# sim.set_diagnostics(False)  # benchmarking
-sim.set_slice_step_diagnostics(True)
+sim.particle_shape = 2  # B-spline order
+sim.space_charg = False
```
n01r:
```suggestion
sim.space_charge = False
```

src/python/ImpactX.cpp
```diff
@@ -133,7 +133,15 @@ void init_ImpactX(py::module& m)
               "Use dynamic (``true``) resizing of the field mesh or static sizing (``false``)."
         )
 
-        .def("set_particle_shape",
+        .def_property("particle_shape",
+            [](ImpactX & /* ix */) {
+                amrex::ParmParse pp_ago("algo");
```
n01r:
```suggestion
                amrex::ParmParse pp_algo("algo");
```

src/python/ImpactX.cpp
```diff
@@ -133,7 +133,15 @@ void init_ImpactX(py::module& m)
...
+                int order = 0;
+                bool const has_shape = pp_ago.query("particle_shape", order);
```
n01r:
```suggestion
                bool const has_shape = pp_algo.query("particle_shape", order);
```

PR #223, opened 2022-08-29
Add Unused Parameter Warnings

n01r:
Adds warnings about unused parameters like in WarpX as requested in #120 

- [x] Can you please check if `amrex.abort_on_unused_inputs = 1` works already as well?
- [x] This one needs to be documented in both WarpX and AMReX docs, please.
- [x] We need to please document `warpx.abort_on_warning_threshold` (or equivalent) in ImpactX.
- [x] implemented it like in WarpX but it does not abort, yet, when I do some warning logger testing as described in the WarpX docs
- [x] move impl to separate `.cpp` file
- [x] add Python control
- [x] add docs for Python control
- ~~[ ] add Python tests for unused inputs and for warning in `tests/python/test_impactx.py`~~ skipped for follow-up since more testing is needed

PR #265, opened 2022-09-28
Python: Fix Interactive Inspection

ax3l:
When autocompleting the `ImpactX()` Python class, IPython uses [Jedi](https://jedi.readthedocs.io) to look into the getters of [properties](https://jedi.readthedocs.io/en/latest/docs/features.html#supported-python-features).
> Jedi does execute properties and in general is not very careful to avoid code execution.

By default, the `ParmParse::get()` we use would throw an `amrex::Abort()`, which exits the interpreter.

We could and should use `amrex.throw_exception = 1` and `amrex.signal_handling = 0` to make this a `std::runtime_error`. Unfortunately, this is still not sufficient to catch it, because C++ exception handling across shared libraries needs extra care.

Thus, we just do a helper function, save some lines of code, and throw directly from ImpactX's Python module.

Thanks to @n01r for spotting this originally.
Follow-up to #260 and #223

### DIFF
--- a/src/python/ImpactX.cpp
+++ b/src/python/ImpactX.cpp
@@ -20,6 +20,7 @@
 #if defined(AMREX_DEBUG) || defined(DEBUG)
 #   include <cstdio>
 #endif
+#include <string>
 
 
 namespace py = pybind11;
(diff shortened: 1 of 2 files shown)


src/python/ImpactX.cpp
```diff
@@ -29,6 +30,34 @@ namespace impactx {
...
+     * This handles most common throw exception logic in ImpactX instead of
+     * going over library boundaries via amrex::Abort().
+     *
+     * @tparam T type of the amrex::ParmParse entry
+     * @param prefix the prefix, e.g., "impactx" or "amr"
+     * @param name the actual key of the entry, e.g., "particle_shape"
+     * @return the queried value (or throws if not found)
+     */
+    template< typename T>
+    auto get_or_throw(std::string const prefix, std::string const name)
```
ax3l, line 49: We can check in the future if we find a reasonable location to set default parameters without setting them multiple times throughout the code.

Maybe we need a `ReadParameters`-like function called `set_default_parameters` after all... :-S

src/python/ImpactX.cpp
```diff
@@ -29,7 +30,34 @@ namespace impactx {
     struct Config {};
 }
 
-void init_ImpactX(py::module& m)
+namespace detail
+{
+    /** Helper Function for Property Getters
```
ax3l, line 35: We can check in the future if we find a reasonable location to set default parameters without setting them multiple times throughout the code.

Maybe we need a `ReadParameters`-like function called `set_default_parameters` after all... :-S

n01r (approved):
Looks great! :)

n01r:
Also doesn't crash on Mac, anymore. ✨

RTSandberg (commented):
This looks good to me and works fine on my Mac

PR #270, opened 2022-10-06
Examples: Simplify Finalize

ax3l:
We don't really need to call this, the order should be done automatically and correctly at shutdown.

Exception: MPI runs that need to call finalize:
```
--------------------------------------------------------------------------
mpiexec has exited due to process rank 1 with PID 0 on
node fv-az205-415 exiting improperly. There are three reasons this could occur:

1. this process did not call "init" before exiting, but others in
the job did. This can cause a job to hang indefinitely while it waits
for all processes to call "init". By rule, if one process calls "init",
then ALL processes must call "init" prior to termination.

2. this process called "init", but exited without calling "finalize".
By rule, all processes that call "init" MUST call "finalize" prior to
exiting or it will be considered an "abnormal termination"

3. this process called "MPI_Abort" or "orte_abort" and the mca parameter
orte_create_session_dirs is set to false. In this case, the run-time cannot
detect that the abort call was an abnormal termination. Hence, the only
error message you will receive is this one.

This may have caused other processes in the application to be
terminated by signals sent by mpiexec (as reported here).

You can avoid this message by specifying -quiet on the mpiexec command line.
--------------------------------------------------------------------------
```

### DIFF
--- a/examples/cfchannel/run_cfchannel.py
+++ b/examples/cfchannel/run_cfchannel.py
@@ -6,7 +6,6 @@
 #
 # -*- coding: utf-8 -*-
 
-import amrex
 from impactx import ImpactX, RefPart, distribution, elements
 
 sim = ImpactX()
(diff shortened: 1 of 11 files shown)


ax3l:
I think I want to keep the clean shutdown, because only then the summary table is printed and MPI_Finalize() called for MPI runs. And in principle, every example should stay runnable with/without MPI :)

PR #272, opened 2022-10-17
SBend: Sin/Cos Upfront

ax3l:
Calculate sin/cos of theta upfront, since those are expensive calls.

Turns out, the compiler is too smart and propably did this already, thus no performance difference (on my quick CPU test).

Reason for change: preparation for a combined `sincos` call that we can use on some platforms like CUDA GPUs. Also GNU and Intel have such functions for CPUs, which can be added by the compiler. The new pattern should assist recognition for this as well.


ax3l:
can add a follow up once https://github.com/AMReX-Codes/amrex/pull/3008 was merged :)

WeiqunZhang:
Yes, the compiler is pretty smart.
https://community.intel.com/t5/Intel-C-Compiler/Where-is-the-sincos-function/td-p/1000409

PR #208, opened 2022-08-20
CI: Windows + Python

ax3l:
Run Python scripts in Windows CI tests.

- [x] depends on #209
- [x] depends on https://github.com/AMReX-Codes/amrex/pull/3013
- [x] depends on https://github.com/ECP-WarpX/WarpX/pull/3503
- [x] depends on https://github.com/AMReX-Codes/pyamrex/pull/91

PR #274, opened 2022-11-02
CMake: Lib & Python Consistency

ax3l:
Fix the logic of forcing the lib build for Python bindings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,6 +90,11 @@ set_default_build_type("Release")
 # (also know as "link-time optimization" or "whole program optimization")
 option(ImpactX_IPO "Compile ImpactX with interprocedural optimization (will take more time)" OFF)
 
+# enforce consistency of dependent options
+if(ImpactX_PYTHON)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -123,9 +128,6 @@ endif()
 
 # Targets #####################################################################
 #
-if(ImpactX_PYTHON)
```
ax3l, line 126: logic in `cmake/dependencies/ABLASTR.cmake` depends on a correctly set `ImpactX_LIB`, so we need to move it up before including that file

cmake/ImpactXFunctions.cmake
```diff
@@ -287,7 +287,7 @@ function(impactx_set_binary_name)
         )
     endif()
     if(ImpactX_LIB)
-        if(WIN32)
+        if(WIN32)  # TODO: handle static lib extensions
```
ax3l, line 290: Probably the reason for the issue in #208, will be fixed therein

ax3l:
MSVC build issue will probably be fixed with #275. Will merge that first.

PR #277, opened 2022-11-05
Fix indentation in nested lists causing bold font

n01r:
Some indentation caused the first line of an item to be bold where it was perhaps not intended. This was fixed and also sentences have been put on a newline each while broken sentences were put on the same line. The only exception are equations in words where the equal sign and right hand side of the equation were kept on a new line as a sublist for better readability.


ax3l (approved):
Thank you, good catch! :)

PR #279, opened 2022-12-02
Added refinements to the expanding beam benchmark.

cemitch99:
Modified the expanding beam space charge benchmark as follows:

1) Distribution type was changed from zero-emittance "waterbag" to zero-emittance "kurth6d", to correctly give a uniformly-distributed 3D ellipsoid.  Note that the test is passed using either distribution, but using "kurth6d" yields a reduction in emittance growth.

2) The resolution was increased from 10^4 to 10^5 particles, with a corresponding increase in mesh resolution and number of kicks, in order to increase the precision of the test (optional).


examples/expanding_beam/analysis_expanding.py
```diff
@@ -58,7 +58,7 @@ def read_all_files(file_pattern):
 final = read_all_files("diags/beam_final.*")
 
 # compare number of particles
-num_particles = 10000
+num_particles = 100000
```
ax3l: We could check if with 10^4 or 10^5 particles we can reduce `atol = 1.0e-8` in line 113

ax3l:
> The resolution was increased from 10^4 to 10^5 particles, with a corresponding increase in mesh resolution and number of kicks, in order to increase the precision of the test (optional).

That works and also just needs about 4-5 seconds on my laptop to run on two cores in `Release` mode.
But, if possible, I would love to run CI in `Debug` mode where it takes 45sec with that many particles. In Debug mode, we can run more costly checks and asserts and thus I would propose to stay with 1e4 particles for the test for now.

ax3l (approved):

PR #293, opened 2023-01-05
Docs: Perlmutter CTest & Python

ax3l:
Update how to run interactive tests on Perlmutter (NERSC), including runs with Python.


cemitch99 (approved):
This worked for me.

PR #302, opened 2023-01-17
Add ideal solenoid element

cemitch99:
Added an ideal, hard-edge solenoid element similar to that used in MAD-X.

A benchmark example and documentation are included.

A follow-up PR will:
- [ ] create a soft-edge solenoid element using on-axis Bz field, similar to IMPACT-Z #309 
- [x] update the Mad-X parser #305


ax3l:
Thanks a lot for this PR, @cemitch99.

Commits added to this PR by me (`git fetch; git pull`):
- [x] Python bindings for `Sol`
- [x] Documentation

In a follow-up, we still need to update the Mad-X parser in:
- `src/python/impactx/`
  - `MADXParser.py`
  - `madx_to_impactx.py`

to add the corresponding hard-edge solenoid name recognition.

src/particles/elements/Sol.H
```diff
@@ -0,0 +1,165 @@
...
+
+#include <AMReX_Extension.H>
+#include <AMReX_REAL.H>
+
+#include <cmath>
+
+
+namespace impactx
+{
+    struct Sol
```
ax3l, line 25: Do we want to name this class `Solenoid` as well?

src/particles/elements/Sol.H
```diff
@@ -0,0 +1,165 @@
...
+    : public elements::BeamOptic<Sol>,
+      public elements::Thick
+    {
+        static constexpr auto name = "Sol";
```
ax3l, line 29: Is this the Mad-X equivalent name or shall we call it `Solenoid` or `IdealSolenoid`?

src/particles/elements/Sol.H
cemitch99: Sure; I think either name is fine.  The abbreviation Sol is in common use for a solenoid, so I don't think there is much chance for confusion.

src/particles/elements/Sol.H
ax3l: @cemitch99 on Slack:
> MAD-X by default has only an ideal solenoid element, denoted by SOLENOID (Section 10.9, p. 78 of the user guide).
>
> It is possible to access some features of the tracking code PTC within MAD-X, which probably allows for a soft-edge model, but PTC is really a separate code, and not standard in MAD.  A soft-edge solenoid appears in MaryLie, IMPACT-Z, and MaryLie/IMPACT.

examples/solenoid/run_solenoid.py
```diff
@@ -0,0 +1,52 @@
...
+    sigmaY=2.205510139392e-3,
+    sigmaT=1.0e-3,
+    sigmaPx=6.41218345413e-4,
+    sigmaPy=9.06819680526e-4,
+    sigmaPt=1.0e-3,
+)
+sim.add_particles(bunch_charge_C, distr, npart)
+
+# design the accelerator lattice
+sim.lattice.append(elements.ConstF(ds=3.820395, ks=0.8223219329893234))
```
ax3l:
```suggestion
sim.lattice.append(elements.Sol(ds=3.820395, ks=0.8223219329893234))
```

ax3l (approved):
Thank you :rocket: :sparkles:

PR #304, opened 2023-01-18
Add user input of RF cavity Fourier coefficients

cemitch99:
1) Creates a new struct "RF_field_data" to contain user-provided RF field data, in the form of Fourier coefficients of the on-axis value of Ez.  (Default values are those of the 9-cell TESLA cavity, just as before this PR.)

2) Initializes a struct of this type called "ezdata" to contain field values for each RF cavity element.

2) Modifies the RF cavity element particle push to make use of Fourier coefficients contained in  the struct "ezdata".

To do:

- [x] Add function for parsing user-provided input file containing these Fourier coefficients.
- [x] GPU support for the coefficient array: copy host-to-device the input
- [x] Python equivalent


src/initialization/InitElement.cpp
```diff
@@ -112,7 +114,8 @@ namespace impactx
                 pp_element.get("phase", phase);
                 pp_element.queryAdd("mapsteps", mapsteps);
                 pp_element.queryAdd("nslice", nslice);
-                m_lattice.emplace_back( RFCavity(ds, escale, freq, phase, mapsteps, nslice) );
+                pp_element.queryAdd("file", field_file);
+                m_lattice.emplace_back( RFCavity(ds, escale, freq, phase, mapsteps, nslice, ezdata) );
```
```C++
        std::vector<double> rf_coeff;
        pp_element.queryarr("coefficients", rf_coeff);
```

src/particles/elements/RFCavity.H
```diff
@@ -42,23 +89,49 @@ namespace impactx
...
+                                  sin_coef.begin(), sin_coef.end(),
+                                  m_sin_coef.begin());
+            amrex::Gpu::synchronize();
+
+            // low-level objects we can use on device
+            m_cos_data = m_cos_coef.data();
+            m_sin_data = m_sin_coef.data();
         }
 
+        AMREX_GPU_HOST
```
ax3l, line 132: This is a bit of a hack to avoid trying to call the destructor on GPU at the end of the kernel, with respect to the `m_cost_coeff`/`m_sin...` data members.

The reason is that we currently copy the whole element on the GPU in `beamoptic.H` - `PushSingleParticle`.

Nonetheless, this runs well on GPU (tested on Nvidia) and I'll clean this up later with a better design.

ax3l (approved):

src/particles/elements/RFCavity.H
```diff
@@ -107,22 +109,28 @@ namespace impactx
           : Thick(ds, nslice),
             m_escale(escale), m_freq(freq), m_phase(phase), m_mapsteps(mapsteps)
         {
-            m_cos_coef.resize(cos_coef.size());
-            m_sin_coef.resize(sin_coef.size());
+            m_cos_size = cos_coef.size();
+            m_sin_size = sin_coef.size();
+            if (m_cos_size != m_sin_size)
+                throw std::runtime_error("RFCavity: cos and sin coefficients must have same length!");
```
ax3l, line 115: Oh, I should better guard this to be host-only, too.

PR #309, opened 2023-01-31
Add soft solenoid

cemitch99:
Added a soft-edge solenoid.

## To Do

- [x] default coefficients
- [x] user-provided coefficients (as in RF cavity)
- [x] Python binding
- [x] Input parameter documentation
- [x] Test (app, py + readme in docs)
- [x] Has MAD-X syntax a soft-edge solenoid? We are not aware of one.


src/particles/elements/SoftSol.H
```diff
@@ -0,0 +1,402 @@
...
+            amrex::ParticleReal const s = refpart.s;
+            amrex::ParticleReal const sedge = refpart.sedge;
+
+            // initialize linear map (deviation) values
+            for (int i=1; i<7; i++) {
+               for (int j=1; j<7; j++) {
+                  if (i == j)
+                      refpart.map(i, j) = 1.0_prt;
+                  else
+                      refpart.map(i, j) = 0.0_prt;
```
ax3l: Compilers can sometimes optimize such constructs that use the ternary (conditional) operator better:
```suggestion
                  auto const default_value = (i == j) ? 1.0_prt : 0.0_prt;
                  refpart.map(i, j) = default_value;
```
https://en.cppreference.com/w/cpp/language/operator_other

src/particles/elements/SoftSol.H
```diff
@@ -0,0 +1,402 @@
...
+            integrators::symp2_integrate_split3(refpart,zin,zout,nsteps,*this);
+            amrex::ParticleReal const ptf = refpart.pt;
+
+            // print computed linear map:
+            //for(int i=1; i<7; ++i){
+            //   for(int j=1; j<7; ++j){
+            //      amrex::PrintToFile("SolMap.txt") << i << " " <<
+            //      j << " " << refpart.map(i,j) << "\n";
+            //   }
+            //}
```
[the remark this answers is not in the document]
ax3l: I am ok if you like to leave this in.

Notes:
- this only works on CPU (not GPU)
- maybe use `/* ... */` for the whole block instead of all the `//` :)
- ~~use `amrex::AllPrintToFile` - otherwise this is a race condition in MPI and OpenMP~~
  - update: ah no, `PrintToFile` makes only the IOProcessor print - that is fine.

src/particles/elements/SoftSol.H
```diff
@@ -0,0 +1,402 @@
...
+            constexpr amrex::ParticleReal zlen = 3.820395_prt;
+            constexpr amrex::ParticleReal zmid = zlen / 2.0_prt;
+            constexpr int ncoef = 2;
+
+            // specify Fourier coefficients:
+            //   Fourier coefficients for the on-axis longitudinal
+            //   magnetic field Bz of ..... (default solenoid model).
+            // May be superseded by user-provided input values (to do).
+            constexpr std::array<amrex::ParticleReal, ncoef> cos_coef = {
+                2.00000000000000000,
```
ax3l:
```suggestion
                2.0,
```

src/particles/elements/SoftSol.H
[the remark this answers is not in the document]
cemitch99: Change applied.

src/particles/elements/SoftSol.H
```diff
@@ -0,0 +1,439 @@
...
+
+namespace impactx
+{
+    struct Sol_field_data
+    {
+       //   Default:
+       //   Fourier coefficients for the on-axis longitudinal
+       //   magnetic field Bz of a soft-edge solenoid based on
+       //   the design used in the LCLS-II photoinjector.
+       //   Superseded by user-provided input values.
```
ax3l: Reminder for ourselves: Let's add a reference (here and in user-facing docs)?
Cite a proceeding paper?

cemitch99: A reference now appears here.

examples/solenoid_softedge/README.rst
```diff
@@ -0,0 +1,54 @@
...
+Soft-edge solenoid
+===================
+
+Proton beam propagating through a 6 m region containing a soft-edge
+solenoid.
+
+The solenoid model used is the default thin-shell model described in:
+P. Granum et al, "Efficient calculations of magnetic fields of solenoids for simulations,"
+NIMA 1034, 166706 (2022)
+https://doi.org/10.1016/j.nima.2022.166706
```
ax3l: Just some nice formatting of the link:
```suggestion
`DOI:10.1016/j.nima.2022.166706 <https://doi.org/10.1016/j.nima.2022.166706>`__
```

examples/solenoid_softedge/analysis_solenoid_softedge.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022-2023 ImpactX contributors
+# Authors: Axel Huebl, Chad Mitchell
```
ax3l: New file you authored: :)
```suggestion
# Authors: Chad Mitchell, Axel Huebl
```

examples/solenoid_softedge/run_solenoid_softedge.py
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022-2023 ImpactX contributors
+# Authors: Marco Garten, Axel Huebl, Chad Mitchell
```
ax3l: New file you authored: :)
```suggestion
# Authors: Chad Mitchell, Axel Huebl
```

src/python/elements.cpp
```diff
@@ -216,6 +216,21 @@ void init_elements(py::module& m)
...
+    py::class_<SoftSolenoid, elements::Thick>(me, "SoftSolenoid")
+        .def(py::init<
+                 amrex::ParticleReal const,
+                 amrex::ParticleReal const,
+                 amrex::Vector<amrex::ParticleReal>,
+                 amrex::Vector<amrex::ParticleReal>,
+                 int const,
+                 int const
+             >(),
+             py::arg("ds"), py::arg("bscale"), py::arg("cos_coefficients"), py::arg("sin_coefficients"),
```
ax3l: Better readable long line:
```suggestion
             py::arg("ds"), py::arg("bscale"),
             py::arg("cos_coefficients"), py::arg("sin_coefficients"),
```

examples/solenoid_softedge/analysis_solenoid_softedge.py
```diff
@@ -0,0 +1,115 @@
...
+print("")
+print("Final Beam:")
+sigx, sigy, sigt, emittance_x, emittance_y, emittance_t = get_moments(final)
+print(f"  sigx={sigx:e} sigy={sigy:e} sigt={sigt:e}")
+print(
+    f"  emittance_x={emittance_x:e} emittance_y={emittance_y:e} emittance_t={emittance_t:e}"
+)
+
+atol = 0.0  # ignored
+rtol = 2.0 * num_particles**-0.5  # from random sampling of a smooth distribution
```
ax3l, line 100: Either there are still bugs or we need to relax this more as seen on Windows: mostly for the emittance values.

src/particles/elements/SoftSol.H
```diff
@@ -0,0 +1,439 @@
...
+            amrex::ParticleReal const ptf = refpart.pt;
+
+            /* print computed linear map:
+               for(int i=1; i<7; ++i){
+                 for(int j=1; j<7; ++j){
+                    amrex::PrintToFile("SolMap.txt") << i << " " <<
+                    j << " " << refpart.map(i,j) << "\n";
+                 }
+               }
+            */
```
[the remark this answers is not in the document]
cemitch99: I think we decided (for the RFCavity) that this was ok to leave in for benchmarking purposes.

examples/solenoid_softedge/analysis_solenoid_softedge.py
cemitch99: We see that the Windows test (and only the Windows test) now fails with NaN for some values, so this is not a precision issue, but some OS/compiler-dependent bug.

cemitch99: The failure is limited to the Windows tests, and this appears to be a compiler-dependent bug (not a statistical fluctuation due to the initial random sampling).  The output on the two Windows test is very different (eg, a factor of ~2 in sig_y).  On both tests, the initial and final emittance values coincide _exactly._  This does not make sense, because the beam is rotated  by the solenoid, resulting in a coupling of the horizontal and vertical planes.  This suggests that there may be a problem with post-processing of output for the Windows tests.

ax3l: THank you for the analysis. I also see that the tests on Windows only fail for the inputs-file driven case, not for the Python-equivalent one.

This points to a hidden bug maybe in init, not post-processing.

ax3l: Found it: #320

ax3l: pushed a fix here as well

ax3l (approved):

PR #320, opened 2023-03-14
Fix: RFCavity Coefficients (Use-After-Free)

ax3l:
Fix the copy (and move) logic of `RFCavity` for the Fourier coefficients. The challenge is that the data container that holds the coefficients on the device can only be host-copied, but not host-device (, device-host or device-device) copied.

Thus, we need to split our data members to keep the element itself copyable without letting the pointers to the device pointers go stale.

Ref.: https://en.cppreference.com/w/cpp/language/rule_of_three

### DIFF
--- a/src/particles/elements/RFCavity.H
+++ b/src/particles/elements/RFCavity.H
@@ -105,7 +105,7 @@ namespace data
         RFCavity_device_copyable(RFCavity_device_copyable &&) = default;
         RFCavity_device_copyable& operator=(RFCavity_device_copyable &&) = default;
     };
-} // namespace details
+} // namespace data
 
     struct RFCavity
     : public elements::BeamOptic<RFCavity>,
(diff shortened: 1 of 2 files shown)


src/particles/elements/RFCavity.H
```diff
@@ -76,9 +77,40 @@ namespace impactx
...
+            int mapsteps = 1
+        ) : m_escale(escale), m_freq(freq), m_phase(phase), m_mapsteps(mapsteps)
+        {}
+
+        RFCavity_device_copyable(RFCavity_device_copyable const &) = default;
+        RFCavity_device_copyable& operator=(RFCavity_device_copyable const &) = default;
+        RFCavity_device_copyable(RFCavity_device_copyable &&) = default;
+        RFCavity_device_copyable& operator=(RFCavity_device_copyable &&) = default;
+    };
+} // namespace details
```
ax3l:
```suggestion
} // namespace data
```

PR #323, opened 2023-03-23
Element Push: Restructure

ax3l:
## Element Push: Restructure

Add an overarching particle container push operator to each element. This one by default pushes the reference particle and then the particle beam, as before.

## Purpose

This is a preparation for diagnostics in #299 and ML elements, which will operate on the whole particle container. It's also a preparation to simplify the bug fixes introduced for elements with dynamically allocated state from #320.

### DIFF
--- a/src/particles/elements/Programmable.H
+++ b/src/particles/elements/Programmable.H
@@ -26,9 +26,21 @@ namespace impactx
         static constexpr auto name = "Programmable";
         using PType = ImpactXParticleContainer::ParticleType;
 
-        /** This element writes the particle beam out to openPMD data.
+        /** This element can be programmed
          */
-        Programmable () {}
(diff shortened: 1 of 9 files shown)


src/particles/elements/Programmable.H
```diff
@@ -26,9 +26,21 @@ namespace impactx
         static constexpr auto name = "Programmable";
         using PType = ImpactXParticleContainer::ParticleType;
 
-        /** This element writes the particle beam out to openPMD data.
+        /** This element can be programmed
          */
-        Programmable () {}
+        Programmable (amrex::ParticleReal ds=0.0, int nslice=1)
+            : m_ds(ds), m_nslice(nslice)
+        {}
```
ax3l, line 33: These constructor arguments were forgotten to be exposed earlier.
The default arguments default to the prior signature.

PR #325, opened 2023-03-27
Programmable Element: Thread Safety

ax3l:
Expose an option to opt-in into thread safety for the programmable element. This controls if the particle iterator loop is performed in serial, even when the OpenMP backend is used.

Thread-safety is easy to achive for regular beamline elements that are implemented via maps or intergrators.

Thread-safety is not easily guaranteed for interaction with external libraries, e.g., I/O or ML packages.

### DIFF
--- a/examples/fodo/run_fodo_programmable.py
+++ b/examples/fodo/run_fodo_programmable.py
@@ -139,6 +139,7 @@ pge1.nslice = ns
 pge1.beam_particles = lambda pti, refpart: my_drift(pge1, pti, refpart)
 pge1.ref_particle = lambda refpart: my_ref_drift(pge1, refpart)
 pge1.ds = 0.25
+pge1.threadsafe = True  # allow OpenMP threading for speed
 
 # attention: assignment is a reference for pge2 = pge1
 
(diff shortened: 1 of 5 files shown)


src/particles/elements/Programmable.H
```diff
@@ -79,6 +79,7 @@ namespace impactx
     public:
         amrex::ParticleReal m_ds = 0.0; //! segment length in m
         int m_nslice = 1; //! number of slices used for the application of space charge
+        bool m_omp_parallel = false; //! allow threading via OpenMP for the particle iterator loop (note: if OMP backend is active)
```
ax3l, line 82: The default is off to do the safe thing by default.
(Correctness >> speed.)

ax3l:
@RTSandberg any thoughts? :)

examples/fodo/run_fodo_programmable.py
```diff
@@ -147,6 +148,7 @@ def my_ref_drift(pge, refpart):
 pge2.beam_particles = lambda pti, refpart: my_drift(pge2, pti, refpart)
 pge2.ref_particle = lambda refpart: my_ref_drift(pge2, refpart)
 pge2.ds = 0.5
+pge2.threadsafe = True  # allow OpenMP threading for speed
```
ax3l, line 151: Comment by @RemiLehe:
We probably need to release the GIL here.

src/particles/elements/Programmable.cpp
```diff
@@ -21,7 +21,7 @@ namespace impactx
     ) const
     {
         if (m_push == nullptr) {
-            push_all(pc, *this, step);
+            push_all(pc, *this, step, m_threadsafe);
```
ax3l, line 24: Comment by @RemiLehe:
We probably need to release the GIL here.

PR #326, opened 2023-03-27
Fix: Push Reference Particle 1x in Step

ax3l:
I just realized that we push the reference particle multiple times per step if we have multiple boxes per MPI rank. This fixes it.


cemitch99 (approved):
Great catch, thanks.

PR #330, opened 2023-03-30
Beam Monitor: Output Name

ax3l:
Allow to change the output name from the default `<element_name>` in inputs files, similar to the option exposed in Python.

Follow-up to #299


cemitch99 (approved):
Looks good to me.

Issue #269, opened 2022-10-03
In-Situ Beam Parameter Calculation for Reduced Diagnostics


n01r:
Adding in-situ calculated beam parameters to ImpactX would reduce I/O and make typically used quantities like beta-function, emittance and distribution momenta readily available without additional post-processing.

- [x] Implement beam parameter calculation
  - [x] sigma_x,y,z
  - [x] <x/y/z> 
  - [x] <p_x/y/z>
  - [x] min/max 
  - [x] emittance_x/y/z
  - [x] alpha: x'-... (only for fully matched beam equal to twiss paramters)
- [x] add ReducedDiagnostics for them
- [x] user-facing docs
- [x] implement/update tests

n01r:
Oh, this was completed a while ago. Closing it now.

PR #336, opened 2023-04-08
Add reduced diagnostics for beam

n01r:
Calculate and print mean values and standard deviations for positions and momenta. 
Derive RMS emittances from those and also print them.

So far, this only prints out the calculated values into the standard output.
However, the values are consistent with the analysis from the test of the FODO cell.

This is the first part for closing #269.

**Edit 2023-04-18: To Do**
- [x] Add Python bindings
- [x] Add documentation for output file


src/particles/diagnostics/BeamRelevant.cpp
```diff
@@ -0,0 +1,225 @@
...
+                        // prepare mean position values
+                        const ParticleReal p_x_mean = p_pos0*p_w;
+                        const ParticleReal p_y_mean = p_pos1*p_w;
+                        const ParticleReal p_t_mean = p_pos2*p_w;
+
+                        const ParticleReal p_ux_mean = p_ux*p_w;
+                        const ParticleReal p_uy_mean = p_uy*p_w;
+                        const ParticleReal p_ut_mean = p_ut*p_w;
+                        const ParticleReal p_gm_mean = std::sqrt(1.0_rt+p_us*inv_c2)*p_w;
+
```
n01r: alpha and beta (CS parameters) are still missing

src/particles/diagnostics/BeamRelevant.cpp
```diff
@@ -0,0 +1,225 @@
...
+                        const ParticleReal p_us = p_ux*p_ux + p_uy*p_uy + p_ut*p_ut;
+                        // prepare mean position values
+                        const ParticleReal p_x_mean = p_pos0*p_w;
+                        const ParticleReal p_y_mean = p_pos1*p_w;
+                        const ParticleReal p_t_mean = p_pos2*p_w;
+
+                        const ParticleReal p_ux_mean = p_ux*p_w;
+                        const ParticleReal p_uy_mean = p_uy*p_w;
+                        const ParticleReal p_ut_mean = p_ut*p_w;
+                        const ParticleReal p_gm_mean = std::sqrt(1.0_rt+p_us*inv_c2)*p_w;
```
cemitch99: I'm not sure about this line.  What is p_gm intended to represent?  Internally, the momenta should be dimensionless, so I don't think 1 + p_us*inv_c2 can work for dimensional reasons.

src/particles/diagnostics/BeamRelevant.cpp
```diff
@@ -0,0 +1,225 @@
...
+                    pc,
+                    [=] AMREX_GPU_DEVICE(const PTDType& ptd, const int i) noexcept -> amrex::GpuTuple
+                            <ParticleReal,ParticleReal,ParticleReal,ParticleReal,ParticleReal,ParticleReal,
+                                    ParticleReal,ParticleReal,ParticleReal,ParticleReal,ParticleReal>
+                    {
+                        const ParticleReal p_ux = ptd.m_rdata[RealSoA::ux][i];
+                        const ParticleReal p_uy = ptd.m_rdata[RealSoA::uy][i];
+                        const ParticleReal p_ut = ptd.m_rdata[RealSoA::pt][i];
+                        const ParticleReal p_us = p_ux*p_ux + p_uy*p_uy + p_ut*p_ut;
+                        const ParticleReal p_gm = std::sqrt(1.0_rt+p_us*inv_c2);
```
cemitch99: The same comment applies here regarding p_gm.

src/particles/diagnostics/BeamRelevant.cpp
```diff
@@ -0,0 +1,225 @@
...
+                        const ParticleReal p_y = p_pos1;
+                        const ParticleReal p_t = p_pos2;
+                        // prepare mean square for positions
+                        const ParticleReal p_x_ms = (p_x-x_mean)*(p_x-x_mean)*p_w;
+                        const ParticleReal p_y_ms = (p_y-y_mean)*(p_y-y_mean)*p_w;
+                        const ParticleReal p_z_ms = (p_t-t_mean)*(p_t-t_mean)*p_w;
+                        // prepare mean square for momenta
+                        const ParticleReal p_ux_ms = (p_ux-ux_mean)*(p_ux-ux_mean)*p_w;
+                        const ParticleReal p_uy_ms = (p_uy-uy_mean)*(p_uy-uy_mean)*p_w;
+                        const ParticleReal p_uz_ms = (p_ut-ut_mean)*(p_ut-ut_mean)*p_w;
```
cemitch99: Is there any advantage to switching notation from ut to uz here?  It's fine, but it may be confusing to those reading the code.

src/particles/diagnostics/BeamRelevant.cpp
```diff
@@ -0,0 +1,225 @@
...
+            ParticleReal sig_t = std::sqrt(t_ms);
+            // standard deviations of momenta
+            ParticleReal sig_px = std::sqrt(ux_ms);
+            ParticleReal sig_py = std::sqrt(uy_ms);
+            ParticleReal sig_pt = std::sqrt(ut_ms);
+            // RMS emittances
+            ParticleReal emittance_x = std::sqrt(x_ms*ux_ms-xux*xux);
+            ParticleReal emittance_y = std::sqrt(y_ms*uy_ms-yuy*yuy);
+            ParticleReal emittance_z = std::sqrt(t_ms*ut_ms-tut*tut);
+
```
cemitch99: More useful to users will be the normalized emittance, so at some stage we should multiply emittance_x,y,z by the reference value of beta*gamma.  The emittance values will then differ from those currently used in the benchmark validation scripts, so we should keep this in mind.

src/particles/diagnostics/BeamRelevant.cpp
```diff
@@ -0,0 +1,225 @@
...
+            amrex::AllPrint() << step << " "
+                              << x_mean << " " << y_mean << " " << t_mean << " "
+                              << sig_x << " " << sig_y << " " << sig_t << " "
+                              << ux_mean << " " << uy_mean << " " << ut_mean << " "
+                              << sig_px << " " << sig_py << " " << sig_pt << " "
+                              << emittance_x << " " << emittance_y << " " << emittance_z << " "
+                               << charge << " "
+                               << "\n";
+
+        }
```
cemitch99: This output looks great.  In IMPACT-Z, we output these quantities, except that the reference value of "s" replaces "step" in the first column, since users are interested in the beam moments as a function of lattice location.

cemitch99 (commented):
The math looks good to me.  I added a few minor comments.

src/particles/diagnostics/BeamRelevant.cpp
n01r: Thanks for the suggestion! I just added "s" to the output and kept the steps, just so we have both available for testing or debugging. :)

src/particles/diagnostics/BeamRelevant.cpp
n01r: Thanks, @cemitch99! I am printing out beta*gamma now as well so that we can just compute this offline, if that's okay. :)
This pull request is going to be an intermediate step towards the output in openPMD but if anyone is going to use it during this stage, we have it available now as well. 👍

src/particles/diagnostics/BeamRelevant.cpp
n01r: Thanks! I removed those :)

src/particles/diagnostics/BeamRelevant.cpp
n01r: Fixed now!

src/particles/diagnostics/ReducedBeamCharacteristics.H
```diff
@@ -0,0 +1,61 @@
...
+ * This file is part of ImpactX.
+ *
+ * Authors: Marco Garten, Chad Mitchell, Axel Huebl
+ * License: BSD-3-Clause-LBNL
+ */
+#ifndef IMPACTX_REDUCED_BEAM_CHARACTERISTICS_H
+#define IMPACTX_REDUCED_BEAM_CHARACTERISTICS_H
+
+#include "particles/ImpactXParticleContainer.H"
+#include "ImpactX.H"
```
ax3l: I think this is an unused import.
```suggestion
```

src/particles/diagnostics/ReducedBeamCharacteristics.H
```diff
@@ -0,0 +1,61 @@
...
+
+#include <cmath>
+
+using namespace amrex;
+
+namespace impactx::diagnostics
+{
+    /** Struct containing reduced beam characteristics data
+     */
+     struct ReducedBeamCharacteristics{
```
ax3l: One could also use a `std::unordered_map<std::string, amrex::ParticleReal>` instead. That way, the Python bindings are a `dict` and one does not need to bind another type.

src/particles/diagnostics/ReducedBeamCharacteristics.H
```diff
@@ -0,0 +1,61 @@
...
+#include "particles/ImpactXParticleContainer.H"
+#include "ImpactX.H"
+
+#include <AMReX_Extension.H>
+#include <AMReX_REAL.H>
+#include <AMReX_GpuComplex.H>
+
+#include <cmath>
+
+using namespace amrex;
```
ax3l: Do not use namespaces in `.H` files - this will "pollute" the include space of every downstream file that includes it.
```suggestion
```

Please only use this in `.cpp` files and even better, only in `{ ... }` sections of functions when needed at all.

src/particles/diagnostics/ReducedBeamCharacteristics.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include "particles/diagnostics/ReducedBeamCharacteristics.H"
+
+#include <AMReX_ParallelDescriptor.H>   // for ParallelDescriptor
+#include <AMReX_REAL.H>                 // for ParticleReal
+#include <AMReX_Reduce.H>               // for ReduceOps
+#include <AMReX_ParticleReduce.H>       // for ParticleReduce
+#include <AMReX_ParallelDescriptor.H>   // for ParallelDescriptor
+#include <AMReX_BLProfiler.H>           // for TinyProfiler
+
+using namespace amrex;
```
ax3l: We don't need this even in C++ files, `amrex::` is short.
```suggestion
```

src/particles/diagnostics/ReducedBeamCharacteristics.H
ax3l: The only thing we really use in `{ ... }` is `using namespace amrex::literals; // for _rt and _prt`. The rest, please always write `amrex::` for clarity.

n01r:
The new commit adds the Python bindings, and I tested already that they work.
Documentation is still missing.

tests/python/test_impactx.py
```diff
@@ -26,6 +26,11 @@ def test_impactx_fodo_file():
     sim.init_lattice_elements_from_inputs()
 
     sim.evolve()
+    # test calculation of reduced beam characteristics
+    reduced_beam_data = sim.reduced_beam_characteristics()
```
n01r: @ax3l, do we want a separate test or is it fine to include this here?

ax3l: Here is good, please add a validation for the final beam here as done in these lines:
https://github.com/ECP-WarpX/impactx/blob/c29f1ec3f6261963c45b1063afac8d496ebd9e7f/examples/fodo/analysis_fodo.py#L88-L104

src/python/ImpactX.cpp
```diff
@@ -6,6 +6,7 @@
 #include "pyImpactX.H"
 
 #include <ImpactX.H>
+#include "particles/diagnostics/ReducedBeamCharacteristics.H"
```
ax3l: See code around your patch :)
```suggestion
#include <particles/diagnostics/ReducedBeamCharacteristics.H>
```

src/particles/diagnostics/ReducedBeamCharacteristics.H
```diff
@@ -0,0 +1,30 @@
...
+#ifndef IMPACTX_REDUCED_BEAM_CHARACTERISTICS_H
+#define IMPACTX_REDUCED_BEAM_CHARACTERISTICS_H
+
+#include "particles/ImpactXParticleContainer.H"
+
+#include <AMReX_Extension.H>
+#include <AMReX_REAL.H>
+#include <AMReX_GpuComplex.H>
+
+#include <cmath>
```
ax3l:
```suggestion
#include <cmath>
#include <string>
#include <unordered_map>

```

tests/python/test_impactx.py
n01r: Thanks! Final beam validation was added

src/particles/diagnostics/ReducedBeamCharacteristics.cpp
```diff
@@ -0,0 +1,224 @@
...
+            amrex::ParticleReal,amrex::ParticleReal,amrex::ParticleReal,amrex::ParticleReal,amrex::ParticleReal>>(
+                    pc,
+                    [=] AMREX_GPU_DEVICE(const PTDType& ptd, const int i) noexcept -> amrex::GpuTuple
+                            <amrex::ParticleReal,amrex::ParticleReal,amrex::ParticleReal,amrex::ParticleReal,amrex::ParticleReal,
+                            amrex::ParticleReal,amrex::ParticleReal>
+                    {
+                        // access AoS particle position data
+                        const amrex::ParticleReal p_pos0 = ptd.m_aos[i].pos(0);
+                        const amrex::ParticleReal p_pos1 = ptd.m_aos[i].pos(1);
+                        const amrex::ParticleReal p_pos2 = ptd.m_aos[i].pos(2);
```
ax3l: Unless I am mistaken, you cannot access `ptd.m_aos` as a whole here.
Use the pattern we use in WarpX for now please.
https://github.com/ECP-WarpX/WarpX/blob/872bae2026a6767f2023f8ba600c87d247eed38e/Source/Diagnostics/ReducedDiags/BeamRelevant.cpp#L198-L209

n01r: > Unless I am mistaken, you cannot access ptd.m_aos as a whole here.   

Sorry, could you explain? 🤔  It works on MacOS and on Perlmutter with all tests that I did. 
I was trying the pattern you mentioned before but the one I am using now is the one that @atmyers suggested.

ax3l: As discussed on Slack: could be a compiler bug for the NVCC version in CI. Please try again w/ the syntax in the link above.

docs/source/dataanalysis/dataanalysis.rst
```diff
@@ -3,8 +3,38 @@
...
+* ``alpha_x``, ``alpha_y``, ``alpha_t``
+    Twiss alpha
+* ``beta_x``, ``beta_y``, ``beta_t``
+    Twiss beta
+* ``charge``
+    Cumulated beam charge
+
 .. note::
 
    TODO :-)
```
ax3l:
```suggestion
   openPMD beam monitor output: TODO :-)
```

docs/source/usage/python.rst
```diff
@@ -113,6 +113,11 @@ General
 
       Run the main simulation loop for a number of steps.
 
+   .. py:method:: reduced_beam_characteristics()
+
+      Compute reduced beam characteristics like the position and momentum moments of the particle distribution, as well as emittance and Twiss parameters.
```
ax3l, line 118:
```suggestion
      Compute reduced beam characteristics like the position and momentum moments of the particle distribution, as well as emittance and Twiss parameters.

      :return: beam properties with string keywords
      :rtype: dict
```

src/particles/diagnostics/DiagnosticOutput.cpp
```diff
@@ -42,9 +43,35 @@ namespace impactx::diagnostics
...
+                             << "charge" << " "
+                             << "\n";
             }
         }
 
+        if (otype == OutputType::PrintReducedBeamCharacteristics) {
+            std::unordered_map<std::string, amrex::ParticleReal> const rbc = diagnostics::reduced_beam_characteristics(
+                    pc);
+
+            file_handler << step << " " << rbc.at("s") << " " << rbc.at("beta_gamma") << " "
```
ax3l:
```suggestion
            file_handler << step << " " << rbc.at("s") << " " << rbc.at("ref_beta_gamma") << " "
```

src/particles/diagnostics/DiagnosticOutput.cpp
```diff
@@ -42,9 +43,35 @@ namespace impactx::diagnostics
                 file_handler << "id H I\n";
             } else if (otype == OutputType::PrintRefParticle) {
                 file_handler << "step s x y z t px py pz pt\n";
+            } else if (otype == OutputType::PrintReducedBeamCharacteristics) {
+                file_handler << "step" << " " << "s_ref_part" << " " << "beta_gamma_ref_part" << " "
```
ax3l:
```suggestion
                file_handler << "step" << " " << "s" << " " << "ref_beta_gamma" << " "
```

docs/source/dataanalysis/dataanalysis.rst
```diff
@@ -3,8 +3,38 @@
...
+
+ImpactX calculates reduced beam characteristics like averaged positions, momenta, beam emittances and Courant-Snyder (Twiss) parameters during runtime.
+These quantities are calculated before, after, and during each step of the simulation.
+If ``diag.slice_step_diagnostics`` is enabled, they will also be calculated during each slice of each beamline element.
+
+The code writes out the values in an ASCII file prefixed ``reduced_beam_characteristics`` containing the follow columns:
+
+* ``step``
+    Iteration within the simulation
+* ``s_ref_part``, ``beta_gamma_ref_part``
```
ax3l:
```suggestion
* ``s``, ``ref_beta_gamma``
```

ax3l:
Looking at the error message, this is CUDA 11.0 but looks similar to a bug in CUDA 11.4-11.8
https://github.com/pybind/pybind11/issues/4193#issuecomment-1270491939
```
... <char,     ::std::char_traits<char> , ::std::allocator<char> > , float> > >  (*)(const  ::impactx::ImpactXParticleContainer &),(& :: impactx::diagnostics::reduced_beam_characteristics[abi:cxx11]), 1> >,const  ::amrex::ConstParticleTileData< ::amrex::Particle<(int)0, (int)0> , (int)5, (int)0> >>> 
```

Update: ah no, actually the casts here are ok-ish, I think.
Maybe a problem with `& :: impactx::diagnostics::reduced_beam_characteristics[abi:cxx11]` or something else from the code transformation in that CUDA version. Testing if bumping to 11.3+ via #361 helps.

src/particles/diagnostics/ReducedBeamCharacteristics.cpp
```diff
@@ -0,0 +1,251 @@
...
+ *
+ * This file is part of ImpactX.
+ *
+ * Authors: Marco Garten, Chad Mitchell, Yinjian Zhao, Axel Huebl
+ * License: BSD-3-Clause-LBNL
+ */
+
+#include "ImpactX.H"
+#include "particles/ImpactXParticleContainer.H"
+#include "particles/diagnostics/ReducedBeamCharacteristics.H"
```
ax3l: Own import always first + `ImpactX.H` not used:
```suggestion
#include "ReducedBeamCharacteristics.H"

#include "particles/ImpactXParticleContainer.H"
#include "particles/ReferenceParticle.H"
```

ax3l (approved):
Thank you, this is great! :sparkles:

PR #339, opened 2023-04-11
amrex.the_arena_is_managed=0

ax3l:
Disable managed memory by default for performance and non-bugness.

Please disable explicitly if needed for some of the current scripting tasks, e.g., in the current Python bindings.

Tests:
- [x] all ctest w/o the pytests of AMReX tests (to fix) and `FODO.programmable.py.run` (no AoS cupy support) ran (CUDA 11.7) and passed as before

Follow-up:
- [x] `test_charge_deposition`: explicit memory copy #340

### DIFF
--- a/tests/python/test_charge_deposition.py
+++ b/tests/python/test_charge_deposition.py
@@ -63,7 +63,7 @@ def test_charge_deposition(save_png=True):
         bx = mfi.validbox()
         rbx = amrex.RealBox(bx, dr, gm.ProbLo())
 
-        arr = rho.array(mfi)
+        arr = rho.array(mfi)  # TODO: explicit device-to-host memory copy
         arr_np = np.array(arr, copy=False)  # indices: comp, z, y, x
 
         # shift box to zero-based local mfi index space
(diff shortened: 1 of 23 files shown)


tests/python/test_charge_deposition.py
```diff
@@ -63,7 +63,7 @@ def test_charge_deposition(save_png=True):
         bx = mfi.validbox()
         rbx = amrex.RealBox(bx, dr, gm.ProbLo())
 
-        arr = rho.array(mfi)
+        arr = rho.array(mfi)  # TODO: explicit device-to-host memory copy
```
ax3l, line 66: This is easy to do with cupy, but here I would potentially rely purely on AMReX for now if we can copy the whole MultiFab before the loop (or each box at this location).

ax3l: @WeiqunZhang commented:
```C++
MultiFab hostmf(devicemf.boxArray(), devicemf.DistributionMap(), devicemf.nComp(), devicemf.nGrowVect(), MFInfo().SetArena(The_Pinned_Arena()));
amrex::dtoh_memcpy(hostmf, devicemf);
```

I will expose the free standing copy function in pyAMReX.

ax3l: pyAMReX: https://github.com/AMReX-Codes/pyamrex/pull/123
Follow-up: #340

PR #340, opened 2023-04-13
Python Tests: w/o Managed Memory

ax3l:
Follow-up to #339

- [x] needs https://github.com/AMReX-Codes/pyamrex/pull/123

### DIFF
--- a/tests/python/test_charge_deposition.py
+++ b/tests/python/test_charge_deposition.py
@@ -56,14 +56,25 @@ def test_charge_deposition(save_png=True):
 
     half_z = sim.n_cell[2] // 2  # order: x,y,z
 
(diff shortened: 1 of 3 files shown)


tests/python/test_charge_deposition.py
```diff
@@ -56,14 +56,25 @@ def test_charge_deposition(save_png=True):
 
     half_z = sim.n_cell[2] // 2  # order: x,y,z
 
+    # copy data from device to host
+    rho_host = amrex.MultiFab(
```
ax3l: We can add an if here for GPU, if CPU only, we can use an alias

ax3l: done

PR #346, opened 2023-04-18
Doc: Theory Ref. Trajectory

ax3l:
Publishing some of the great details Chad shared with us on Slack :pray:

### DIFF
--- a/docs/source/theory/intro.rst
+++ b/docs/source/theory/intro.rst
@@ -9,8 +9,11 @@ Concepts
 Reference Trajectory
 """"""""""""""""""""
 
-ImpactX is an *s*-based beam dynamics code, evolving particles *relative* to a reference trajectory.
+ImpactX is an *s*-based beam dynamics code, tracking particles using the longitudinal accelerator lattice position :math:`s` as the independent dynamical variable.
+Particle phase space coordinates are specified relative to a nominal reference trajectory.
 
+The reference trajectory plays an important role in ImpactX.
+In addition to specifying the nominal machine orbit, it specifies the relationship between the local beam coordinate system and the global lab coordinate system.
 A reference trajectory is favorable instead of, e.g., differences to the beam centroid, because:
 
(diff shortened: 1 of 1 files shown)


docs/source/theory/intro.rst
```diff
@@ -3,6 +3,27 @@
...
 ============
 
+Concepts
+--------
+
+Reference Trajectory
+""""""""""""""""""""
+
+ImpactX is an *s*-based beam dynamics code, evolving particles *relative* to a reference trajectory.
+
```
cemitch99, line 15: Instead of "evolving particles..." what about "tracking particles using the longitudinal lattice position _s_ as the independent dynamical variable.  Particle phase space coordinates are specified relative to a nominal reference trajectory.

docs/source/theory/intro.rst
```diff
@@ -3,6 +3,27 @@
...
+A reference trajectory is favorable instead of, e.g., differences to the beam centroid, because:
```
cemitch99, line 18: Before this line, I would add:  

"The reference trajectory plays an important role in ImpactX.  In addition to specifying the nominal machine orbit, it specifies the relationship between the local beam coordinate system and the global lab coordinate system.

cemitch99 (commented):
Thanks!  I made a few suggestions.

docs/source/theory/intro.rst
ax3l: Thanks! I thought about tracking but wanted to avoid the impression that we do not model space charge (as in: tracking code). Let me see how to rephrase it.

cemitch99: How about starting with, "ImpactX is an s-based beam dynamics code with space charge."  The rest can be left as above?

cemitch99: "Particles are tracked using..."

ax3l: Thanks! Applied and I added one more concept section about the space charge kick.

cemitch99 (approved):

RTSandberg (approved):
Thanks for putting this in, it looks good to me

PR #349, opened 2023-04-24
Fix openPMD Meta-Data

ax3l:
## Tests

Check created files with `openPMD-validator` to conform with the `openPMD-standard`.

## Fixes

Add the `positionOffset` record of openPMD, which can be used for coarse particle positions or general offsets of a particle species.
This refers now to the offset added by the reference particle.

Follow-up to #299

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -55,6 +55,7 @@ jobs:
         python3 -m pip install -U pip setuptools wheel pytest
         python3 -m pip install -r requirements.txt
         python3 -m pip install -r examples/requirements.txt
+        python3 -m pip install -U openPMD-validator
     - name: Build
       run: |
         $env:HDF5_DIR = "C:/Program Files/HDF_Group/HDF5/1.12.2/cmake/"
@@ -137,6 +138,7 @@ jobs:
         python3 -m pip install -U pip setuptools wheel pytest
         python3 -m pip install -r requirements.txt
         python3 -m pip install -r examples/requirements.txt
+        python3 -m pip install -U openPMD-validator
     - name: Build
       shell: cmd
       run: |
@@ -175,3 +177,6 @@ jobs:
 #        if errorlevel 1 exit 1
 #        cmake --build build --target pip_install
 #        if errorlevel 1 exit 1
+
+    - name: validate created openPMD files
+      run: find build -name *.h5 | xargs -n1 -I{} openPMD_check_h5 -i {}
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -82,6 +82,8 @@ jobs:
         if(!$?) { Exit $LASTEXITCODE }
         cmake --build build --config RelWithDebInfo --target pip_install
         if(!$?) { Exit $LASTEXITCODE }
+    - name: validate created openPMD files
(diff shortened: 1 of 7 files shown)


src/particles/elements/diagnostics/openPMD.cpp
```diff
@@ -253,6 +254,17 @@ namespace detail
...
         }
+
+        // openPMD coarse position
+        {
+            beam["positionOffset"]["x"].resetDataset(d_fl);
+            beam["positionOffset"]["x"].makeConstant(ref_part.x);
+            beam["positionOffset"]["y"].resetDataset(d_fl);
+            beam["positionOffset"]["y"].makeConstant(ref_part.y);
+            beam["positionOffset"]["ct"].resetDataset(d_fl);
+            beam["positionOffset"]["ct"].makeConstant(ref_part.t);
```
ax3l, line 265: @cemitch99 here for `positionOffset` and in `position`, do we want to output `s`+`delta s` instead of `ct` + `c*dt`?

cemitch99: Not sure I follow what the intended output is here.  I understand that by "ct" you mean "ref_part.t", but what is meant by "dt"?  Does this represent one or more particle values of the "t" coordinate?

In general, I would say that "s" is qualitatively different from the other variables like ct.  It is an independent variable, it is associated only with the reference particle, and the other (beam particle) values are given as functions of "s".

ax3l: Thanks, ah yes I was confused for a second... So per particle, splitting the longitudinal position (for a fixed s) into `ref_part.t` (should we call this `ref_part.ct`?) and `position.ct` makes sense...

.github/workflows/windows.yml
```diff
@@ -175,3 +177,6 @@ jobs:
 #        if errorlevel 1 exit 1
 #        cmake --build build --target pip_install
 #        if errorlevel 1 exit 1
+
+    - name: validate created openPMD files
+      run: find build -name *.h5 | xargs -n1 -I{} openPMD_check_h5 -i {}
```
ax3l:
```suggestion
    - name: validate created openPMD files
      run: Get-ChildItem -Recurse -Filter *.h5 | %{openPMD_check_h5 -i $_}
```

src/particles/elements/diagnostics/openPMD.cpp
cemitch99: Yes, that's true.

cemitch99 (approved):
It looks ok to me, thanks.

.github/workflows/windows.yml
```diff
@@ -175,3 +177,6 @@ jobs:
 #        if errorlevel 1 exit 1
 #        cmake --build build --target pip_install
 #        if errorlevel 1 exit 1
+
+    - name: validate created openPMD files
+      run: Get-ChildItem -Recurse -Filter *.h5 | %{openPMD_check_h5 -i $_}
```
ax3l:
```suggestion
      run: Get-ChildItem -Path build -Recurse -Filter *.h5 | %{openPMD_check_h5 -i $_}
```

src/particles/elements/diagnostics/openPMD.cpp
ax3l: Ok, will do in #350

PR #350, opened 2023-04-25
Reference Particle: t -> ct

ax3l:
Rename for consistency with the same attribute on particles and to reflect its dimensionality of a length.


cemitch99 (commented):
This must have been tedious, but I'm not sure it is necessary.  This notation is equivalent to specifying the units for "t", which should be done anyway in the documentation.  (Likewise, the units of px, etc. are not apparent without visiting the documentation.)

ax3l:
Both is fine, I used the `rename` feature of my IDE for development, which is relatively fast.

I would still make particle `ct` and reference particle `t` consistent, both are normalized with `c` but named differently.

ax3l:
Will do the other way in #351

PR #351, opened 2023-04-25
Particle Enums: Use Consistently and Rename

ax3l:
- rename: `m_qm` to `qm` because it is charge over mass, not its inverse
- rename AoS to use s-based description by default: `z`->`t`
- Rename `AddNParticles` parameter, since we add particles for fixed `s`
- Use enums consistently whe we access `.pos` arguments, instead of `0-2`

PR #353, opened 2023-05-03
Particle Enums: Rename Momentum

ax3l:
Rename momentum to start with `p` for all components. Fix a comment for `t`.

Follow-up to #351


cemitch99 (approved):
Looks great to me.

PR #357, opened 2023-05-09
All.H: Newlines & Order

ax3l:
Format the distribution and element lists to be:
- `None` first
- rest alphabetical
- one per line

This should simplify merges and reduce merge conflicts.


cemitch99 (approved):
Looks good.

PR #358, opened 2023-05-11
Docs: Pip Requirements w/ upgrade

ax3l:
Make sure the requirements documented in pip files are upgraded if found but old.


cemitch99 (approved):
Looks good to me.

PR #359, opened 2023-05-11
Python: Fix Unnecessary Lambda

ax3l:
From CodeQL:
> This 'lambda' is just a simple wrapper around a callable object. Use that object directly.
>
> A lambda that calls a function without modifying any of its parameters is unnecessary. Python functions are first class objects and can be passed around in the same way as the resulting lambda.


n01r (approved):
Thanks :) looks good

PR #361, opened 2023-05-17
CUDA: Drop <11.3

ax3l:
Drop support for older CUDA releases. The CUDA Toolkit 11.0 added C++17 support but had a few severe issues with it, which are better in 11.3+. These releases are now widely available, so we can drop older releases.

Related to #336


ax3l:
This fixes the CUDA issue in #336

PR #362, opened 2023-05-18
Add positron LPA channel example

cemitch99:
Adds the input file for an example illustrating acceleration and focusing of a positron beam for a possible LPA-based linear electron-positron collider.  This is currently a draft only, illustrating a single lattice period.

- [x] Add input file for a single period
- [x] Add Python equivalent
- [x] Add and benchmark output of reduced diagnostics (moments, emittances)
- [x] Modify input file to support repetition to 250 periods (2.5 TeV)
- [x] use #365 
- [x] Update documentation


examples/positron_channel/README.rst
```diff
@@ -0,0 +1,74 @@
+.. _examples-fodo:
+
+FODO Cell
+=========
```
ax3l: since this is an anchor, we need to give it a unique name.
Adjusted title to avoid confusion.
```suggestion
.. _examples-fodo-positron:

FODO Cell w/ Positrons
======================
```

examples/positron_channel/README.rst
```diff
@@ -0,0 +1,74 @@
...
+Stable FODO cell with a zero-current phase advance of 67.8 degrees.
+
+The matched Twiss parameters at entry are:
+
+* :math:`\beta_\mathrm{x} = 2.82161941` m
+* :math:`\alpha_\mathrm{x} = -1.59050035`
+* :math:`\beta_\mathrm{y} = 2.82161941` m
+* :math:`\alpha_\mathrm{y} = 1.59050035`
+
+We use a 2 GeV electron beam with initial unnormalized rms emittance of 2 nm.
```
ax3l: Is the emittance correct?
```suggestion
We use a 10 GeV positron beam with initial unnormalized rms emittance of 2 nm.
```

examples/positron_channel/run_positron.py
```diff
@@ -0,0 +1,76 @@
...
+sim.particle_shape = 2  # B-spline order
+sim.space_charge = False
+# sim.diagnostics = False  # benchmarking
+sim.slice_step_diagnostics = True
+
+# domain decomposition & space charge mesh
+sim.init_grids()
+
+# load a 2 GeV electron beam with an initial
+# unnormalized rms emittance of 2 nm
```
ax3l: Is the emittance correct?
```suggestion
# load a 10 GeV positron beam with an initial
# unnormalized rms emittance of 2 nm
```

examples/positron_channel/README.rst
```diff
@@ -0,0 +1,74 @@
+.. _examples-fodo:
```
ax3l: This needs to be added to the table of contents (ToC) in `docs/source/usage/examples.rst` please, otherwise the doc builder will not pick the file up.

examples/positron_channel/README.rst
```diff
@@ -0,0 +1,74 @@
...
+
+       .. literalinclude:: run_fodo.py
+          :language: python3
+          :caption: You can copy this file from ``examples/fodo/run_fodo.py``.
+
+   .. tab-item:: App Input File
+
+       .. literalinclude:: input_fodo.in
+          :language: ini
+          :caption: You can copy this file from ``examples/fodo/input_fodo.in``.
```
ax3l:
```suggestion
This example can be run as a Python script (``python3 run_positron.py``) or with an app with an input file (``impactx input_positron.in``).
Each can also be prefixed with an `MPI executor <https://www.mpi-forum.org>`__, such as ``mpiexec -n 4 ...`` or ``srun -n 4 ...``, depending on the system.

.. tab-set::

   .. tab-item:: Python Script

       .. literalinclude:: run_positron.py
          :language: python3
          :caption: You can copy this file from ``examples/positron_channel/run_positron.py``.

   .. tab-item:: App Input File

       .. literalinclude:: input_fodo.in
          :language: ini
          :caption: You can copy this file from ``examples/positron_channel/input_positron.in``.
```

examples/positron_channel/README.rst
```diff
@@ -0,0 +1,74 @@
...
+Analyze
+-------
+
+We run the following script to analyze correctness:
+
+.. dropdown:: Script ``analysis_fodo.py``
+
+   .. literalinclude:: analysis_fodo.py
+      :language: python3
+      :caption: You can copy this file from ``examples/fodo/analysis_fodo.py``.
```
ax3l:
```suggestion
.. dropdown:: Script ``analysis_positron.py``

   .. literalinclude:: analysis_positron.py
      :language: python3
      :caption: You can copy this file from ``examples/positron_channel/analysis_positron.py``.
```

examples/positron_channel/README.rst
```diff
@@ -0,0 +1,74 @@
...
+
+.. figure:: https://user-images.githubusercontent.com/1353258/180287840-8561f6fd-278f-4856-abd8-04fbdb78c8ff.png
+   :alt: focusing, defocusing and preserved emittane in our FODO cell benchmark.
+
+   FODO transversal beam width and emittance evolution
+
+.. figure:: https://user-images.githubusercontent.com/1353258/180287845-eb0210a7-2500-4aa9-844c-67fb094329d3.png
+   :alt: focusing, defocusing and phase space rotation in our FODO cell benchmark.
+
+   FODO transversal beam width and phase space evolution
```
ax3l: Can the same visualization script be used as for the regular fodo channel?

At least, we need to update the figures here (Upload in a comment to the PR and then use the link to the image for the URLs here).

examples/positron_channel/README.rst
cemitch99: The previous documentation was only a placeholder.  The documentation has been updated.

examples/positron_channel/README.rst
cemitch99: Done

examples/positron_channel/run_positron.py
```diff
@@ -0,0 +1,76 @@
...
+    elements.ChrQuad(ds=0.1, k=-6.674941, units=1, nslice=ns),
+    elements.ChrDrift(ds=0.1, nslice=ns),
+    elements.ChrAcc(ds=1.8, ez=10871.950994502130424, bz=1.0e-3, nslice=ns),
+    elements.ChrDrift(ds=0.1, nslice=ns),
+    monitor,
+]
+
+num_periods = 250
+
+lattice = period * (num_periods)
```
ax3l:
```suggestion
lattice = period * num_periods
```

cemitch99:
Tests fail because the value of sigma_t in the initial beam differs from the predicted value with an error that is well outside of tolerance.  This is not the case when I run the input file manually on Perlmutter via srun ./impactx ...  Trying to understand whether/why sigma_t should be especially sensitive to the choice of random seed.

The sensitivity occurs because the initial t-pt correlation is large, while the uncorrelated energy spread is small.  As a result, the correlation coefficient rho = -mu_tpt is nearly 1.  In the final beam generation step, the coordinate t is divided by sqrt(1-mu_tpt^2), which is nearly zero, resulting in a strong sensitivity to the random seed. For now, this can be fixed by increasing the uncorrelated energy spread.  In the future, we might think about special handling of the case when mu_tp approx 1.

cemitch99:
Resolved.

examples/positron_channel/run_positron.py
```diff
@@ -0,0 +1,76 @@
...
+    elements.ChrAcc(ds=1.8, ez=10871.950994502130424, bz=1.0e-12, nslice=ns),
+    elements.ChrDrift(ds=0.1, nslice=ns),
+    monitor,
+]
+
+num_periods = 250
+
+lattice = period * num_periods
+
+sim.lattice.extend(lattice)
```
ax3l, line 69: This works and that should do the trick as well, which would be 100% as in the `input_positron.in` file:
```suggestion
sim.lattice.extend(lattice)

# number of periods through the lattice
sim.periods = 250
```

`sim.periods = 250` runs a loop in our evolve function, while the other syntax here creates a 250x longer list (should be no problem either :) )
The current `run_positron.py` file is closer to defining a line in `input_positron.in` and adding `line.repeat = 250`.

ax3l (approved):
Looks awesome, love it :)

examples/positron_channel/run_positron.py
ax3l: Discussed on Slack and decided to modify as such. PR in #373 :)

PR #364, opened 2023-05-22
Fix IOTA nonlinear lens element

cemitch99:
Fixed a bug in the implementation of the IOTA nonlinear lens element.  (Correct the scaling of coordinates x,y.)

Improved the benchmark example to include 18 segments of length 10 cm, analogous to the physical IOTA insert magnet.

Correct scaling of dH_max and dI_max with the number of segments was verified.


ax3l (approved):
Awesome, congrats on finding that bug! :star_struck:

PR #261, opened 2022-09-21
Mesh Refinement

ax3l:
Close #102

This implements one-level static MR.

- [x] implement `sync_rho` from coarse to finer levels
- [x] implement user-friendly input control
- [x] document
- [x] add a test
- [x] generalize to more than one level

Depends on:
- [x] https://github.com/ECP-WarpX/WarpX/pull/3433

PR #367, opened 2023-05-24
Change Dynamic Resizing

ax3l:
Change `geometry.prob_relative` to include the beam extent itself in the definition.

Now, `1.0` is exactly the beam, `3.0` is a simulation box with a size 3x than the beam size (as measured by min/max of the most outermost particles per direction).

Preparation for #261, where we want to define higher MR levels to be potentially cutting into the beam distribution.

- [x] depends on https://github.com/AMReX-Codes/amrex/pull/3328 / #371

### DIFF
--- a/src/initialization/InitMeshRefinement.cpp
+++ b/src/initialization/InitMeshRefinement.cpp
@@ -160,7 +160,7 @@ namespace impactx
                 ablastr::warn_manager::WMRecordWarning(
                     "ImpactX::ResizeMesh",
                     "Dynamic resizing of the mesh uses a geometry.prob_relative "
-                    "with less than 3 the beam size. This might result in boundary "
+                    "with less than 3x the beam size. This might result in boundary "
                     "artifacts for space charge calculation. "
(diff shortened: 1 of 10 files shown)


src/initialization/InitMeshRefinement.cpp
```diff
@@ -153,29 +153,31 @@ namespace impactx
...
+            amrex::Real frac = 3.0;
             pp_geometry.query("prob_relative", frac);
 
-            if (frac < 1.0)
+            if (frac < 3.0)
                 ablastr::warn_manager::WMRecordWarning(
                     "ImpactX::ResizeMesh",
                     "Dynamic resizing of the mesh uses a geometry.prob_relative "
-                    "padding of less than 1. This might result in boundary "
+                    "with less than 3 the beam size. This might result in boundary "
```
RemiLehe:
```suggestion
                    "with less than 3x the beam size. This might result in boundary "
```

RemiLehe (approved):

PR #368, opened 2023-05-24
Implement Lattice Periods

### LINKED ISSUE #338: Implement Turns
Implement an outer loop for turns in a ring.

ax3l:
Implement turns through the lattice.
:ring:

Test:
- [x] 5x turns through the IOTA "bare" linear lattice
- [x] adjust tolerances and name

Close #338

### DIFF
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -219,13 +219,12 @@ Lattice Elements
 * ``lattice.elements`` (``list of strings``) optional (default: no elements)
     A list of names (one name per lattice element), in the order that they appear in the lattice.
 
-* ``lattice.turns`` (``integer``) optional (default: ``1``)
-    The number of turns through the lattice.
-    Note: this does not check if the lattice is periodic.
+* ``lattice.periods`` (``integer``) optional (default: ``1``)
+    The number of periods to repeat the lattice.
 
 * ``lattice.reverse`` (``boolean``) optional (default: ``false``)
     Reverse the list of elements in the lattice.
-    Note: the list of elements is reverted and then turns are applied.
+    If ``reverse`` and ``periods`` both appear, then ``reverse`` is applied before ``periods``.
 
 * ``lattice.nslice`` (``integer``) optional (default: ``1``)
     A positive integer specifying the number of slices used for the application of
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -219,8 +219,13 @@ Lattice Elements
 * ``lattice.elements`` (``list of strings``) optional (default: no elements)
     A list of names (one name per lattice element), in the order that they appear in the lattice.
 
+* ``lattice.turns`` (``integer``) optional (default: ``1``)
+    The number of turns through the lattice.
+    Note: this does not check if the lattice is periodic.
+
 * ``lattice.reverse`` (``boolean``) optional (default: ``false``)
     Reverse the list of elements in the lattice.
+    Note: the list of elements is reverted and then turns are applied.
 
 * ``lattice.nslice`` (``integer``) optional (default: ``1``)
     A positive integer specifying the number of slices used for the application of
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -97,7 +97,7 @@ General
    .. py:property:: periods
 
       The number of periods to repeat the lattice.
-      Note: this does not check if the lattice is periodic.
+  
 
    .. py:property:: abort_on_warning_threshold
 
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -94,6 +94,11 @@ General
       Access the elements in the accelerator lattice.
       See :py:mod:`impactx.elements` for lattice elements.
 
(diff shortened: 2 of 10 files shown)


examples/iota_lattice/analysis_iotalattice.py
```diff
@@ -78,7 +78,7 @@ def get_moments(beam):
 )
 
 atol = 0.0  # a big number
-rtol = 1.5 * num_particles**-0.5  # from random sampling of a smooth distribution
+rtol = 1.5 * num_particles**-0.5 * 15  # from random sampling of a smooth distribution
```
ax3l: @cemitch99 If we run the iotalattice example for 5 turns instead of 1 - how would you expect the tolerance check to change in terms of error bounds?

I determined this factor 15 by running it locally and checking when the test starts to pass again for 5 turns.

docs/source/usage/parameters.rst
```diff
@@ -219,8 +219,13 @@ Lattice Elements
 * ``lattice.elements`` (``list of strings``) optional (default: no elements)
     A list of names (one name per lattice element), in the order that they appear in the lattice.
 
+* ``lattice.turns`` (``integer``) optional (default: ``1``)
+    The number of turns through the lattice.
+    Note: this does not check if the lattice is periodic.
```
cemitch99: I think this comment could lead to confusion.  If lattice.turns > 1, then the particles are being tracked through a periodic lattice by definition (all the elements are repeated).  I suspect that what this means is:  "this does not check whether the reference orbit is periodic (closed)".  We could say that, but we are also interested in periodic structures that are not rings, so that also may cause confusion.

examples/iota_lattice/analysis_iotalattice.py
cemitch99: If the beam is well-matched (as it was intended), then the error should remain small--this seems too large.  Maybe the initial beam moments need to be tuned slightly (which I would probably do in a follow-up PR).  What about the reference orbit?  It should be periodic and return to its original value after 5 turns.

docs/source/usage/parameters.rst
ax3l: Thank you, yes I mean to write that there is no closed orbit check. Will remove, good point.

ax3l: We can also call this `lattice.periods` - depending what is more common lingo...?

See also: https://github.com/ECP-WarpX/impactx/pull/366 (currently with line properties called `repeat` and `revert`).

examples/iota_lattice/analysis_iotalattice.py
cemitch99: Ah - this comment does not apply to sig_t, because there is no longitudinal focusing.  I suggest removing sig_t from the analysis of the final beam.  Then the tolerance can be reduced.

cemitch99:
```suggestion
rtol = 1.5 * num_particles**-0.5   # from random sampling of a smooth distribution
```

examples/iota_lattice/analysis_iotalattice.py
```diff
@@ -78,7 +78,7 @@ def get_moments(beam):
...
 print(f"  rtol={rtol} (ignored: atol~={atol})")
 
 assert np.allclose(
```
cemitch99, line 84: Suggest deleting sigt below, along with the corresponding numerical value.  The test should then pass.  (I hope!)

docs/source/usage/parameters.rst
cemitch99: I don't really know, but I think `lattice.periods` is probably clearer because it does not necessarily suggest bending.

examples/iota_lattice/analysis_iotalattice.py
```diff
@@ -78,15 +78,14 @@ def get_moments(beam):
 )
 
 atol = 0.0  # a big number
-rtol = 1.5 * num_particles**-0.5  # from random sampling of a smooth distribution
+rtol = 1.8 * num_particles**-0.5  # from random sampling of a smooth distribution
```
ax3l, line 81: Bumped the tolerance slightly after removing `sigt`. Now it passes.

Is the range still reasonable to you?

docs/source/usage/python.rst
```diff
@@ -94,6 +94,11 @@ General
       Access the elements in the accelerator lattice.
       See :py:mod:`impactx.elements` for lattice elements.
 
+   .. py:property:: periods
+
+      The number of periods to repeat the lattice.
+      Note: this does not check if the lattice is periodic.
```
```suggestion
  
```

examples/iota_lattice/analysis_iotalattice.py
cemitch99: Yes.

cemitch99 (commented):
This looks good to me.  If we want to keep the cyclotron example, I can help with the analysis next week.

ax3l:
Thank you!

>  If we want to keep the cyclotron example, I can help with the analysis next week.

Ooops, I unintentionally committed that in #371
Thank you, yes  let us design it proper next week, I did not intentionally add that file :sweat_smile:

PR #369, opened 2023-05-24
Fix: openPMD Output Offset

ax3l:
Reset the offset counter between outputs.

First seen as an issue in #368


src/particles/elements/diagnostics/openPMD.cpp
```diff
@@ -226,12 +226,11 @@ namespace detail
 
         // calculate & update particle offset in MPI-global particle array, per level
         auto const num_levels = pc.finestLevel() + 1;
-        m_offset.reserve(num_levels);
```
ax3l, line 229: This did not reset the `size` to zero.

PR #372, opened 2023-05-25
Python: More Bindings for Sim & Beam

ax3l:
Add more bindings as needed for debugging and inspecting the simulation in #261.

### DIFF
--- a/src/python/ImpactX.cpp
+++ b/src/python/ImpactX.cpp
@@ -6,13 +6,6 @@
 #include "pyImpactX.H"
 
 #include <ImpactX.H>
-#include <particles/diagnostics/ReducedBeamCharacteristics.H>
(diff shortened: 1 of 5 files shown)


src/python/ImpactX.cpp
```diff
@@ -272,22 +275,22 @@ void init_ImpactX (py::module& m)
...
+        )
+
         .def("particle_container",
              [](ImpactX & ix) -> ImpactXParticleContainer & {
                 return *ix.m_particle_container;
              },
              py::return_value_policy::reference_internal,
              "Access the beam particle container."
         )
-        .def("reduced_beam_characteristics",
```
ax3l, line 285: @n01r @RTSandberg I moved  this now over to the ImpactX particle container after all, where it is more natural with the other methods I added.

PR #380, opened 2023-06-09
openPMD: add software name version

s-sajid-ali:
Add software name and version as attributes to openPMD outputs.

### DIFF
--- a/src/particles/elements/diagnostics/openPMD.cpp
+++ b/src/particles/elements/diagnostics/openPMD.cpp
@@ -9,6 +9,7 @@
  */
 
 #include "openPMD.H"
+#include "ImpactXVersion.H"
 #include "particles/ImpactXParticleContainer.H"
 
 #include <ablastr/particles/IndexHandling.H>
@@ -198,6 +199,7 @@ namespace detail
 #   endif
                 , "adios2.engine.usesteps = true"
             );
+            series.setSoftware("impactx", IMPACTX_VERSION);
             series.setIterationEncoding( series_encoding );
             m_series = series;
             m_unique_series[m_series_name] = series;
--- a/src/particles/elements/diagnostics/openPMD.cpp
+++ b/src/particles/elements/diagnostics/openPMD.cpp
@@ -199,7 +199,7 @@ namespace detail
 #   endif
                 , "adios2.engine.usesteps = true"
             );
-            series.setSoftware("impactx", IMPACTX_VERSION);
+            series.setSoftware("ImpactX", IMPACTX_VERSION);
             series.setIterationEncoding( series_encoding );
             m_series = series;
             m_unique_series[m_series_name] = series;


s-sajid-ali:
Pinging @ax3l for review.

CMakeLists.txt
```diff
@@ -305,6 +313,7 @@ get_source_version(ImpactX ${ImpactX_SOURCE_DIR})
 configure_file(
     ${ImpactX_SOURCE_DIR}/src/ImpactXVersion.H.in
     ${ImpactX_BINARY_DIR}/src/ImpactXVersion.H
+    IMMEDIATE
```
ax3l: What does this do? :)

Have never seen it: https://cmake.org/cmake/help/latest/command/configure_file.html

CMakeLists.txt
```diff
@@ -3,6 +3,14 @@
 cmake_minimum_required(VERSION 3.20.0)
 project(ImpactX VERSION 23.06)
 
+# generate compile_commands.json
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+# and place it in the source directory https://stackoverflow.com/a/60910583
+add_custom_target(
+  copy-compile-commands ALL
+  ${CMAKE_COMMAND} -E copy_if_different
+  ${CMAKE_BINARY_DIR}/compile_commands.json ${CMAKE_CURRENT_LIST_DIR})
```
ax3l: Why do you need the `compile_commands.json`?
Does your IDE not support CMake directly?

ax3l (commented):
Thank you! :+1: 

Do you mind isolating the IDE-related changes and the I/O-related ones in two PRs? :)

CMakeLists.txt
s-sajid-ali: Looks like it was a deprecated argument.

CMakeLists.txt
s-sajid-ali: Maybe there are other ways, but the `compile_commands.json` is a rather convenient method : `https://clangd.llvm.org/config#compileflags`.

There are tools like [Bear](https://github.com/rizsotto/Bear) that can generate `compile_commands.json`, but generating it alongside the build reduces the number of tools needed.

s-sajid-ali:
>Do you mind isolating the IDE-related changes and the I/O-related ones in two PRs? :)

Done: https://github.com/ECP-WarpX/impactx/pull/381

CMakeLists.txt
s-sajid-ali: compile_commands.json is used by the C++ LSP (clangd) to [parse the source files](https://clangd.llvm.org/config#compileflags). Other build systems such as meson/bazel/etc [can also generate it](https://sarcasm.github.io/notes/dev/compilation-database.html#build-systems-and-compilers).

There are tools like [Bear](https://github.com/rizsotto/Bear) that can generate compile_commands.json, but generating it alongside the build reduces the number of tools needed.

s-sajid-ali: compile_commands.json is used by the C++ LSP (clangd) to [parse the source files](https://clangd.llvm.org/design/compile-commands). Other build systems such as meson/bazel/etc [can also generate it](https://sarcasm.github.io/notes/dev/compilation-database.html#build-systems-and-compilers), so it not really `CMake` specific.

There are tools like [Bear](https://github.com/rizsotto/Bear) that can generate compile_commands.json, but generating it alongside the build reduces the number of tools needed.

src/particles/elements/diagnostics/openPMD.cpp
```diff
@@ -198,6 +199,7 @@ namespace detail
 #   endif
                 , "adios2.engine.usesteps = true"
             );
+            series.setSoftware("impactx", IMPACTX_VERSION);
```
ax3l: For excellent spelling :)
```suggestion
            series.setSoftware("ImpactX", IMPACTX_VERSION);
```

ax3l (approved):
Thank you!

Please note the updated spelling I introduced in the last commit :)

PR #381, opened 2023-06-12
Generate compile_commands.json for clangd

s-sajid-ali:
`compile_commands.json` is used by [`clangd` LSP](https://clangd.llvm.org/) to [interpret the source files](https://clangd.llvm.org/design/compile-commands)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,13 +4,16 @@ cmake_minimum_required(VERSION 3.20.0)
 project(ImpactX VERSION 23.06)
 
 # generate compile_commands.json
-if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
-    set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
-    # and place it in the source directory https://stackoverflow.com/a/60910583
-    add_custom_target(
-      copy-compile-commands ALL
-      ${CMAKE_COMMAND} -E copy_if_different
-      ${CMAKE_BINARY_DIR}/compile_commands.json ${PROJECT_SOURCE_DIR})
+if(${CMAKE_GENERATOR} STREQUAL "Unix Makefiles" OR ${CMAKE_GENERATOR} STREQUAL
+                                                   "Ninja")
+  if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
+      set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+      # and place it in the source directory https://stackoverflow.com/a/60910583
+      add_custom_target(
+        copy-compile-commands ALL
+        ${CMAKE_COMMAND} -E copy_if_different
+        ${CMAKE_BINARY_DIR}/compile_commands.json ${PROJECT_SOURCE_DIR})
+  endif()
 endif()
 
 include(${ImpactX_SOURCE_DIR}/cmake/ImpactXFunctions.cmake)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,14 @@
 cmake_minimum_required(VERSION 3.20.0)
 project(ImpactX VERSION 23.06)
 
+# generate compile_commands.json
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+# and place it in the source directory https://stackoverflow.com/a/60910583
+add_custom_target(
+  copy-compile-commands ALL
+  ${CMAKE_COMMAND} -E copy_if_different
+  ${CMAKE_BINARY_DIR}/compile_commands.json ${CMAKE_CURRENT_LIST_DIR})
+
 include(${ImpactX_SOURCE_DIR}/cmake/ImpactXFunctions.cmake)
 
 # In-source tree builds are messy and can screw up the build system.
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -3,6 +3,14 @@
 cmake_minimum_required(VERSION 3.20.0)
 project(ImpactX VERSION 23.06)
 
+# generate compile_commands.json
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+# and place it in the source directory https://stackoverflow.com/a/60910583
+add_custom_target(
+  copy-compile-commands ALL
+  ${CMAKE_COMMAND} -E copy_if_different
+  ${CMAKE_BINARY_DIR}/compile_commands.json ${CMAKE_CURRENT_LIST_DIR})
```
ax3l: Let's only change this if we are the main project, so superbuilds are not affected
```suggestion
if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
    set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
    # and place it in the source directory https://stackoverflow.com/a/60910583
    add_custom_target(
      copy-compile-commands ALL
      ${CMAKE_COMMAND} -E copy_if_different
      ${CMAKE_BINARY_DIR}/compile_commands.json ${PROJECT_SOURCE_DIR})
endif()
```

Is placing this in the source directory best practice?

s-sajid-ali:
>Is placing this in the source directory best practice?

Per the [docs](https://clangd.llvm.org/design/compile-commands#where-do-compile-commands-come-from): 
```
We first check for a compilation database in the directory containing the source file,
and then walk up into parent directories until we find one.
```

ax3l:
Clang on Windows seems happy but MSVC is not yet:
```
     4>C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VC\v170\Microsoft.CppCommon.targets(248,5): error MSB8066: Custom build for 'D:\a\impactx\impactx\build\CMakeFiles\c134c98d71a9573b76308dc0d94c59ed\copy-compile-commands.rule;D:\a\impactx\impactx\CMakeLists.txt' exited with code 1. [D:\a\impactx\impactx\build\copy-compile-commands.vcxproj]
```

Actually, how does this handle MultiConfig generators?
https://cmake.org/cmake/help/latest/prop_gbl/GENERATOR_IS_MULTI_CONFIG.html

CMakeLists.txt
```diff
@@ -3,6 +3,16 @@
...
 project(ImpactX VERSION 23.06)
 
+# generate compile_commands.json
+if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
+    set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+    # and place it in the source directory https://stackoverflow.com/a/60910583
+    add_custom_target(
+      copy-compile-commands ALL
+      ${CMAKE_COMMAND} -E copy_if_different
+      ${CMAKE_BINARY_DIR}/compile_commands.json ${PROJECT_SOURCE_DIR})
```
ax3l: This has to be guarded for all [build generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#cmake-generators) that are not GNUmake or Ninja:
https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html

> Note
>
> This option is implemented only by [Makefile Generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#makefile-generators) and [Ninja Generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#ninja-generators). It is ignored on other generators.
>
> This option currently does not work well in combination with the [UNITY_BUILD](https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html#prop_tgt:UNITY_BUILD) target property or the [CMAKE_UNITY_BUILD](https://cmake.org/cmake/help/latest/variable/CMAKE_UNITY_BUILD.html#variable:CMAKE_UNITY_BUILD) variable.

ax3l:
> > Is placing this in the source directory best practice?
>
> Per the [docs](https://clangd.llvm.org/design/compile-commands#where-do-compile-commands-come-from):
>
> We first check for a compilation database in the directory containing the source file,
> and then walk up into parent directories until we find one.

I wonder if they mean they also look recursively into directories inside source, e.g., in `build/`.

Generally, I do not see how this can work with multi-config generators at all - every config needs its own `compile_commands.json`.

Clion's IDE solves this a bit smarter, they are aware of the build configs, generate a file in the respective build directory and run clangd on that. Is there a way in your IDE to configure this equally? Which IDE are you using? Or did you configure clangd manually to work with your toolchain?

CMakeLists.txt
s-sajid-ali: Added a guard to only generate `compile_commands.json` for `Unix Makefiles` and `Ninja` generators.

s-sajid-ali:
>Generally, I do not see how this can work with multi-config generators at all - every config needs its own compile_commands.json.

Upon further thought, I agree that placing it in the build directory is a better solution than placing it in the source directory. I've removed the command that copies it to the source directory.

s-sajid-ali:
>Which IDE are you using? Or did you configure clangd manually to work with your toolchain?
 
 I'm using [`neovim`](https://neovim.io/) which comes with a built-in LSP client that I've configured to use `clangd`. 
 
 >Clion's IDE solves this a bit smarter, they are aware of the build configs, generate a file in the respective build directory and run clangd on that. Is there a way in your IDE to configure this equally? 

Maybe there is, but I'm not aware of it. 

If we place `compile_commands.json` only in the build directory, the only restriction would be to browse from the build directory to get the LSP to work. Sometimes if the project is built with different options turned on/off and there are `#ifdef`s, the code that gets highlighted changes with the build configuration. Hence it might be preferable to open the editor from the build directory rather than the source directory.

ax3l:
The better solution might be to set in an IDE that needs it `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON`

PR #382, opened 2023-06-14
Doc: Update Desktop Pkg Managers

ax3l:
Including OpenMP libs for macOS.

### DIFF
--- a/docs/source/install/cmake.rst
+++ b/docs/source/install/cmake.rst
@@ -36,7 +36,7 @@ ImpactX depends on popular third party software.
    .. code-block:: bash
 
       python3 -m pip install -U pip setuptools wheel pytest
-      python3 -m pip install -r examples/requirements.txt
+      python3 -m pip install -U -r examples/requirements.txt
 
 
 Compile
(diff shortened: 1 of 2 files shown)


cemitch99 (approved):
Excellent.  This worked for me (using conda to install on a MacBook M2 with OpenMP).

Warning:  For those who also want to use gcc (eg, for gfortran), a Conda install of gcc may cause problems on a Mac running Ventura:  
[https://stackoverflow.com/questions/75263306/gfortran-does-not-work-properly-when-conda-is-installed-on-macos-ventura](https://stackoverflow.com/questions/75263306/gfortran-does-not-work-properly-when-conda-is-installed-on-macos-ventura) 
[https://github.com/ContinuumIO/anaconda-issues/issues/13087](https://github.com/ContinuumIO/anaconda-issues/issues/13087)
The workaround--an install of gcc using Homebrew--appears to cause conflict with the Conda packages required here.

ax3l:
Great! :tada: 

For conda, the `compilers` package should download a vanilla `clang` for you, which then is compatible with the other dependencies in the virtual environment.

You need to have the conda environment active and remove your previous `build/` directory to make sure the new compilers are picked up.

PR #384, opened 2023-06-20
Doc: IPAC23 Paper

ax3l:
Document the IPAC23 paper, which uses WarpX to train an ML model for ImpactX.


cemitch99 (approved):
Looks good to me.

PR #388, opened 2023-06-27
CI: macOS CMake

ax3l:
Not sure why it does not work as in WarpX, so we pull CMake in with `pipx` now.


ax3l:
issue fixed by gh image update on runners.

PR #393, opened 2023-06-29
[WIP] Add GSI space charge benchmark lattice

cemitch99:
Added the lattice used for space charge benchmarks described here:
http://web-docs.gsi.de/~giuliano/research_activity/trapping_benchmarking/main.html

- [ ] The problem parameters need to be double-checked. In particular, verify the longitudinal dynamics.
- [x] Implement first for a 2.5D space charge solver? #1214
- [x] test of linear (envelope) optics w/o space charge, w/o sextupole
- [x] test of orbits (resonance behavior) w/o space charge, w/ sextupole
- [x] test of tune calculation w/o space charge, w/o sextupole
- [x] test of tune calculation w/ space charge, w/o sextupole
- [x] test of tune calculation w/ space charge, w/o sextupole using Gauss2p5d solver
- [x] tests w/ space charge, w/ sextupole

There is much that can be useful here.  For now, data will be archived here, and we can assess later what is most useful to leave in the examples directory.  I will move the status to WIP for the time being.


cemitch99:
This was sitting on the back shelf of the "to do" list.  Now it's time to revisit.

examples/gsi_benchmark/run_gsilattice_in_situ_tunes.py
```diff
@@ -0,0 +1,146 @@
...
+for idcpu, particle_track in grouped_by_idcpu:
+    xarr = np.array(particle_track["x"])
+    pxarr = np.array(particle_track["px"])
+    xn = xarr / np.sqrt(beta_x)
+    pxn = pxarr * np.sqrt(beta_x) + xarr * alpha_x / np.sqrt(beta_x)
+    z = xn - 1j * pxn
+    # To plot the normalized data:
+    # plt.scatter(xn, pxn, s=5)
+    output = pnf.naff(z, turns=n, nterms=4, skipTurns=0, getFullSpectrum=True, window=1)
+    tune = output[0, 1]
```
ax3l: Analysis: since we only read `output[0, 1]` we can be 3.4x faster when doing:
```suggestion
    output = pnf.naff(z, turns=n, nterms=1, skipTurns=0, getFullSpectrum=True, window=1)
    tune = output[0, 1]
```
Each extra term is another FFT + the expensive  #`frefin` refinement loop that we do not use.

examples/gsi_benchmark/run_gsilattice_in_situ_tunes.py
```diff
@@ -0,0 +1,146 @@
...
+# loop over one particle track at a time
+for idcpu, particle_track in grouped_by_idcpu:
+    xarr = np.array(particle_track["x"])
+    pxarr = np.array(particle_track["px"])
+    xn = xarr / np.sqrt(beta_x)
+    pxn = pxarr * np.sqrt(beta_x) + xarr * alpha_x / np.sqrt(beta_x)
+    z = xn - 1j * pxn
+    # To plot the normalized data:
+    # plt.scatter(xn, pxn, s=5)
+    output = pnf.naff(z, turns=n, nterms=4, skipTurns=0, getFullSpectrum=True, window=1)
```
ax3l: Claude:
`turns=n` is an off-by-one. PyNAFF needs `len(data) ≥ turns+1`. When `n` is a multiple of 6 it crashes , otherwise it silently rounds turns down to a multiple of 6 and drops up to 5 turns. Use turns=`len(z)-1`.

examples/gsi_benchmark/run_gsilattice_in_situ_tunes.py
ax3l: Claude bug comment:
  - You never remove the DC/closed-orbit component. With an offset and nterms=1, naff returned an empty array  (0,) → output[0,1] would IndexError. After z -= z.mean() it correctly returned tune 0.317. (This is likely  why you reached for nterms=4 — it was masking the DC line. Subtract the mean and nterms=1 is both faster   and robust.)

examples/gsi_benchmark/run_gsilattice_in_situ_tunes.py
```diff
@@ -0,0 +1,146 @@
...
+    tune = output[0, 1]
+    plt.scatter(xarr[0] * mm_scale, tune)
+
+plt.xlabel("x [mm]", fontsize=12)
```
ax3l, line 155: Claude suggested improvements: 2 bug fixes, small perf improvement, works with pyNAFF <1.2.0, too:
```suggestion
beta_x = 12.79711091
sqrt_beta = np.sqrt(beta_x)              # hoist out of the loop

x0_mm, tunes = [], []                    # accumulate, plot once
for idcpu, track in grouped_by_idcpu:
    xarr  = track["x"].to_numpy()        # cheaper than np.array(track["x"])
    pxarr = track["px"].to_numpy()
    
    xn  = xarr / sqrt_beta
    pxn = pxarr * sqrt_beta + xarr * alpha_x / sqrt_beta
    z   = xn - 1j * pxn
    z  -= z.mean()                        # remove closed-orbit / DC line
    
    out = pnf.naff(
        z,
        turns=len(z) - 1,                 # FIX: turns+1 must be <= len(z)
        nterms=1,                         # FIX: you only use term 0  -> 3.4x faster
        getFullSpectrum=True,             # required: z is complex
        window=1,
        warnings=False,
    )
    if len(out) == 0:                     # no line (e.g. lost/clipped particle)
        continue
    x0_mm.append(xarr[0] * mm_scale)
    tunes.append(out[0, 1])

plt.scatter(x0_mm, tunes, s=5)            # single draw call (was 1 per particle)
plt.xlabel("x [mm]", fontsize=12)
```

examples/gsi_benchmark/run_gsilattice_in_situ_tunes.py
ax3l: I added a full-change suggestion below, so do _not_ apply this one-liner here please.

cemitch99: Got it.  All looks reasonable to me--I'll accept here, and then run some local tests.

cemitch99: Note:  These Claude-suggested changes do not appear to affect the output.

examples/gsi_benchmark/distribution_gsi.py
```diff
@@ -0,0 +1,46 @@
+from impactx import distribution, twiss
+
+
+def get_reference_params():
+    # beam parameters
+    kin_energy_MeV = 11.4  # reference kinetic energy
+    # bunch_charge_C = 4.4e-9  #This works with 10K particles.  With 100K particles, it is too high.
+    # bunch_charge_C = 3.7e-9  #4 nC = 0.24, 3.5 nC = 0.26..)
+    bunch_charge_C = 4.2e-9  # Try this with 100K particles. 3.7 nC gives 0.26, increase. 4 nC gives 2.55, increase
+    beam_current_A = 1.5  # used with 2D space charge
```
cemitch99:
```suggestion
```

PR #396, opened 2023-07-05
Store additional beam metadata

RemiLehe:
In offline discussions with @s-sajid-ali, it was proposed to add `particle_mass, beta_ref, gamma_ref` as attributes in the openPMD file, in order to make the ImpactX output readable by Synergia.

### DIFF
--- a/src/particles/elements/diagnostics/openPMD.cpp
+++ b/src/particles/elements/diagnostics/openPMD.cpp
@@ -257,6 +257,11 @@ namespace detail
             }
         }
 
+        // beam mass
+        beam.setAttribute( "mass", ref_part.mass );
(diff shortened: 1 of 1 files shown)


cemitch99 (approved):
This looks good to me.

ax3l:
This is great, thank you!

We can as a follow-up generalize this to use the openPMD BeamPhysics extension by @ChristopherMayes @DavidSagan :star2: 
https://github.com/openPMD/openPMD-standard/blob/upcoming-2.0.0/EXT_BeamPhysics.md

DavidSagan:
Note: the BeamPhysics extension uses the species instead of particle mass, totalMomentumOffset instead of a ref beta or ref gamma.

Issue #394, opened 2023-06-30
Apertures & Lost Particle Book Keeping


ax3l:
We can add apertures to ImpactX.

Programmatically, we will set particles that are to be removed to a negative `id`.
They will then the removed during the communication step by AMReX.

If we want to keep track of lost particles (6D position of particle in lattice when removed), we would process these particles before the communication is called. This is a useful functionality to have as well.

Mad-X reference:
- radially symmetric
- rectangular
- elliptical

cemitch99:
Three options for application of the aperture boundary:

1) A thin "aperture" element, whose purpose is to compare particles against the aperture boundary (available in TraceWin, MAD-X)
2) Adding aperture restrictions to all existing elements (currently used in IMPACT-Z)
3) Allowing the user to provide a file of the form (s, xmax, ymax, type) specifying the aperture boundary vs s.

ax3l:
General support for option 1 added in #398, part of release `23.12`.

GPU support tracked in #499

PR #398, opened 2023-07-12
Add aperture element

cemitch99:
Added a first draft of collimator/aperture element.  For now, this element simply compares each particle against an aperture boundary (rectangular or elliptical), and records whether the particle has been lost.

For this initial draft, coordinates of lost particles are updated to x = y = 0 (so that all particles lie within the aperture).  This will be changed to modify the sign of the particle index.

- [x] Add new element type (inputs, etc.)
- [x] Update documentation
- [x] Modify particle index handling
- [x] Add benchmark example
- [x] Verify correct index handling
- [x] move particles to a separate particle container and enable output for them #394
  - [x] finalize move (see todos)
  - [x] add `s` of reference particle at point of particle loss to "lost" particle container/species
  - [x] output

*To be followed-up with additional bookkeeping of lost particles.
Metadata to add:
- `s`
- turn number

### DIFF
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -520,13 +520,13 @@ Lattice Elements
         * ``aperture`` for a thin collimator element applying a transverse aperture boundary.
           This requires these additional parameters:
 
-            * ``<element_name>.shape`` (``integer``, dimensionless) shape of the aperture boundary
-
-                (m = 0) rectangular, (m = 1) elliptical
-
             * ``<element_name>.xmax`` (``float``, in meters) maximum value of the horizontal coordinate
 
             * ``<element_name>.ymax`` (``float``, in meters) maximum value of the vertical coordinate
+
+            * ``<element_name>.shape`` (``integer``, dimensionless) shape of the aperture boundary
+
+                (m = 0) rectangular, (m = 1) elliptical
 
         * ``beam_monitor`` a beam monitor, writing all beam particles at fixed ``s`` to openPMD files.
           If the same element name is used multiple times, then an output series is created with multiple outputs.
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -524,9 +524,7 @@ Lattice Elements
 
             * ``<element_name>.ymax`` (``float``, in meters) maximum value of the vertical coordinate
 
-            * ``<element_name>.shape`` (``integer``, dimensionless) shape of the aperture boundary
-
-                (m = 0) rectangular, (m = 1) elliptical
+            * ``<element_name>.shape`` (``string``) shape of the aperture boundary: ``rectangular`` (default) or ``elliptical``
 
         * ``beam_monitor`` a beam monitor, writing all beam particles at fixed ``s`` to openPMD files.
           If the same element name is used multiple times, then an output series is created with multiple outputs.
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -509,7 +509,6 @@ Lattice Elements
 
             * ``<element_name>.units`` (``string``) specification of units: ``dimensionless`` (default, in units of the magnetic rigidity of the reference particle) or ``T-m``
 
-
         * ``thin_dipole`` for a thin dipole element.
           This requires these additional parameters:
 
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -509,12 +509,24 @@ Lattice Elements
 
             * ``<element_name>.units`` (``string``) specification of units: ``dimensionless`` (default, in units of the magnetic rigidity of the reference particle) or ``T-m``
 
+
         * ``thin_dipole`` for a thin dipole element.
           This requires these additional parameters:
 
             * ``<element_name>.theta`` (``float``, in degrees) dipole bend angle
 
             * ``<element_name>.rc`` (``float``, in meters) effective radius of curvature
+
+        * ``aperture`` for a thin collimator element applying a transverse aperture boundary.
+          This requires these additional parameters:
+
+            * ``<element_name>.shape`` (``integer``, dimensionless) shape of the aperture boundary
+
+                (m = 0) rectangular, (m = 1) elliptical
+
+            * ``<element_name>.xmax`` (``float``, in meters) maximum value of the horizontal coordinate
+
+            * ``<element_name>.ymax`` (``float``, in meters) maximum value of the vertical coordinate
 
         * ``beam_monitor`` a beam monitor, writing all beam particles at fixed ``s`` to openPMD files.
           If the same element name is used multiple times, then an output series is created with multiple outputs.
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -701,6 +701,11 @@ Diagnostics and output
 * ``diag.file_min_digits`` (``integer``, optional, default: ``6``)
     The minimum number of digits used for the step number appended to the diagnostic file names.
 
+* ``diag.backend`` (``string``, default value: ``default``)
+
+  Diagnostics for particles lost in apertures, stored as ``diags/openPMD/particles_lost.*`` at the end of the simulation.
+  See the ``beam_monitor`` element for backend values.
+
 .. _running-cpp-parameters-diagnostics-reduced:
 
 Reduced Diagnostics
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -116,6 +116,11 @@ General
       The minimum number of digits (default: ``6``) used for the step
       number appended to the diagnostic file names.
 
+   .. py:property:: particle_lost_diagnostics_backend
+
+      Diagnostics for particles lost in apertures.
+      See the ``BeamMonitor`` element for backend values.
+
    .. py:method:: init_grids()
 
       Initialize AMReX blocks/grids for domain decomposition & space charge mesh.
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -712,13 +712,13 @@ References:
    :param phi_in: angle of the reference particle with respect to the longitudinal (z) axis in the original frame in degrees
    :param phi_out: angle of the reference particle with respect to the longitudinal (z) axis in the rotated frame in degrees
 
-.. py:class:: impactx.elements.Aperture(xmax, ymax, shape=0)
+.. py:class:: impactx.elements.Aperture(xmax, ymax, shape="rectangular")
 
    A thin collimator element, applying a transverse aperture boundary.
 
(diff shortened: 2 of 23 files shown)


src/particles/elements/Aperture.H
```diff
@@ -0,0 +1,116 @@
...
+                [[maybe_unused]] amrex::ParticleReal & AMREX_RESTRICT py,
+                [[maybe_unused]] amrex::ParticleReal & AMREX_RESTRICT pt,
+                [[maybe_unused]] RefPart const & refpart) const {
+
+            using namespace amrex::literals; // for _rt and _prt
+
+            // access AoS data such as positions and cpu/id
+            amrex::ParticleReal const x = p.pos(RealAoS::x);
+            amrex::ParticleReal const y = p.pos(RealAoS::y);
+            //uint64_t const global_id = ablastr::particles::localIDtoGlobal(p.id(), p.cpu());
```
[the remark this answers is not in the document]
cemitch99: This line will probably be needed eventually.

src/python/elements.cpp
```diff
@@ -359,4 +359,16 @@ void init_elements(py::module& m)
         )
     ;
 
+    py::class_<Aperture, elements::Thin>(me, "Aperture")
+        .def(py::init<
+        int,
+                amrex::ParticleReal const,
+                amrex::ParticleReal const>(),
+             py::arg("shape") = 0, py::arg("xmax"), py::arg("ymax"),
```
ax3l: I think that defaulted arguments need to come last in C++ and Python.

docs/source/usage/python.rst
```diff
@@ -651,6 +651,14 @@ References:
    :param phi_in: angle of the reference particle with respect to the longitudinal (z) axis in the original frame in degrees
    :param phi_out: angle of the reference particle with respect to the longitudinal (z) axis in the rotated frame in degrees
 
+.. py:class:: impactx.elements.Aperture(shape=0, xmax, ymax)
```
ax3l: I think that defaulted arguments need to come last in C++ and Python.

examples/aperture/README.rst
```diff
@@ -0,0 +1,47 @@
+.. _examples-aperture:
+
+Aperture collimation
+=====================
```
ax3l:
```suggestion
Aperture Collimation
====================
```

examples/aperture/run_aperture.py
```diff
@@ -0,0 +1,62 @@
...
+sim.add_particles(bunch_charge_C, distr, npart)
+
+# add beam diagnostics
+monitor = elements.BeamMonitor("monitor", backend="h5")
+
+# design the accelerator lattice
+sim.lattice.extend(
+    [
+        monitor,
+        elements.Aperture(shape=0, xmax=1.0e-3, ymax=1.5e-3),
```
ax3l: I think that defaulted arguments need to come last in C++ and Python.
```suggestion
        elements.Aperture(xmax=1.0e-3, ymax=1.5e-3, shape=0),
```

src/initialization/InitElement.cpp
```diff
@@ -224,7 +224,14 @@ namespace detail
...
             m_lattice.emplace_back( ChrAcc(ds, ez, bz, nslice) );
-        } else if (element_type == "beam_monitor") {
+        } else if (element_type == "aperture") {
+            amrex::Real xmax, ymax;
+            int shape = 0;
+            pp_element.get("xmax", xmax);
+        pp_element.get("ymax", ymax);
+            pp_element.queryAdd("shape", shape);
+            m_lattice.emplace_back( Aperture(shape, xmax, ymax) );
+    } else if (element_type == "beam_monitor") {
```
```suggestion
            pp_element.get("ymax", ymax);
            pp_element.queryAdd("shape", shape);
            m_lattice.emplace_back( Aperture(xmax, ymax, shape) );
        } else if (element_type == "beam_monitor") {
```

src/particles/elements/All.H
```diff
@@ -40,7 +41,8 @@ namespace impactx
 {
     using KnownElements = std::variant<
         None, /* must be first, so KnownElements creates a default constructor */
-        CFbend,
+    Aperture,
+    CFbend,
         ChrAcc,
```
ax3l:
```suggestion
        Aperture,
        CFbend,
        ChrAcc,
```

src/particles/elements/All.H
```diff
@@ -55,7 +57,7 @@ namespace impactx
         Programmable,
         PRot,
         Quad,
-        RFCavity,
+    RFCavity,
```
ax3l:
```suggestion
        RFCavity,
```

src/particles/elements/Aperture.H
```diff
@@ -0,0 +1,119 @@
...
+        /** A thin collimator element that applies a transverse aperture boundary.
+         *  Particles outside the boundary are considered lost.
+         *
+         * @param shape aperture shape (0 rectangular, 1 elliptical)
+         * @param xmax maximum value of horizontal coordinate (m)
+         * @param ymax maximum value of vertical coordinate (m)
+         */
+        Aperture( int const shape,
+                   amrex::ParticleReal const xmax,
+                   amrex::ParticleReal const ymax )
```
ax3l: We should move `shape` last here, too.

I would suggest we can make it an enum class or string instead of an int, because then the option is self-describing at its call sites.

src/particles/elements/Aperture.H
```diff
@@ -0,0 +1,119 @@
...
+
+            // scale horizontal and vertical coordinates
+            amrex::ParticleReal const u = x/m_xmax;
+            amrex::ParticleReal const v = y/m_ymax;
+
+            // compare against the aperture boundary
+
+            switch(m_shape) {
+
+           // rectangular aperture (default)
```
ax3l: Not super important yet, but since only one option can be true at the same time, we potentially want to move this switch decision out of kernel code, where it increases register pressure, and into a host-side decision.

For this, we could write a general aperture base class and derive it into special rectangular and elliptical classes.

Anyway, I think the part here is too small to warrant this just yet and this will be fine.

ax3l (commented):
Thank you for the draft of the aperture element, this is great!

Why do you set the id negative and the position to zero?
If we only set the id to negative, we can keep the transversal position where the particle was lost as information.

Generally, particles with negative ids will be removed on the next communication call. I would propose we add a filter step, that moves all negative-ly id's particles to a new species, so we can keep track and add the step and `s` where the particle was lost?

cemitch99:
I set the transverse position to zero in the initial draft so that the test passes (all particles in the output lie inside the aperture).  This should be removed once we add the step of moving the particles with negative index.

docs/source/usage/python.rst
cemitch99: Ok, I will change the argument order.

src/particles/CollectLost.cpp
```diff
@@ -0,0 +1,58 @@
...
+                PType *AMREX_RESTRICT aos_ptr = aos().dataPtr();
+
+                amrex::ParallelFor(np, [=] AMREX_GPU_DEVICE(long i) {
+                    PType &p = aos_ptr[i];
+                    p.id() = -p.id();
+                });
+            }
+        }
+
+        // TODO remove particles with negative ids in source
```
ax3l: Need to find an efficient way to do this. (Note: No need to do any comms for this step.)

cc @WeiqunZhang @atmyers :)

src/particles/elements/Aperture.H
ax3l: done :)

src/particles/CollectLost.cpp
ax3l: @atmyers recommended the `filterAndTransform` helpers that operate on a tile:
https://amrex-codes.github.io/amrex/doxygen/namespaceamrex.html#af6ec82174b9e85180e43d20c64b52b26

src/particles/CollectLost.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+                // we will copy particles that were marked as lost, with a negative id
+                auto predicate = [](const SrcData& src, int ip, const amrex::RandomEngine& /*engine*/) noexcept
+                {
+                    return src.id(ip) < 0;
+                };
+
+                // FIXME This seems to segfault
+                auto& ptile_dest = dest.DefineAndReturnParticleTile(
+                        lev, pti.index(), pti.LocalTileIndex());
```
ax3l: This seems to segfault.
Maybe the destination container is not properly initialized yet?

WeiqunZhang: Can you print out `source.finestLevel()` and `dest.finestLevel()`?

ax3l: Both fail for a non-MR sim on level 0:
```
src.finest=0
dest.finest=0
Segfault
```

src/particles/CollectLost.cpp
ax3l: Added now in `CollectLost.cpp`

src/particles/CollectLost.cpp
ax3l: We found it - dependency order issue between `AddRealComp` (called too early), `InitFromScratch` and `reserveData`, `resizeData`.

src/particles/CollectLost.cpp
```diff
@@ -0,0 +1,150 @@
...
+                auto const np = ptile_source.numParticles();
+                if (np == 0) continue;  // no particles in source tile
+
+                // we will copy particles that were marked as lost, with a negative id
+                auto predicate = [](const SrcData& src, int ip, const amrex::RandomEngine& /*engine*/) noexcept
+                {
+                    return src.id(ip) < 0;
+                };
+
+                // FIXME This seems to segfault
```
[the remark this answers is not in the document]
ax3l: fixed

cemitch99 (commented):
The bookkeeping changes look good to me.

docs/source/usage/parameters.rst
```diff
@@ -509,13 +509,23 @@ Lattice Elements
 
             * ``<element_name>.units`` (``string``) specification of units: ``dimensionless`` (default, in units of the magnetic rigidity of the reference particle) or ``T-m``
 
+
```
```suggestion
```

examples/CMakeLists.txt
```diff
@@ -619,6 +619,7 @@ add_impactx_test(kicker.py
     examples/kicker/analysis_kicker.py
     OFF  # no plot script yet
 )
+
```
ax3l:
```suggestion
```

examples/aperture/run_aperture.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022-2023 ImpactX contributors
+# Authors: Marco Garten, Axel Huebl, Chad Mitchell
```
ax3l:
```suggestion
# Authors: Axel Huebl, Chad Mitchell
```

ax3l (approved):

PR #402, opened 2023-07-21
Re-initialize geometry object(s) in initGrids

atmyers:
This will allow us to delay setting the `amr.n_cells` parameter until *after* the ImpactX instance is constructed.

### DIFF
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -50,6 +50,28 @@ namespace impactx
     {
         BL_PROFILE("ImpactX::initGrids");
 
+        // n_cells has been set using dummy values earlier. We now know the true value of
+        // n_cells, so we recompute the Geometry objects for each level here *if* the user
+        // has set n_cells in the inputs file
+        {
+            amrex::Vector<int> n_cell(AMREX_SPACEDIM);
+            amrex::ParmParse pp_amr("amr");
+            pp_amr.queryarr("n_cell", n_cell);
+
+            amrex::IntVect lo(amrex::IntVect::TheZeroVector()), hi(n_cell);
+            hi -= amrex::IntVect::TheUnitVector();
+            amrex::Box index_domain(lo,hi);
+            for (int i = 0; i <= max_level; i++)
+            {
+                geom[i].Domain(index_domain);
+                if (i < max_level) {
+                    index_domain.refine(ref_ratio[i]);
+                }
+            }
+        }
+
+        // the particle container has been set to track the same Geometry as ImpactX
+
         // this is the earliest point that we need to know the particle shape,
         // so that we can initialize the guard size of our MultiFabs
         m_particle_container->SetParticleShape();
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -50,7 +50,7 @@ namespace impactx
     {
         BL_PROFILE("ImpactX::initGrids");
 
-        // n_cells has been set using dummy values earlier. We now know the true value of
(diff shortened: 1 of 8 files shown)


src/initialization/InitAmrCore.cpp
```diff
@@ -88,9 +88,8 @@ namespace details
     {
         amrex::ParmParse pp_amr("amr");
 
-        // Domain index space
-        amrex::Vector<int> n_cell;
-        pp_amr.getarr("n_cell", n_cell);
+        // Domain index space - we use dummy values here, then fix later
+        amrex::Vector<int> n_cell = {AMREX_D_DECL(256,256,256)};
```
ax3l, line 92: Do these values work with any MPI size?

src/initialization/InitAmrCore.cpp
```diff
@@ -88,9 +88,8 @@ namespace details
     {
         amrex::ParmParse pp_amr("amr");
 
-        // Domain index space
-        amrex::Vector<int> n_cell;
-        pp_amr.getarr("n_cell", n_cell);
+        // Domain index space - we use dummy values here, then fix later
```
ax3l: dummy -> temporary? :)

src/initialization/InitAmrCore.cpp
atmyers: They should work for any MPI size as long as this is done before `initGrids` - they won't be used for anything other than setting the Geometry objects.

ax3l: Awesome :star_struck: 

And we then overwrite the geometry objects simply? Are they used before `initGrids` for something (domain decomposition or so?).

src/ImpactX.cpp
```diff
@@ -50,6 +50,28 @@ namespace impactx
     {
         BL_PROFILE("ImpactX::initGrids");
 
+        // n_cells has been set using dummy values earlier. We now know the true value of
```
```suggestion
        // n_cells has been set using temporary values earlier. We now know the true value of
```

src/initialization/InitAmrCore.cpp
cemitch99: It looks like user-specification of `n_cell` is optional, with (256,256,256) as the current default.  Do I interpret correctly?

cemitch99 (approved):
This looks good to me.  Elegant solution.

ax3l (approved):
Thank you Andrew, this is great :tada:

src/ImpactX.cpp
```diff
@@ -50,6 +50,28 @@ namespace impactx
...
+        // has set n_cells in the inputs file
+        {
+            amrex::Vector<int> n_cell(AMREX_SPACEDIM);
+            amrex::ParmParse pp_amr("amr");
+            pp_amr.queryarr("n_cell", n_cell);
+
+            amrex::IntVect lo(amrex::IntVect::TheZeroVector()), hi(n_cell);
+            hi -= amrex::IntVect::TheUnitVector();
+            amrex::Box index_domain(lo,hi);
+            for (int i = 0; i <= max_level; i++)
```
ax3l, line 64: Delayed MR init does not yet work with this. Let's revisit after #261 was merged :)

src/initialization/InitAmrCore.cpp
ax3l: Looking at this again, I do not see how this path would be ever taken. If the user does not pre-inject `amr.n_cell` into the ParmParser then the below function `one_box_per_rank` us called.

PR #410, opened 2023-08-01
Add transverse kicker

cemitch99:
Added a thin transverse kicker element, corresponding to MAD-X elements 'hkicker', 'vkicker', 'kicker', and 'tkicker'.

- [x] Add element
- [x] Add Python equivalent
- [x] Add example
- [x] Add documentation


src/particles/elements/Kicker.H
```diff
@@ -0,0 +1,117 @@
...
+            amrex::ParticleReal const x = p.pos(RealAoS::x);
+            amrex::ParticleReal const y = p.pos(RealAoS::y);
+            amrex::ParticleReal const t = p.pos(RealAoS::t);
+
+            // normalize quad units to MAD-X convention if needed
+            amrex::ParticleReal dpx = m_xkick;
+            amrex::ParticleReal dpy = m_ykick;
+            if (m_unit == 1) {
+                  dpx = m_xkick / refpart.rigidity_Tm();
+                  dpy = m_ykick / refpart.rigidity_Tm();
```
ax3l: Since the kick is already assigned:
```suggestion
                  dpx /= refpart.rigidity_Tm();
                  dpy /= refpart.rigidity_Tm();
```

src/python/elements.cpp
```diff
@@ -182,6 +182,16 @@ void init_elements(py::module& m)
         )
     ;
 
+    py::class_<Kicker, elements::Thin>(me, "Kicker")
+        .def(py::init<
+                amrex::ParticleReal const,
+                amrex::ParticleReal const,
+                int const>(),
+             py::arg("xkick"), py::arg("ykick"), py::arg("units") = 0,
+             "A thin transverse kicker element"
```
```suggestion
             "A thin transverse kicker element. Kicks are for unit 0 dimensionless, or for 1 in T-m."
```

ax3l: We probably can make this an enum once we have the example in place :)

src/particles/elements/Kicker.H
```diff
@@ -0,0 +1,117 @@
...
+    private:
+        amrex::ParticleReal m_xkick; //! horizontal kick strength
+        amrex::ParticleReal m_ykick; //! vertical kick strength
+        int m_unit; //! multipole index
+
+    };
+
+} // namespace impactx
+
+#endif // IMPACTX_MULTIPOLE_H
```
ax3l:
```suggestion
#endif // IMPACTX_KICKER_H
```

src/particles/elements/Kicker.H
```diff
@@ -0,0 +1,117 @@
...
+
+        }
+
+        /** This pushes the reference particle. */
+        using Thin::operator();
+
+    private:
+        amrex::ParticleReal m_xkick; //! horizontal kick strength
+        amrex::ParticleReal m_ykick; //! vertical kick strength
+        int m_unit; //! multipole index
```
ax3l:
```suggestion
        int m_unit; //! Kicks are for 0 dimensionless, or for 1 in T-m."
```

src/particles/elements/Kicker.H
```diff
@@ -0,0 +1,117 @@
...
+                int const unit )
+        : m_xkick(xkick), m_ykick(ykick), m_unit(unit)
+        {
+        }
+
+        /** Push all particles */
+        using BeamOptic::operator();
+
+        /** This is a multipole functor, so that a variable of this type can be used like a
+         *  multipole function.
```
ax3l:
```suggestion
        /** This is a transverse kicker functor, so that a variable of this type can be
         * used like a function.
```

ax3l (approved):
Thank you, that looks great.

I added a small commit using self-describing arguments at the call site for the kicker: a C++ enum and strings when describing the unit system of the Kicker. I also clarified a few details on the normalization of the dimensionless option.

ax3l (approved):

cemitch99 (commented):
These changes look reasonable to me, thanks.  I agree that this should make the choice of units more transparent to the user.

PR #412, opened 2023-08-07
Add MAD-X equivalent short RF model

cemitch99:
Rename the element ShortRF -> Buncher.  (This is a linear element at zero-crossing used only for longitudinal bunching.)

Add a MAD-X equivalent short RF model called "ShortRF".  This element allows for a phase-dependent nonzero longitudinal kick resulting in energy gain.

- [x] Add element and particle push.
- [x] Add reference particle push.
- [x] Add example.
- [x] Update docs.


examples/compression/analysis_compression.py
```diff
@@ -0,0 +1,120 @@
...
+        5.028325e-09,
+        1.997821e-08,
+        40.1389432485322889,
+    ],
+    rtol=rtol,
+    atol=atol,
+)
+
+# TODO assert initial_gamma_ref
+
```
ax3l: Done below.
```suggestion
```

ax3l (approved):
Looks great! :)

PR #419, opened 2023-08-17
Add KICKER to accepted MAD-X input

n01r:
This PR adds the KICKER element to the list of accepted MAD-X inputs.
The functionality was tested with a new Python run file that loads a MAD-X example input file and produces results that match the existing example.


examples/kicker/run_kicker_madx.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022-2023 ImpactX contributors
+# Authors: Ryan Sandberg, Axel Huebl, Chad Mitchell, Marco Garten
```
ax3l: Isn't this copied from the inputs & python equivalent of the kicker, thus?
```suggestion
# Authors: Chad Mitchell, Axel Huebl, Marco Garten
```
?

ax3l (commented):
Thank you, that's great!

Please add the test to be executed with `ctest` by adding it in `examples/CMakeLists.txt`

examples/kicker/run_kicker_madx.py
n01r: Yep, I copied the license header of `run_kicker.py` and just added my name to it.

Why the change of order? Is that because Chad is the main author of the kicker PR?

cemitch99 (approved):
This looks good to me.

examples/kicker/run_kicker_madx.py
n01r: *Outside GitHub conversation:* it is probably the above, and we just overlooked changing it for `run_kicker.py`.

ax3l (approved):

PR #423, opened 2023-08-28
pyImpactX.H: Add Include Guard

ax3l:
Add an include guard to this header file.


n01r (approved):
Thanks, looks good  :)

PR #424, opened 2023-08-28
pytest: Temporary Working Directory

ax3l:
Tests can generate temporary output and directories, which might clash and/or need cleaning between tests.
This changes the current working directory to a unique directory per test.

Seen as:
AMReX counts seconds passed (and factions thereof) since initialization. For subsequent tests, this can collide.
Also, the `UniqueStrings` in AMReX' `UtilCreateCleanDirectory` is pretty coarse in MPI contexts:
https://github.com/AMReX-Codes/amrex/blob/23.08/Src/Base/AMReX_Utility.cpp#L186-L199

If the target exist, the behavior of `std::rename` is implementation dependent. We see crashes of this on macOS.
https://en.cppreference.com/w/cpp/io/c/rename

### DIFF
--- a/tests/python/conftest.py
+++ b/tests/python/conftest.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+import os
+
 import pytest
 
 import amrex.space3d as amr
@@ -11,6 +13,9 @@ if impactx.Config.have_mpi:
     print("loaded mpi4py")
 else:
     print("NO mpi4py load")
+
+# base path for input files
+basepath = os.getcwd()
 
 
 @pytest.fixture(autouse=True, scope="function")
--- a/tests/python/conftest.py
+++ b/tests/python/conftest.py
@@ -31,5 +31,5 @@ def amrex_init(tmpdir):
                 "amrex.the_arena_is_managed=0",
             ]
         )
-    yield
-    amr.finalize()
+        yield
+        amr.finalize()
--- a/tests/python/conftest.py
+++ b/tests/python/conftest.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+import time
+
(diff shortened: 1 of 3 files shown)


ax3l:
Update: the names are odd
this is what I see on macOS CI:
```
amrex::UtilCreateCleanDirectory():  diags exists.  Renaming to:  diags.old.4000000
amrex::UtilCreateCleanDirectory():  diags exists.  Renaming to:  diags.old.0000000
amrex::UtilCreateCleanDirectory():  diags exists.  Renaming to:  diags.old.4000000
amrex::UtilCreateCleanDirectory():  diags exists.  Renaming to:  diags.old.6400000
amrex::UtilCreateCleanDirectory():  diags exists.  Renaming to:  diags.old.5300000
amrex::UtilCreateCleanDirectory():  diags exists.  Renaming to:  diags.old.1400000
amrex::UtilCreateCleanDirectory():  diags exists.  Renaming to:  diags.old.2300000
amrex::UtilCreateCleanDirectory():  diags exists.  Renaming to:  diags.old.7600000
```
this does not seem right

tests/python/conftest.py
```diff
@@ -14,21 +14,22 @@
...
+                "amrex.throw_exception=1",
+                "amrex.signal_handling=0",
+                # abort GPU runs if out-of-memory instead of swapping to host RAM
+                "amrex.abort_on_out_of_gpu_memory=1",
+                # do not rely on implicit host-device memory transfers
+                "amrex.the_arena_is_managed=0",
+            ]
+        )
     yield
     amr.finalize()
```
ax3l: According to @sayerhs in https://github.com/AMReX-Codes/pyamrex/pull/181#discussion_r1308006503
```suggestion
        yield
        amr.finalize()
```

PR #429, opened 2023-09-08
Space charge Mayes IPAC2018 benchmark

cemitch99:
This PR refers to:

C. E. Mayes, R. D. Ryne, and D. C. Sagan, "3D Space Charge in BMAD," in proc. IPAC2018,
[DOI:10.18429/JACoW-IPAC2018-THPAK085](https://doi.org/10.18429/JACoW-IPAC2018-THPAK085)

This is a replacement for PR #257 
Closes @ChristopherMayes' #256.

Added the expanding Gaussian beam test and plot script corresponding to phase space shown in Fig. 4.  

- [x] add expanding Gaussian beam test
- [x] add plot script to reproduce Fig. 4
- [x] add static Gaussian beam field calculation
- [x] add plot script to reproduce Fig. 1
- [x] add documentation
- [ ] verify resolution and convergence
- [ ] potentially use IGF solver (with much smaller grid) #605 

Figure_1

PR #437, opened 2023-09-29
MLMG: Expose Solver Parameters

ax3l:
Expose the numeric parameters of the multi-grid solver that we currently use for space charge calculations.

Used in #429


cemitch99 (approved):
Looks great.  It is good to have access to these.

PR #443, opened 2023-10-19
Diagnostics: Remove ASCII Particles

ax3l:
Remove the ASCII print option for particles. This is very costly to transfer, convert and store and was only temporary in place until we had proper binary openPMD output.

openPMD output can be converted easily to ASCII when needed, as [documented in our manual](https://impactx.readthedocs.io/en/latest/dataanalysis/dataanalysis.html).

### DIFF
--- a/src/particles/diagnostics/DiagnosticOutput.H
+++ b/src/particles/diagnostics/DiagnosticOutput.H
@@ -22,7 +22,7 @@ namespace impactx::diagnostics
     enum class OutputType
     {
         PrintNonlinearLensInvariants, ///< ASCII diagnostics for the IOTA nonlinear lens, for small tests only
-        PrintRefParticle, ///< ASCII diagnostics, for small tests only
+        PrintRefParticle, ///< ASCII diagnostics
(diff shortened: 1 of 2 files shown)


ax3l:
This is ready for review & merge and will speed up runs.

src/particles/diagnostics/DiagnosticOutput.H
```diff
@@ -21,7 +21,6 @@ namespace impactx::diagnostics
      */
     enum class OutputType
     {
-        PrintParticles, ///< ASCII diagnostics, for small tests only
         PrintNonlinearLensInvariants, ///< ASCII diagnostics for the IOTA nonlinear lens, for small tests only
         PrintRefParticle, ///< ASCII diagnostics, for small tests only
```
cemitch99: Not sure why the documentation says "for small tests only" here.  The write to ASCII for the ref particle data should be a small cost, independent of the numerical resolution.  It is valuable data to have for any test involving a moderately complex beamline.

ax3l: True, that was probably copy-pasted.
```suggestion
        PrintRefParticle, ///< ASCII diagnostics
```

PR #449, opened 2023-10-23
Python: Programmable Element w/ Dynamic

ax3l:
Enable support for dynamic attributes in the `Programmable` element. This avoids that we need to derive the element to carry around additional Python state.
https://pybind11.readthedocs.io/en/stable/classes.html#dynamic-attributes


RTSandberg (approved):
Slick!! This should be useful

PR #451, opened 2023-10-27
Remove distutils

ax3l:
Distutils is gone in Python 3.12+.

Refs.:
- https://peps.python.org/pep-0632/#migration-advice
- https://github.com/AMReX-Codes/pyamrex/pull/180

### DIFF
--- a/.github/workflows/dependencies/clang-san-openmpi.sh
+++ b/.github/workflows/dependencies/clang-san-openmpi.sh
@@ -26,7 +26,8 @@ sudo apt-get install -y \
     python3-pip         \
     wget
 
-python3 -m pip install -U pip setuptools wheel
+python3 -m pip install -U pip
+python3 -m pip install -U build packaging setuptools wheel
 python3 -m pip install -U cmake pytest
 python3 -m pip install -U -r requirements_mpi.txt
 python3 -m pip install -U -r examples/requirements.txt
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -30,189 +30,7 @@ Our workflow is described in [CONTRIBUTING.rst](CONTRIBUTING.rst).
 
 ## Developer Environment
 
-Please prepare you local development environment as follows.
```
dpgrote: This is fine here, but would be cleaner if this change was done in a separate PR since it isn't related to the main topic.

dpgrote:
This looks good and I can approve, though I have one comment posted.

Would there be a way to put all of the "python -m pip install" commands in a script so that they don't need to be written out in each of the 10 files as is now?

README.md
ax3l: Agreed, sorry that was a it sloppy and makes this harder to Backport/patch where needed indeed.

ax3l:
Yes, technically this is all in pyproject.yaml, let me see how to install that in a two liner or so.

Pip has to definitely be upgraded first and solely, because some people start with ancient pip installs that would not understand later pip steps we do.

Update: since we do not build with `--no-build-isolation`, technically all we would need is the latest `pip`... Will revisit that part another time, I have not problem updating this with `git grep` where listed. Avoids that one has to have the source cloned at that step.

README.md
ax3l: Moved this to #453

PR #458, opened 2023-11-05
Pandas: `ImpactXParticleContainer.to_df()`

ax3l:
Copy all particles into a `pandas.DataFrame`. Supports local and MPI-gathered results.

Depends on:
- [x] https://github.com/AMReX-Codes/pyamrex/pull/220


ax3l:
Looks like I need to generalize the reduce (`local=False`) in pyAMReX for non-MPI runs :)
Pushed via https://github.com/AMReX-Codes/pyamrex/pull/224

PR #470, opened 2023-11-20
[Draft] Dashboard

ax3l:
Early draft of a control and status dashboard written in trame.

- [ ] probably should depend on #469
- [ ] requires a preparatory PR to expose `step()` as an `async def` function instead of a synchronous `evolve`
- [ ] use a class for the application as in https://kitware.github.io/trame/examples/core/basics.html#class-for-trame-application


ax3l:
Replaced by #651

PR #480, opened 2023-12-05
CI: Pipx for openPMD-validator

ax3l:
Replace `pip` with `pipx` for proper entry points.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -42,9 +42,9 @@ jobs:
         python3 -m pip install --upgrade pytest
         python3 -m pip install --upgrade -r requirements_mpi.txt
         python3 -m pip install --upgrade -r examples/requirements.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/macos.yml
```diff
@@ -42,7 +42,8 @@ jobs:
         python3 -m pip install --upgrade pytest
         python3 -m pip install --upgrade -r requirements_mpi.txt
         python3 -m pip install --upgrade -r examples/requirements.txt
-        python3 -m pip install --upgrade openPMD-validator
+        python3 -m pip install --upgrade pipx
+        python3 -m pipx install --upgrade openPMD-validator
         set -e
```
ax3l:
```suggestion
        set -e
        python3 -m pip install --upgrade pipx
        python3 -m pipx install openPMD-validator
```

PR #422, opened 2023-08-28
Adding EPAC2004 space charge benchmarks

cemitch99:
Adding the space charge benchmark problems from:

R. D. Ryne et al, "A Test Suite of Space-Charge Problems for Code Benchmarking," in proceedings of EPAC2004, Lucerne, Switzerland, p. 1942 (2004).

Some of these are already implemented.  The following need to be added:

- [x] Add FODO+RF space charge benchmark.
- [x] Add Python equivalent.
- [x] Add generation of a stationary thermal/bi-thermal distribution:
- [x] Implement numerical integrator for radial CDF
- [x] Implement phase space particle sampling via CDF
- [x] Implement exchange of data
- [x] Add stationary thermal CF benchmark.
- [x] Add stationary bi-thermal CF benchmark.
- [x] Add Python equivalents + CI.
- [x] Update documentation.

Plot for bithermal test (full resolution):
bithermal

PR #492, opened 2024-01-02
Distributions: Cleaning

ax3l:
Clean type qualifiers for initial particle distributions and some formatting.

Seen in Thermal distribution in #422.


cemitch99 (approved):
These changes look good to me.

PR #493, opened 2024-01-02
Elements: `sincos(...)`

ax3l:
Update a few elements to use the `sincos` function to calculate `sin` and `cos` in one go.

Most compilers identify this as an automatic optimization, but we can also make it explicit to be sure.


cemitch99 (approved):
Looks good to me.

n01r (commented):
Looks great :)

PR #501, opened 2024-01-08
Example Updates: CFChannel, Thermal

ax3l:
CFChannel: diagnostics per slice not needed.

Thermal: Python example was not included in a test yet.
Note: both the existing inputs file and Python script for the thermal example are very slow when run in pure `Debug` mode. Let's check before merge.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -668,6 +668,13 @@ add_impactx_test(thermal
     examples/epac2004_benchmarks/analysis_thermal.py
     OFF  # no plot script yet
 )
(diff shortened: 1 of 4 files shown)


cemitch99 (approved):
This simple change looks good.

cemitch99:
The Python script for the thermal example does not run successfully in CI.  Will need to investigate.

ax3l:
Yes, both `thermal` and the `run_thermal` are running also very long in debug mode (thus in CI). We should not merge this as is yet, until we either only run these tests in `Release` mode in another CI job or make them smaller/faster.

PR #504, opened 2024-01-10
Fix: MPI-Distribution w/o SC

### LINKED ISSUE #503: MPI Domain-Decomposition w/o Space Charge
Yesterday, I realized that domain decomposition and load balancing with MPI is broken when space charge is not used. All particles end up on the same MPI rank now, causing load imbalance. Need to investigate.

Seems to show up as an issue for N>2 ranks or so, because on `ResizeMesh` we pad the beam by 3x vacuum by default for dynamic mesh resizes.

Related to #498, but likely earlier caused by #181 / #261 / #367 / #402.

ax3l:
Without space charge, we can simply split the particles up during init in equal chunks and leave them there forever.

This fixes grid generation by setting the `max_grid_size` to the AMReX `blocking_factor` (if not set by the user). Also, this avoids all calls to `ResizeMesh()` and `Redistribute` for now.

Fix #503

### DIFF
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -154,7 +154,7 @@ namespace impactx
         }
 
         amrex::ParmParse pp_algo("algo");
-        bool space_charge = true;
+        bool space_charge = false;
         pp_algo.queryAdd("space_charge", space_charge);
         amrex::Print() << " Space Charge effects: " << space_charge << "\n";
 
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -56,27 +56,6 @@ namespace impactx
         // move old diagnostics out of the way
         amrex::UtilCreateCleanDirectory("diags", true);
 
-        // n_cells has been set using temporary values earlier. We now know the true value of
-        // n_cells, so we recompute the Geometry objects for each level here *if* the user
-        // has set n_cells in the inputs file
-        {
-            amrex::Vector<int> n_cell(AMREX_SPACEDIM);
-            amrex::ParmParse const pp_amr("amr");
-            pp_amr.queryarr("n_cell", n_cell);
-
-            amrex::IntVect const lo(amrex::IntVect::TheZeroVector());
-            amrex::IntVect hi(n_cell);
-            hi -= amrex::IntVect::TheUnitVector();
-            amrex::Box index_domain(lo,hi);
-            for (int i = 0; i <= amr_data->maxLevel(); i++)
-            {
-                amr_data->Geom(i).Domain(index_domain);
-                if (i < amr_data->maxLevel()) {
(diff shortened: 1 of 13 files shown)


src/ImpactX.cpp
```diff
@@ -220,6 +226,19 @@ namespace impactx
...
+                    {
+                        // Resize the mesh, based on `m_particle_container` extent
+                        amrex::ParmParse pp_geometry("geometry");
+                        amrex::Vector<amrex::Real> prob_relative(max_level + 1, 1.0);
+                        pp_geometry.addarr("prob_relative", prob_relative);
+
+                        ResizeMesh();
+
+                        // Redistribute particles in the new mesh in x, y, t
+                        m_particle_container->Redistribute();
```
ax3l: Currently, this slightly looses particles over time.
We could and maybe should find a solution for this, so we can use this later on for load balancing, e.g., if we create secondaries or loose particles (e.g., to apertures/collimators).

src/particles/elements/diagnostics/openPMD.cpp
```diff
@@ -318,7 +318,17 @@ namespace detail
 
         // pinned memory copy
         PinnedContainer pinned_pc = pc.make_alike<amrex::PinnedArenaAllocator>();
+
+        // FIXME: Looks like this only copies particles on rank 0...
+        auto const global_np_before = pc.TotalNumberOfParticles(true, false);
+        auto const local_np_before = pc.TotalNumberOfParticles(true, true);
         pinned_pc.copyParticles(pc, true);  // no filtering
```
ax3l: @atmyers hm, that is confusing: it only copies particles on rank 0 (if we never redistribute and operate on a non-OK() container w/o space charge grid)

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -124,7 +125,19 @@ namespace impactx
...
+        {
+            const auto& pmap = ParticleDistributionMap(lid).ProcessorMap();
+            auto it = std::find(pmap.begin(), pmap.end(), amrex::ParallelDescriptor::MyProc());
+            if (it == std::end(pmap)) {
+                amrex::Abort("Attempting to add particles to box that does not exist.");
+            } else {
+                gid = *it;
+            }
+        }
+        auto& particle_tile = DefineAndReturnParticleTile(lid, gid, tid);
```
ax3l, line 139: @atmyers @WeiqunZhang for `ParIter` loops after initialization, with OpenMP parallelism, does it matter which `tid` we use?

Or will be have now always one tile and no on-rank parallelism?

ax3l: (Please note we never call `pc.Redistribute()` in this mode)

PR #507, opened 2024-01-19
Perlmutter (NERSC): January 2024 Update

ax3l:
Update modules of Perlmutter after the system upgrade.

Since the default compilers changed, please run the install of dependencies scripts again and rebuild your WarpX executables, too. Essentially, the order documented here from the top: https://impactx.readthedocs.io/en/latest/install/hpc/perlmutter.html

GPU update example:
```bash
source $HOME/perlmutter_gpu_impactx.profile  # updates in this PR!!

bash $HOME/src/impactx/Tools/machines/perlmutter-nersc/install_gpu_dependencies.sh
source ${CFS}/${proj%_g}/${USER}/sw/perlmutter/gpu/venvs/impactx/bin/activate

cd ~/src/impactx
rm -rf build
...
```

X-ref: https://github.com/ECP-WarpX/WarpX/pull/4620

### DIFF
--- a/docs/source/install/hpc/perlmutter-nersc/install_cpu_dependencies.sh
+++ b/docs/source/install/hpc/perlmutter-nersc/install_cpu_dependencies.sh
@@ -43,6 +43,9 @@ python3 -m pip uninstall -qqq -y mpi4py 2>/dev/null || true
 # General extra dependencies ##################################################
 #
 
+# tmpfs build directory: avoids issues often seen with $HOME and is faster
+build_dir=$(mktemp -d)
+
 # c-blosc (I/O compression)
 if [ -d $HOME/src/c-blosc ]
 then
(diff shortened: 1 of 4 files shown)


cemitch99 (commented):
A general comment on the installation of dependencies:  Using, eg, "install_gpu_dependencies.sh" requires storage space in $HOME, which by default is only 40 GB, and it is very easy to exceed quota.  (I keep doing it!)  It could be useful to allow an install that avoids $HOME entirely by always using $CFS/project (which allows much larger quotas).

ax3l:
Hi @cemitch99 ,

Thanks for the comment. Yes, absolutely. For that reason, I actually changed the install location for these dependency artifacts last year to point to CFS:
https://github.com/ECP-WarpX/impactx/blob/24.01/docs/source/install/hpc/perlmutter-nersc/install_gpu_dependencies.sh#L34-L36

So currently, there should be nothing but the source code be added to your home.

NERSC told me about an even better location for software dependencies than CFS on their file systems, but the default quota is so small per user that we do not use it yet (otherwise every user has to open a ticket to get a larger quota).

ax3l:
Let me move the build directories to TMPFS to avoid taking up temporary storage in HOME for those as well.

ax3l:
Done :)

Issue #391, opened 2023-06-28
Tune Calculation for Test Particle Tracking Data


cemitch99:
Calculation of particle tunes is an essential diagnostic output necessary for the modeling of rings.  In IMPACT-Z, this is usually performed using post-processing of particle tracking data in Python.  Perhaps we should support this capability intrinsically within ImpactX?

Bartolini et al, "Tune Evaluation in Simulations and Experiments," Part. Accel. 52, pp. 147-177 (1996)
http://cds.cern.ch/record/292773

Previous work related to IMPACT-Z postprocessing:
https://github.com/kilean20/pImpactR

ax3l:
@cemitch99 shall we keep this issue open for follow-up tasks after #516 was merged or open another one?

PR #516, opened 2024-02-02
Add example of tune calculation (postprocessing)

cemitch99:
An example of the calculation of tunes using postprocessing of tracking data.

The lattice is a periodic FODO channel for which the tune = phase advance per period / (2pi) is exactly known.

The horizontal tune of a single particle is computed and compared against the predicted value.  
Calculation is performed using both 4-turn formulae and using PyNAFF, https://pypi.org/project/PyNAFF/.

First attempt to address issue #391


examples/fodo_tune/README.rst
```diff
@@ -0,0 +1,59 @@
+.. _examples-fodo-tune:
+
+Tune Calculation in a Periodic FODO Channel
+============================================
```
ax3l:
```suggestion
Tune Calculation in a Periodic FODO Channel
===========================================
```

examples/fodo_tune/README.rst
```diff
@@ -0,0 +1,59 @@
...
+* :math:`\beta_\mathrm{y} = 2.82161941` m
+* :math:`\alpha_\mathrm{y} = 1.59050035`
+
+We use a 2 GeV electron beam with initial unnormalized rms emittance of 2 nm.
+
+The horizontal tune of a single particle is obtained from period-by-period tracking data using several algorithms.
+
+In this test, the computed horizontal tune must agree with the nominal value to within acceptable tolerance.
+
+This example requires installation of PyNAFF:  https://pypi.org/project/PyNAFF/
```
ax3l: I checked the requirements on PyNAFF: https://github.com/nkarast/PyNAFF/blob/master/requirements.txt

They are a subset of ours, so we could easily make it a mandatory dependency for Python builds in our `requirements.txt` :)

cemitch99: That's fantastic.  I have some ideas about a few additional testing/details to add to the script, so it isn't finalized (but it works for me with PyNAFF installed).

Also, it's still worth exploring independently some methods to calculate the tunes in-situ.

ax3l:
Tests: there seems to be one more value in `vars['TWIN']` than in `vars['ZTABS']`

examples/fodo_tune/README.rst
```diff
@@ -0,0 +1,59 @@
+.. _examples-fodo-tune:
+
+Tune Calculation in a Periodic FODO Channel
+===========================================
+
+This is identical to the FODO example, except that tracking for 100 periods is used to extract the horizontal tune.
```
ax3l:
```suggestion
This is identical to the :ref:`FODO example <examples-fodo>`, except that tracking for 100 periods is used to extract the horizontal tune.
```

examples/fodo_tune/README.rst
```diff
@@ -0,0 +1,59 @@
...
+* :math:`\beta_\mathrm{y} = 2.82161941` m
+* :math:`\alpha_\mathrm{y} = 1.59050035`
+
+We use a 2 GeV electron beam with initial unnormalized rms emittance of 2 nm.
+
+The horizontal tune of a single particle is obtained from period-by-period tracking data using several algorithms.
+
+In this test, the computed horizontal tune must agree with the nominal value to within acceptable tolerance.
+
+This example requires installation of PyNAFF:  https://pypi.org/project/PyNAFF/
```
ax3l:
```suggestion
This example requires installation of `PyNAFF <https://github.com/nkarast/PyNAFF>`__.
```

ax3l (approved):
This is awesome, thank you!

PR #522, opened 2024-02-06
CI: Add `ruff`

ax3l:
This adds `ruff`, a modern version of `pyflakes`, for code linting in pre-commits.

### DIFF
--- a/tests/python/conftest.py
+++ b/tests/python/conftest.py
@@ -8,7 +8,6 @@ import amrex.space3d as amr
 import impactx
 
 if impactx.Config.have_mpi:
-    from mpi4py import MPI
 
     print("loaded mpi4py")
 else:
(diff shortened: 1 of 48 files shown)


tests/python/conftest.py
```diff
@@ -8,7 +8,6 @@
 import impactx
 
 if impactx.Config.have_mpi:
-    from mpi4py import MPI
```
RTSandberg, line 11: is ``mpi4py`` not needed here anymore?

ax3l: good catch!

tests/python/conftest.py
```diff
@@ -8,7 +8,6 @@
 import impactx
 
 if impactx.Config.have_mpi:
```
ax3l, line 10:
```suggestion
if impactx.Config.have_mpi:
    from mpi4py import MPI  # noqa
```

PR #523, opened 2024-02-06
CI: 4 Cores Linux/Win, 3 Cores macOS

ax3l:
Increase build and test parallelism according to new increased core limits on [public GH hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories).

image

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -62,7 +62,7 @@ jobs:
       - name: Build (C++)
         if: ${{ matrix.language == 'cpp' }}
         run: |
-          $CMAKE --build build -j 2
+          $CMAKE --build build -j 4
 
       - name: Perform CodeQL Analysis
         uses: github/codeql-action/analyze@v2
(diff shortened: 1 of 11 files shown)


ax3l:
@WeiqunZhang I am surprised that OpenMPI does not find the 4 cores in CI... wondering if it actually has four in these workflows?

ax3l:
`/proc/cpuinfo` and `lscpu` are pretty clear. Seems that OpenMPI is might spot that some of those are vcores not physical cores, or is confused by the container/VM.

ax3l:
## Before
Screenshot from 2024-02-20 22-14-56
Screenshot from 2024-02-20 22-15-00

## After
Screenshot from 2024-02-20 22-48-45
Screenshot from 2024-02-20 22-48-49

ax3l:
@WeiqunZhang Looks like CI actually gets slower using more cores. I think the reason for that is that we likely only get 2 physical cores when they say "4 cores", and they mean actually 4 vcores, so 2 cores w/ 2x HT or so...

PR #527, opened 2024-02-13
Add initialization from Twiss parameters

### LINKED ISSUE #406: Beam Initialization from Courant-Snyder (Twiss) Parameters
Implement a beam generator from Twiss parameters

- [x] implementation
- [x] documentation
- [x] tests passing
- [x] new example
- [x] Python bindings

n01r:
This changes the initialization of the beam particle distribution to allow for Twiss parameters to be supplied.
The code first checks if the name for the supplied distribution has a suffix indicated by an underscore. It then checks if the base name of the distribution is an element of the list of allowed distributions that share the same input logic. Then it chooses to initialize the beam ellipse parameters either directly from the given inputs, or it calculates them from the Twiss parameters and proceeds with initialization.

Closes #406.

Tasks:
- [x] implementation
- [x] documentation
  - [x] emphasize that unnormalized emittance is the input and/or provide conversion via reference beta-gamma
- [x] dedicated test
- [x] Python bindings $\Longrightarrow$ follow-up PR #697


cemitch99 (commented):
The math here looks good to me.  We should specify in the docs that the unnormalized emittances are used as input (or add conversion by the reference beta*gamma).

n01r:
Thanks for checking, @cemitch99! Right, I added the topic to the meeting agenda today but I will also add some tasks to this PR, like docs etc. :)

src/initialization/InitDistribution.H
```diff
@@ -98,6 +98,26 @@ namespace impactx::initialization
         amrex::ParticleReal* const AMREX_RESTRICT m_part_py;
         amrex::ParticleReal* const AMREX_RESTRICT m_part_pt;
     };
+
+    /** Initialize the input parameters for all distributions that read phase space ellipse parameters from the input
+     *
```
ax3l, line 104: Please add `@param[in] name description here` doxygen strings on the two new function declarations here.

src/initialization/InitDistribution.cpp
```diff
@@ -121,6 +123,91 @@ namespace impactx
...
+            amrex::Abort("Input Error: The beta function values need to be non-zero positive values in all dimensions.");
+        }
+
+        if (emittx <= 0.0 || emitty <= 0.0 || emittt <= 0.0) {
+            amrex::Abort("Input Error: Emittance values need to be non-zero positive values in all dimensions.");
+        }
+
+        amrex::Vector<amrex::ParticleReal> alphas = {alphax, alphay, alphat};
+        amrex::Vector<amrex::ParticleReal> betas = {betax, betay, betat};
+        amrex::Vector<amrex::ParticleReal> emittances = {emittx, emitty, emittt};
```
ax3l: I think these can be `std::array<amrex::ParticleReal, 3> const`?

src/initialization/InitDistribution.cpp
```diff
@@ -22,8 +22,10 @@
 #include <AMReX_ParmParse.H>
 #include <AMReX_Print.H>
 
+#include <cmath>
 #include <string>
 #include <type_traits>
+#include <tuple>
```
ax3l: Please keep stdlib includes ordered alphabetically :)

n01r: I will remove this include actually because my current implementation does not make use of `std::tuple` anymore.

docs/source/theory/intro.rst
```diff
@@ -55,6 +55,38 @@ Unlike particles within the beam, the reference particle is described by a set o
 in a global laboratory coordinate system (x,y,z).  The momenta of the reference particle are normalized by :math:`mc`, so that :math:`p_x=\beta_x\gamma`, etc.  A parameteric plot of
 the reference trajectory variables (x,z) allows the user to view the global geometry of the accelerator structure (footprint).
 
+.. _theory-collective-beam-distribution-input:
+
+Beam Distribution Input
+-----------------------
+
+.. figure:: phase_space_ellipse.svg
```
ax3l, line 63: With 1.1kB, this svg seems small enough to check in (instead of adding to the PR description)

src/initialization/InitDistribution.H
```diff
@@ -98,6 +98,58 @@ namespace impactx::initialization
...
+     * @param sigt Reference to the variable storing the axis intersection of the phase space ellipse for time.
+     * @param sigpx Reference to the variable storing the axis intersection of the phase space ellipse for momentum in x direction.
+     * @param sigpy Reference to the variable storing the axis intersection of the phase space ellipse for momentum in y direction.
+     * @param sigpt Reference to the variable storing the axis intersection of the phase space ellipse for energy.
+     * @param muxpx Reference to the variable storing the mean for momentum in x direction.
+     * @param muypy Reference to the variable storing the mean for momentum in y direction.
+     * @param mutpt Reference to the variable storing the mean for energy.
+     */
+    void
+    setDistributionParametersFromPhaseSpaceInputs (
```
ax3l: Feel free to use `set_distribution_...` spelling for functions/methods in ImpactX. I started a bit inconsistently, but we want to use that going forward. Also makes it nicely the same in Python bindings.

src/initialization/InitDistribution.cpp
```diff
@@ -121,6 +125,91 @@ namespace impactx
...
+        pp_dist.query("alphaT", alphat);
+        pp_dist.get("betaX", betax);
+        pp_dist.get("betaY", betay);
+        pp_dist.get("betaT", betat);
+        pp_dist.get("emittX", emittx);
+        pp_dist.get("emittY", emitty);
+        pp_dist.get("emittT", emittt);
+
+        if (betax <= 0.0_prt || betay <= 0.0_prt || betat <= 0.0_prt) {
+            amrex::Abort("Input Error: The beta function values need to be non-zero positive values in all dimensions.");
```
ax3l: Please throw a runtime exception instead. This is input, so a user might want to recover in a Python notebook from it.

AMReX' abort does something more complex: https://amrex-codes.github.io/amrex/docs_html/Basics.html?highlight=amrex+abort#abort-assertion-and-backtrace

src/initialization/InitDistribution.cpp
```diff
@@ -121,6 +125,91 @@ namespace impactx
...
+        }
+
+        if (emittx <= 0.0_prt || emitty <= 0.0_prt || emittt <= 0.0_prt) {
+            amrex::Abort("Input Error: Emittance values need to be non-zero positive values in all dimensions.");
```
ax3l: See above.

src/initialization/InitDistribution.cpp
```diff
@@ -171,163 +260,144 @@ namespace impactx
...
-          pp_dist.query("muypy", muypy);
-          pp_dist.query("mutpt", mutpt);
-
-          distribution::KnownDistributions const triangle(distribution::Triangle(
-            sigx, sigy, sigt,
-            sigpx, sigpy, sigpt,
-            muxpx, muypy, mutpt));
-
-          add_particles(bunch_charge, triangle, npart);
+        std::string base_dist_type = distribution_type;
```
ax3l: The "has `_`" is not as robust as I would prefer. Can we instead change the logic to "ends in `_from_cs` || `_from_twiss`"?

ax3l (commented):
This looks great! Small comments added for improvement suggestions :)

src/initialization/InitDistribution.cpp
```diff
@@ -22,13 +22,17 @@
 #include <AMReX_ParmParse.H>
 #include <AMReX_Print.H>
 
+#include <cmath>
 #include <string>
```
ax3l, line 27: For `std::runtime_error`:
```suggestion
#include <stdexcept>
#include <string>
```

src/initialization/InitDistribution.cpp
```diff
@@ -121,6 +125,91 @@ namespace impactx
...
     }
 
+    void initialization::set_distribution_parameters_from_twiss_inputs (
+        amrex::ParmParse const & pp_dist,
+        amrex::ParticleReal& sigx, amrex::ParticleReal& sigy, amrex::ParticleReal& sigt,
+        amrex::ParticleReal& sigpx, amrex::ParticleReal& sigpy, amrex::ParticleReal& sigpt,
+        amrex::ParticleReal& muxpx, amrex::ParticleReal& muypy, amrex::ParticleReal& mutpt
+    )
+    {
+
```
ax3l, line 135:
```suggestion
    {
        using namespace amrex::literals; // for _rt and _prt

```

src/initialization/InitDistribution.cpp
```diff
@@ -22,13 +22,17 @@
...
 #include <string>
 #include <type_traits>
 #include <variant>
 
 
 namespace impactx
 {
+
+    using namespace amrex::literals; // for _rt and _prt
+
```
ax3l: Let's scope this more, not needed in default arguments, no?
```suggestion
```

src/initialization/InitDistribution.cpp
```diff
@@ -121,6 +125,91 @@ namespace impactx
...
+    }
+
+    void initialization::set_distribution_parameters_from_phase_space_inputs (
+        amrex::ParmParse const & pp_dist,
+        amrex::ParticleReal& sigx, amrex::ParticleReal& sigy, amrex::ParticleReal& sigt,
+        amrex::ParticleReal& sigpx, amrex::ParticleReal& sigpy, amrex::ParticleReal& sigpt,
+        amrex::ParticleReal& muxpx, amrex::ParticleReal& muypy, amrex::ParticleReal& mutpt
+    )
+    {
+
```
```suggestion
```

src/initialization/InitDistribution.cpp
```diff
@@ -121,6 +125,91 @@ namespace impactx
...
+        }
+
+        std::array<amrex::ParticleReal, 3> const alphas = {alphax, alphay, alphat};
+        std::array<amrex::ParticleReal, 3> const betas = {betax, betay, betat};
+        std::array<amrex::ParticleReal, 3> const emittances = {emittx, emitty, emittt};
+
+        // calculate Twiss / Courant-Snyder gammas
+        amrex::Vector<amrex::ParticleReal> gammas;
+        for (size_t i = 0; i < alphas.size(); i++)
+            gammas.push_back((1.0 + pow(alphas.at(i), 2)) / betas.at(i));
```
ax3l: You want this: https://en.cppreference.com/w/cpp/numeric/math/pow
```suggestion
            gammas.push_back((1.0 + std::pow(alphas.at(i), 2)) / betas.at(i));
```

Otherwise it is the C function for `double`:
https://en.cppreference.com/w/c/numeric/math/pow

src/initialization/InitDistribution.cpp
ax3l, line 135:
```suggestion
```

src/initialization/InitDistribution.cpp
ax3l: I do not see where it is needed, but if you do need it somewhere please keep `using namespace` inside function bodies whenever possible.

src/initialization/InitDistribution.cpp
```diff
@@ -121,6 +125,91 @@ namespace impactx
...
+        sigx = sigmas_pos.at(0);
+        sigy = sigmas_pos.at(1);
+        sigt = sigmas_pos.at(2);
+        sigpx = sigmas_mom.at(0);
+        sigpy = sigmas_mom.at(1);
+        sigpt = sigmas_mom.at(2);
+        muxpx = correlations.at(0);
+        muypy = correlations.at(1);
+        mutpt = correlations.at(2);
+
```
```suggestion
```

src/initialization/InitDistribution.cpp
```diff
@@ -121,6 +125,91 @@ namespace impactx
...
+
+        amrex::Vector<amrex::ParticleReal> sigmas_pos;
+        amrex::Vector<amrex::ParticleReal> sigmas_mom;
+        amrex::Vector<amrex::ParticleReal> correlations;
+
+        // calculate intersections of phase space ellipse with coordinate axes and the correlation factors
+        for (size_t k = 0; k < betas.size(); k++){
+            sigmas_pos.push_back(sqrt(emittances.at(k)/gammas.at(k)));
+            sigmas_mom.push_back(sqrt(emittances.at(k)/betas.at(k)));
```
```suggestion
            sigmas_pos.push_back(std::sqrt(emittances.at(k)/gammas.at(k)));
            sigmas_mom.push_back(std::sqrt(emittances.at(k)/betas.at(k)));
```

Same as for pow :)

src/initialization/InitDistribution.cpp
```diff
@@ -121,6 +125,91 @@ namespace impactx
...
+
+            correlations.push_back(alphas.at(k) / sqrt(betas.at(k) * gammas.at(k)));
```
```suggestion
            correlations.push_back(alphas.at(k) / std::sqrt(betas.at(k) * gammas.at(k)));
```

src/initialization/InitDistribution.cpp
n01r: I am using this to have the exact same types in my comparison statements.
It's probably overkill
```C++
if (betax <= 0.0_prt || betay <= 0.0_prt || betat <= 0.0_prt) {
    throw std::runtime_error("Input Error: The beta function values need to be non-zero positive values in all dimensions.");
}
```

n01r: I moved it inside the function body now :+1:

n01r:
Thanks for your review!
New changes are pushed.

ax3l (approved):

PR #531, opened 2024-02-22
Use new function from Alex for removing invalid particles.

### LINKED ISSUE #499: GPU Support for Lost Particle Collection
This part is not yet GPU capable:
https://github.com/ECP-WarpX/impactx/blob/f43277da81a1df37da876e63213917e169fc0f61/src/particles/CollectLost.cpp#L130-L163

@AlexanderSinn is working on a general, low-cost "remove invalid particles locally" function for AMReX right now, which we can reuse here.

## Work-Around for Users

For now, users simulating aperture elements on GPU need to use `amrex.the_arena_is_managed=1` as a work-around until this issue is solved.

Python equivalent:
```py
# ...
pp_amrex = amr.ParmParse("amrex")
pp_amrex.add("the_arena_is_managed", 1)

sim = ImpactX()
# ...
```

atmyers:
This uses the new function from @AlexanderSinn to remove invalid particles in CollectLost instead of the current way, which doesn't work in the GPU. 

Fixes #499.


ax3l (approved):
Thanks a lot! :sparkles:

PR #541, opened 2024-03-01
Add theory documentation for covariance matrix/quadratic form.

cemitch99:
Added an additional section to docs/source/theory/intro.rst expressing the covariance matrix in terms of beam input parameters.

PR #552, opened 2024-03-14
Rename sigma to lambda where not stddev

n01r:
This commit replaces (hopefully) all occurrences of `sigma<...>` as input to particle distributions with `lambda<...>` since the parameter describes the axis intercepts of the phase space ellipse and is only identical to the RMS extents when there is no correlation between q (position) and p (momentum).

This breaking change belongs to the same set of changes as #541.


n01r:
This also solves the name duplication where in most of our examples there is an analysis script which calculates `sig<...>` values which actually describe beam moments and standard deviations. So I made no change to these files. But I hope I got everything else and did not change too much.

RTSandberg (approved):
That is a lot of sigmas to change to lambdas!  It looks good to me

ax3l (approved):
Thank you, @n01r !

Waiting for @cemitch99 to confirm and then we synchronize the merge with #541

cemitch99 (approved):
This looks good to me.

ax3l:
@n01r can you please rebase / merge development? :)

n01r:
Rebased :)

ax3l (approved):
Thank you! :)

PR #567, opened 2024-03-28
Improve the Handling of Runtime Arguments

ax3l:
Simplify and remove earlier hacks.
Will mid term be replaced once AMReX has upstream support for named SoA components.


ax3l:
More logic added via #571

PR #573, opened 2024-04-08
Doc: Beam Distributions

ax3l:
Add a new subsection in examples for distribution generation.


RTSandberg (approved):
It looks good to me.  Are there other distributions or distribution descriptions to introduce as stand-alone examples?

cemitch99 (approved):
This looks good.  I can add examples/distgen/ as a small follow-up PR.

Issue #533, opened 2024-02-22
Reduced beam diagnostic output - kinetic energy


SchroederSa:
Hi,
I was trying to extract the kinetic energy of the particle beam using the reduced diagnostic file output.
Doing so for the apochromatic example, I get the following outputs:

 ```
red_diag = pd.read_csv('.../impactx/examples/apochromatic/diags/reduced_beam_characteristics.0.0', sep=' ', header=0)
print(red_diag['px_mean'][0]*0.511)  # Should be 0
>> -1.882738482372865e-09            # Ok!
print(red_diag['py_mean'][0]*0.511)  # Should be 0
>> 4.0816649768864856e-09            # Ok!
print(red_diag['pt_mean'][0]*0.511)  # Kinetic energy of electron beam is 10GeV
>> 1.2226565625060562e-06            <------------------ ?
print(red_diag['sig_pt'][0])         # 0.01% energy spread    
>> 0.0099944213575608                # Ok!
red_diag = pd.read_csv('.../impactx/examples/apochromatic/diags/ref_particle.0.0', sep=' ', header=0)
print(red_diag['pt'][0]*0.511)       # Kinetic energy of electron beam is 10GeV
>> -100000.7164806249                # Ok!
```

In the reduced beam characteristics file, _pt_mean_ is essentially 0. For the reference particle, however, _pt_ is 100GeV. How can I understand mean_pt other than as longitudinal momentum?

Thanks!
Sarah

p.s.: The terminology pt_**mean**, pt_**min**, pt_**max** but **sig**_pt  is a bit confusing.

ax3l:
We have the reference `beta * gamma` in the reduced diagnostics, but we could add all reference particle info there and explicitly also add kinetic energy.

ax3l:
Commented:
https://impactx.readthedocs.io/en/latest/dataanalysis/dataanalysis.html#reduced-beam-characteristics

`px_mean/min/max, py_mean/min/max, pt_mean/min/max`
- Average / minimum / maximum beam momenta normalized by reference particle momentum (unit: dimensionless, radians for transverse momenta)

Not explicit / clear enough.

cemitch99:
[...]

PR #584, opened 2024-04-18
Add Total Bunch Attributes to Monitor

ax3l:
Add our reduced beam characteristics (moments of the particle bunch) as metadata to openPMD files (the beam monitor element).

X-Ref.:
- #533

## Action Items

- [x] implement
- [x] test that `reduced_beam_characteristics` is indeed an AllReduce now.
- [x] document changes in reduced beam characteristics (no ref particle info anymore)
- [x] document openPMD attributes
- [x] test and use the new metadata from openPMD over the text file in tests (only `visualize_ml_....py`?) @ax3l @RTSandberg 

## Follow-ups

- [ ] document openPMD particle "columns" if read back to a dataframe @ax3l #618

### DIFF
--- a/examples/pytorch_surrogate_model/visualize_ml_surrogate_15_stage.py
+++ b/examples/pytorch_surrogate_model/visualize_ml_surrogate_15_stage.py
@@ -291,14 +291,15 @@ parser.add_argument(
 )
 args = parser.parse_args()
 
+impactx_surrogate_ref_particle = read_time_series("diags/ref_particle.*")
 impactx_surrogate_reduced_diags = read_time_series(
     "diags/reduced_beam_characteristics.*"
 )
-ref_gamma = np.sqrt(1 + impactx_surrogate_reduced_diags["ref_beta_gamma"] ** 2)
+ref_gamma = np.sqrt(1 + impactx_surrogate_ref_particle["ref_beta_gamma"] ** 2)
 beam_gamma = (
     ref_gamma
     - impactx_surrogate_reduced_diags["pt_mean"]
-    * impactx_surrogate_reduced_diags["ref_beta_gamma"]
+    * impactx_surrogate_ref_particle["ref_beta_gamma"]
 )
 beam_u = np.sqrt(beam_gamma**2 - 1)
 emit_x = impactx_surrogate_reduced_diags["emittance_x"]
@@ -315,13 +316,13 @@ ymarker = "^"
 ax = axT[0][0]
 scale = 1e6
 ax.plot(
-    impactx_surrogate_reduced_diags["s"][ix_slice],
+    impactx_surrogate_ref_particle["s"][ix_slice],
     emit_nx[ix_slice] * scale,
     "bo",
     label="x",
 )
 ax.plot(
-    impactx_surrogate_reduced_diags["s"][ix_slice],
+    impactx_surrogate_ref_particle["s"][ix_slice],
     emit_ny[ix_slice] * scale,
     "r",
     marker=ymarker,
@@ -335,7 +336,7 @@ ax.set_ylabel(r"emittance (mm-mrad)")
 ax = axT[0][1]
 scale = m_e * c**2 / e * 1e-9
 ax.plot(
-    impactx_surrogate_reduced_diags["s"][ix_slice],
+    impactx_surrogate_ref_particle["s"][ix_slice],
     beam_gamma[ix_slice] * scale,
     "go",
 )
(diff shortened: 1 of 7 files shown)


n01r (approved):
Looks good. That's really useful! :)

ax3l:
Will push one more update on the TODOs we marked in the RBC routines - removing reference particle info from it to avoid duplication and collisions.

ax3l:
ping @RTSandberg @n01r for review :)

cemitch99 (commented):
Just to clarify regarding the reference particle:  
The reference particle data is still written to "ref_particle_," but is no longer duplicated in "reduced_beam_characteristics_".  Instead, it is duplicated in the openPMD output.  

Is this the intent?  Did I miss anything?

examples/pytorch_surrogate_model/visualize_ml_surrogate_15_stage.py
```diff
@@ -291,14 +291,15 @@ def plot_beam_df(
 )
 args = parser.parse_args()
 
+impactx_surrogate_ref_particle = read_time_series("diags/ref_particle.*")
 impactx_surrogate_reduced_diags = read_time_series(
     "diags/reduced_beam_characteristics.*"
 )
-ref_gamma = np.sqrt(1 + impactx_surrogate_reduced_diags["ref_beta_gamma"] ** 2)
+ref_gamma = np.sqrt(1 + impactx_surrogate_ref_particle["ref_beta_gamma"] ** 2)
```
RTSandberg:
```suggestion
ref_gamma = impactx_surrogate_ref_particle["gamma"]
```

examples/pytorch_surrogate_model/visualize_ml_surrogate_15_stage.py
```diff
@@ -291,14 +291,15 @@ def plot_beam_df(
...
 beam_gamma = (
     ref_gamma
     - impactx_surrogate_reduced_diags["pt_mean"]
-    * impactx_surrogate_reduced_diags["ref_beta_gamma"]
+    * impactx_surrogate_ref_particle["ref_beta_gamma"]
```
RTSandberg:
```suggestion
    * impactx_surrogate_ref_particle["beta_gamma"]
```

RTSandberg (changes requested):
This looks good to me; I added a couple revisions to the 15 stages example and it runs for me

ax3l:
> The reference particle data is still written to "ref_particle_," but is no longer duplicated in "reduced_beam_characteristics_".

Correct. The issue was that it would otherwise collide in names once we combine it to the openPMD output.
If we want, we can still keep the information in the text file output.

> Instead, it is duplicated in the openPMD output.

It should now only be once in the openPMD output using a consistent name for reference particle and collective beam properties; stored as attributes to the beam in each step (iteration).

src/particles/elements/diagnostics/openPMD.cpp
```diff
@@ -265,8 +267,14 @@ namespace detail
...
         beam.setAttribute( "pz_ref", ref_part.pz );
         beam.setAttribute( "pt_ref", ref_part.pt );
-        beam.setAttribute( "mass", ref_part.mass );
-        beam.setAttribute( "charge", ref_part.charge );
+        beam.setAttribute( "mass_ref", ref_part.mass );
+        beam.setAttribute( "charge_ref", ref_part.charge );
+
+        // total particle bunch information
+        //   @see impactx::diagnostics::reduced_beam_characteristics
+        for(const auto &kv : m_rbc) {
```
ax3l:
```suggestion
        for (const auto &kv : m_rbc) {
```

RTSandberg (approved):
Looks good to me!

cemitch99 (commented):
I think it's important that we keep the reference particle text output.  It is a small cost for I/O, and allows a quick check by the user without any postprocessing.

n01r (approved):
Looks good to me :)

ax3l:
> I think it's important that we keep the reference particle text output. It is a small cost for I/O, and allows a quick check by the user without any postprocessing.

Good point. We have currently in `diags/` the
- `ref_particle*`
- `reduced_beam_characteristics*`

In the latter, this would remove `s` and `ref_beta_gamma` now. Shall I add `s` again (or both?).
(Later, we could also make both one file for simplicity.)

cemitch99:
I think it's good to remove `ref_beta_gamma` from `reduced_beam_characteristics`, since that information is duplicated in the `ref_particle` text output, and it is a "reference" quantity.  However, it would be good to keep the `s` output in both text files, since this is effectively the independent variable characterizing the lattice location.

ax3l:
Perfect, ok that is what I thought, too. I added the `s` column back to the text file now :)

Issue #537, opened 2024-02-27
Monitor Helper: Add every ds & Slice Elements


ax3l:
Phase space animation: we need a convenient way to add a monitor every `ds

- in an s interval between `[a;b]`, dump `N` times equally spaced in `ds`
- For general animation of the beam evolution.

Implementation ideas:
- Add a helper function to add monitors into a lattice with inputs given above.
- Details: sort and add thin monitor elements.
- Details: find if the requested position slices an element. Slice it up into multiple elements if hit.
- We should even be able to implement this in C++, not only Python.

PR #585, opened 2024-04-18
Clarify that ReducedBeamCharacteristics are relative to reference particle

n01r:
Make clear how the ReducedBeamCharacteristics quantities are in reference to the reference particle.

### DIFF
--- a/docs/source/dataanalysis/dataanalysis.rst
+++ b/docs/source/dataanalysis/dataanalysis.rst
@@ -59,6 +59,7 @@ Reduced Beam Characteristics
 ----------------------------
 
 ImpactX calculates reduced beam characteristics like averaged positions, momenta, beam emittances and Courant-Snyder (Twiss) parameters during runtime.
+Averaged positions and momenta are given as deviations with respect to the reference particle (see `Coordinates and Units <theory-coordinates-and-units>`_).
 These quantities are calculated before, after, and during each step of the simulation.
 If ``diag.slice_step_diagnostics`` is enabled, they will also be calculated during each slice of each beamline element.
 
@@ -69,13 +70,13 @@ The code writes out the values in an ASCII file prefixed ``reduced_beam_characte
 * ``s``, ``ref_beta_gamma``
     Reference particle coordinate ``s`` (unit: meter) and relativistic momentum normalized by the particle mass and the speed of light (unit: dimensionless)
 * ``x_mean/min/max``, ``y_mean/min/max``, ``t_mean/min/max``
-    Average / minimum / maximum beam particle position in the dimensions of ``x``, ``y`` (transverse coordinates, unit: meter), and ``t`` (normalized time difference :math:`ct`, unit: meter)
+    Average / minimum / maximum beam particle displacement with respect to the reference particle in the dimensions of ``x``, ``y`` (transverse coordinates, unit: meter), and ``t`` (normalized time difference :math:`ct`, unit: meter)
 * ``sig_x``, ``sig_y``, ``sig_t``
     RMS of the average beam particle positions (unit: meter)
 * ``px_mean/min/max``, ``py_mean/min/max``, ``pt_mean/min/max``
-    Average / minimum / maximum beam momenta normalized by reference particle momentum (unit: dimensionless, radians for transverse momenta)
+    Average / minimum / maximum beam momentum deviation from the reference momenta and normalized by the longitudinal reference particle momentum (unit: dimensionless, radians for transverse momenta)
 * ``sig_px``, ``sig_py``, ``sig_pt``
-    RMS of the average beam momenta (energy difference for ``pt``) (unit: dimensionless)
+    RMS of the average beam momentum deviations (energy difference for ``pt``) normalized by the longitudinal reference particle momentum (unit: dimensionless)
 * ``emittance_x``, ``emittance_y``, ``emittance_t``
     Normalized beam emittance (unit: meter)
 * ``alpha_x``, ``alpha_y``, ``alpha_t``
--- a/docs/source/dataanalysis/dataanalysis.rst
+++ b/docs/source/dataanalysis/dataanalysis.rst
@@ -59,7 +59,7 @@ Reduced Beam Characteristics
 ----------------------------
 
 ImpactX calculates reduced beam characteristics based on the beam moments during runtime.  These include averaged positions, momenta, beam emittances and Courant-Snyder (Twiss) parameters.
-Averaged positions and momenta are given as deviations with respect to the reference particle (see `Coordinates and Units <theory-coordinates-and-units>`_).
+For computing beam moments (as elsewhere), positions and momenta are given as deviations with respect to the reference particle (see :ref:`Coordinates and Units <theory-coordinates-and-units>`).
 The reduced beam characteristics are calculated before, after, and during each step of the simulation.
 If ``diag.slice_step_diagnostics`` is enabled, they will also be calculated during each slice of each beamline element.
 
--- a/docs/source/dataanalysis/dataanalysis.rst
+++ b/docs/source/dataanalysis/dataanalysis.rst
@@ -70,13 +70,13 @@ The code writes out the values in an ASCII file prefixed ``reduced_beam_characte
 * ``s``, ``ref_beta_gamma``
     Reference particle coordinate ``s`` (unit: meter) and relativistic momentum normalized by the particle mass and the speed of light (unit: dimensionless)
 * ``x_mean/min/max``, ``y_mean/min/max``, ``t_mean/min/max``
-    Average / minimum / maximum beam particle displacement with respect to the reference particle in the dimensions of ``x``, ``y`` (transverse coordinates, unit: meter), and ``t`` (normalized time difference :math:`ct`, unit: meter)
+    Average / minimum / maximum particle displacement with respect to the reference particle in the dimensions of ``x``, ``y`` (transverse coordinates, unit: meter), and ``t`` (normalized time difference :math:`ct`, unit: meter)
 * ``sig_x``, ``sig_y``, ``sig_t``
     RMS of the average beam particle positions (unit: meter)
 * ``px_mean/min/max``, ``py_mean/min/max``, ``pt_mean/min/max``
-    Average / minimum / maximum beam momentum deviation from the reference momenta and normalized by the longitudinal reference particle momentum (unit: dimensionless, radians for transverse momenta)
+    Average / minimum / maximum particle momentum deviation from the reference particle momentum, divided by the magnitude of the reference particle momentum (unit: dimensionless, radians for transverse momenta)
 * ``sig_px``, ``sig_py``, ``sig_pt``
-    RMS of the average beam momentum deviations (energy difference for ``pt``) normalized by the longitudinal reference particle momentum (unit: dimensionless)
+    Standard deviation of the particle momentum deviations (energy difference for ``pt``) normalized by the magnitude of the reference particle momentum (unit: dimensionless)
 * ``emittance_x``, ``emittance_y``, ``emittance_t``
     Normalized beam emittance (unit: meter)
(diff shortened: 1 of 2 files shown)


n01r:
This PR addresses #533.
@SchroederSa, let me know if you'd prefer it more explicit or differently phrased

docs/source/dataanalysis/dataanalysis.rst
```diff
@@ -69,13 +70,13 @@ The code writes out the values in an ASCII file prefixed ``reduced_beam_characte
 * ``s``, ``ref_beta_gamma``
     Reference particle coordinate ``s`` (unit: meter) and relativistic momentum normalized by the particle mass and the speed of light (unit: dimensionless)
 * ``x_mean/min/max``, ``y_mean/min/max``, ``t_mean/min/max``
-    Average / minimum / maximum beam particle position in the dimensions of ``x``, ``y`` (transverse coordinates, unit: meter), and ``t`` (normalized time difference :math:`ct`, unit: meter)
+    Average / minimum / maximum beam particle displacement with respect to the reference particle in the dimensions of ``x``, ``y`` (transverse coordinates, unit: meter), and ``t`` (normalized time difference :math:`ct`, unit: meter)
 * ``sig_x``, ``sig_y``, ``sig_t``
     RMS of the average beam particle positions (unit: meter)
 * ``px_mean/min/max``, ``py_mean/min/max``, ``pt_mean/min/max``
-    Average / minimum / maximum beam momenta normalized by reference particle momentum (unit: dimensionless, radians for transverse momenta)
+    Average / minimum / maximum beam momentum deviation from the reference momenta and normalized by the longitudinal reference particle momentum (unit: dimensionless, radians for transverse momenta)
```
cemitch99:
```suggestion
    Average / minimum / maximum particle momentum deviation from the reference particle momentum, divided by the magnitude of the reference particle momentum (unit: dimensionless, radians for transverse momenta)
```

docs/source/dataanalysis/dataanalysis.rst
```diff
@@ -69,13 +70,13 @@ The code writes out the values in an ASCII file prefixed ``reduced_beam_characte
...
 * ``sig_px``, ``sig_py``, ``sig_pt``
-    RMS of the average beam momenta (energy difference for ``pt``) (unit: dimensionless)
+    RMS of the average beam momentum deviations (energy difference for ``pt``) normalized by the longitudinal reference particle momentum (unit: dimensionless)
```
cemitch99:
```suggestion
    Standard deviation of the particle momentum deviations (energy difference for ``pt``) normalized by the magnitude of the reference particle momentum (unit: dimensionless)
```

docs/source/dataanalysis/dataanalysis.rst
```diff
@@ -69,13 +70,13 @@ The code writes out the values in an ASCII file prefixed ``reduced_beam_characte
 * ``s``, ``ref_beta_gamma``
     Reference particle coordinate ``s`` (unit: meter) and relativistic momentum normalized by the particle mass and the speed of light (unit: dimensionless)
 * ``x_mean/min/max``, ``y_mean/min/max``, ``t_mean/min/max``
-    Average / minimum / maximum beam particle position in the dimensions of ``x``, ``y`` (transverse coordinates, unit: meter), and ``t`` (normalized time difference :math:`ct`, unit: meter)
+    Average / minimum / maximum beam particle displacement with respect to the reference particle in the dimensions of ``x``, ``y`` (transverse coordinates, unit: meter), and ``t`` (normalized time difference :math:`ct`, unit: meter)
```
cemitch99:
```suggestion
    Average / minimum / maximum particle displacement with respect to the reference particle in the dimensions of ``x``, ``y`` (transverse coordinates, unit: meter), and ``t`` (normalized time difference :math:`ct`, unit: meter)
```

docs/source/dataanalysis/dataanalysis.rst
cemitch99: While we refer to "beam particles" vs. "reference particle", using the word "beam" here could lead to confusion.  I think the meaning is now clear without it.

n01r: Good point, thank you Chad! :slightly_smiling_face:

cemitch99 (approved):
This looks clear to me.

docs/source/dataanalysis/dataanalysis.rst
```diff
@@ -59,6 +59,7 @@ Reduced Beam Characteristics
 ----------------------------
 
 ImpactX calculates reduced beam characteristics like averaged positions, momenta, beam emittances and Courant-Snyder (Twiss) parameters during runtime.
```
SchroederSa: Would it make sense to state in this sentence that this is related to diag.enable, i.e. the '_diag._' section in the input file docs? I remember I was confused by what this section is about, where the reduced diags is specified in the program and actually the whole diagnostics setup. The third sentence where 'diag.slice_step_diagnostic' is mentioned gives a glimpse but describes a detail. (Likewise an explicit pointer to <element_name>.beam_monitor would be helpful in the section above.)

I think what contributes to the confusion is the subtle difference in terminology and a fundamental difference to the other codes (WarpX/HiPACE++). In this section it is called '**reduced** beam diagnostics' but in the inputs description it is called just 'diagnostics' while the holistic 6D diagnostic of the beam can only be done through monitors, which are not specified through the '_diag._' inputs and are not explicitly mentioned in the diagnostics section. While for the other codes with 'diag.' the holistic diagnostic is specified and the in-situ (~reduced) diagnostic appears as an additional feature within the '_diag._' specifications. I guess, this follows the idea, that the reduced diagnostics is the relevant diagnostic but then the changing terminology is confusing :).

Maybe adding a sentence in the 'Parameters: Inputs File/Diagnostics and output' section that clarifies the two options of diagnostics and access points would be helpful? Right now the headline 'Diagnostics and output' kinda infers that '_diag._' specifies THE diagnostic but as much as I understand it mainly relates to the reduced beam diagnostic and 6D phase space openPMD output is generated via the elements. The latter can only be passively grasped when digesting the side notes: 'This option is ignored for the openPMD output elements (remove them from the lattice to disable).' and 'See the beam_monitor element for backend values.'.

docs/source/dataanalysis/dataanalysis.rst
```diff
@@ -59,6 +59,7 @@ Reduced Beam Characteristics
...
+Averaged positions and momenta are given as deviations with respect to the reference particle (see `Coordinates and Units <theory-coordinates-and-units>`_).
```
SchroederSa: Maybe 'deviation' is a dangerous word here due to its mathematical definition but also plain language usage. Wouldn't just '_Positions and momenta are given with respect to the reference particle_' be less confusing  also because they are treated differently?

Also: it's not only the _averaged_ position and momenta but also minimum and maximum.

docs/source/dataanalysis/dataanalysis.rst
```diff
@@ -59,6 +59,7 @@ Reduced Beam Characteristics
...
 These quantities are calculated before, after, and during each step of the simulation.
```
SchroederSa:
```suggestion
The reduced beam characteristics are calculated before, after, and during each step of the simulation.
```

SchroederSa: 'These' lost cohesion due to the added sentence above :)

docs/source/dataanalysis/dataanalysis.rst
SchroederSa, line 1: You changed the wording from '**beam** momentum' / '**beam** position' etc to '**particle** momentum'. Is there a particular reason for it? The section title is '**beam** characteristic'. I am confused :)

SchroederSa (commented):
Thanks Marco! That makes it much clearer for sure. And the coordinates and units section is marvelous! (Maybe the Theory section in the docs should come right after Installation and well before the Usage section - note to myself :D) 
I added some thoughts.
Sarah

docs/source/dataanalysis/dataanalysis.rst
ax3l: Thanks Sarah, that is a good point.

Before adding the monitors we had some ad-hoc diagnostics (in ASCII) in ImpacX, where some of the legacy wording here comes from. We do not have a "full" diagnostics outside of monitors in ImpactX, for simplicity (still needs #537).

The "reduced diagnostics" (in situ diagnostics) is now also added to the monitors #584, so that we can technically streamline this even a bit more.

Let's add the cross-section hints you propose for now, to clarify what means what.

docs/source/dataanalysis/dataanalysis.rst
ax3l: I think deviation is correctly meant in the mathematical way (difference) here, no? @cemitch99 

There is a small linkage issue here though
```suggestion
Averaged positions and momenta are given as deviations with respect to the reference particle (see :ref:`Coordinates and Units <theory-coordinates-and-units>`).
```

cemitch99: I agree with Sarah's comment about the use of "Averaged" here.  I placed an alternative suggestions above.

cemitch99:
```suggestion
For computing beam moments (as elsewhere), positions and momenta are given as deviations with respect to the reference particle (see :ref:`Coordinates and Units <theory-coordinates-and-units>`).
```

docs/source/dataanalysis/dataanalysis.rst
cemitch99:
```suggestion
ImpactX calculates reduced beam characteristics based on the beam moments during runtime.  These include averaged positions, momenta, beam emittances and Courant-Snyder (Twiss) parameters.
```

docs/source/dataanalysis/dataanalysis.rst
cemitch99: That was me, sorry.  I did this because the word "beam" usually implies a collective property, like "beam energy" or "beam size".  At this point in the documentation, we are referring to the individual particle position and momentum.  (Of course, after the reduced diagnostics are computed, those are properties of the beam--hence the section title.)  Is that too confusing?

docs/source/dataanalysis/dataanalysis.rst
cemitch99: I made an edit to line 61, so these comments now are invisible under the Files Changed tab.  Just FYI - these are good comments and we don't want to lose them.

docs/source/dataanalysis/dataanalysis.rst
```diff
@@ -58,8 +58,9 @@ Example to print the integrated orbit path length ``s`` at each beam monitor pos
 Reduced Beam Characteristics
 ----------------------------
 
-ImpactX calculates reduced beam characteristics like averaged positions, momenta, beam emittances and Courant-Snyder (Twiss) parameters during runtime.
-These quantities are calculated before, after, and during each step of the simulation.
+ImpactX calculates reduced beam characteristics based on the beam moments during runtime.  These include averaged positions, momenta, beam emittances and Courant-Snyder (Twiss) parameters.
+For computing beam moments (as elsewhere), positions and momenta are given as deviations with respect to the reference particle (see :ref:`Coordinates and Units <theory-coordinates-and-units>`).
+The reduced beam characteristics are calculated before, after, and during each step of the simulation.
```
ax3l:
```suggestion
ImpactX calculates reduced beam characteristics based on the beam moments during runtime.
These include averaged positions, momenta, beam emittances and Courant-Snyder (Twiss) parameters.
For computing beam moments (as elsewhere), positions and momenta are given as deviations with respect to the reference particle (see :ref:`Coordinates and Units <theory-coordinates-and-units>`).

The reduced beam characteristics are stored with the output of the beam monitor element.
They are also calculated before, after, and during each step of the simulation.
```

docs/source/dataanalysis/dataanalysis.rst
SchroederSa: I think it's fine! :)

ax3l (approved):
Thank you all! :)

PR #583, opened 2024-04-16
Add README documentation for distgen

cemitch99:
Added a README containing documentation for all examples in examples/distgen.

PR #596, opened 2024-04-29
Improve treatment of RFCavity endpoints

cemitch99:
Modified the treatment of RF cavity endpoints in the quantity `efieldint` returned by the function `RF_Efield` in RFCavity.H.

This is a correction to the RF cavity field implementation that avoid spurious energy kicks associated with evaluation of the function `RF_Efield` at data values beyond the specified cavity endpoints.  The indefinite integral of Ez should be a nonzero constant outside the field range of interest, to ensure continuity across the two endpoints.

This has a small effect on the reference particle energy, and no modification of the benchmark analysis script analysis_rfcavity.py is needed.


qianglbl (approved):

examples/distgen/README.rst
```diff
@@ -0,0 +1,261 @@
+
+Generation of beam distributions
+=================================
```
ax3l, line 3:
```suggestion
Generation of Beam Distributions
================================
```

examples/distgen/README.rst
```diff
@@ -0,0 +1,261 @@
+
```
ax3l: Should this file be part of the current PR? I think it only belongs in #583

src/particles/elements/RFCavity.H
```diff
@@ -377,6 +377,22 @@ namespace RFCavityData
...
+               }
+            }
+            else
+            {
+               efieldint = zmid*0.5_prt*cos_data[0];;
+               for (int j=1; j < m_ncoef; ++j)
+               {
+                 efieldint = efieldint - zlen*sin_data[j]*cos(j*pi)/(j*2*pi);
+               }
+            }
```
ax3l: Adding an inline comment and avoiding another code branch:
```suggestion
            else  // endpoint of the RF, outsize zlen
            {
               efieldint = std::copysign(z, z)*zmid*0.5_prt*cos_data[0];;
               for (int j=1; j < m_ncoef; ++j)
               {
                 efieldint = efieldint - zlen*sin_data[j]*cos(j*pi)/(j*2*pi);
               }
            }
```
https://en.cppreference.com/w/cpp/numeric/math/copysign

examples/distgen/README.rst
cemitch99: No, agreed.

ax3l (approved):
:+1:

PR #601, opened 2024-05-01
Doc: Split Theory Into Sections

ax3l:
Split the theory chapter into sections for readability.


ax3l:
Merging - as discussed & approved in developer meeting last week :)

PR #603, opened 2024-05-06
Doc: TaperedPL Details

ax3l:
Add more details to the documentation.

Follow-up to #562

### DIFF
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -918,7 +918,13 @@ This module provides elements for the accelerator lattice.
 
 .. py:class:: impactx.elements.TaperedPL(k, taper, units, dx=0, dy=0, rotation=0)
 
-   A thin nonlinear plasma lens with transverse taper.
+   A thin nonlinear plasma lens with transverse (horizontal) taper
+
+   .. math::
+
+      B_x = g \left( y + \frac{xy}{D_x} \right), \quad \quad B_y = -g \left(x + \frac{x^2 + y^2}{2 D_x} \right)
+
+   where :math:`g` is the (linear) field gradient in T/m and :math:`D_x` is the targeted horizontal dispersion in m.
 
    :param k:  integrated focusing strength in m^(-1) (if units = 0)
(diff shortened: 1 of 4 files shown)


docs/source/usage/python.rst
```diff
@@ -918,7 +918,13 @@ This module provides elements for the accelerator lattice.
...
 .. py:class:: impactx.elements.TaperedPL(k, taper, units, dx=0, dy=0, rotation=0)
 
-   A thin nonlinear plasma lens with transverse taper.
+   A thin nonlinear plasma lens with transverse (horizontal) taper
+
+   .. math::
+
+      B_x = g \left( y + \frac{xy}{D_x} \right), \quad \quad B_y = -g \left(x + \frac{x^2 + y^2}{2 D_x} \right)
+
+   where :math:`g` is the (linear) field gradient in T/m and :math:`D_x` is the targeted horizontal dispersion in m.
```
RemiLehe, line 927: It would be nice to connect the above parameters (`g`, `D_x`), with the ones that are actually used by the user (`k`, `taper`)

cemitch99: Done.

cemitch99 (commented):
This looks good.  I made more explicit the connection between the field parameters and the user parameters by adding symbols "g" and "Dx" in the user parameter definitions (Remi's comment).

I think we need to make the same changes in "parameters.rst".

ax3l:
Thanks!

Ooops, oh no I overwrite you k description
Can you push the same changeagain? I am so sorry, I was playing with an issue in the cross-reference and did not look carefully before doing a git force push.

docs/source/usage/python.rst
ax3l: Ooops, oh no I overwrote your k description
Can you push the same change again? I am so sorry, I was playing with an issue in the cross-reference and did not look carefully before doing a git force push.

ax3l: Added again :)

cemitch99 (approved):
This looks good to me.

Issue #610, opened 2024-05-14
PyTest Tests: Ensure Individually Runable


ax3l:
Ensure sensible tests inside `tests/python/` are also runnable from the command line with and without MPI support.

For instance, the logic inside `tests/python/test_charge_deposition.py` needs an update for the logic of external file loading (inputs) from AMReX. Could be simplified to avoid this or restructured.

Seen by @cemitch99 and @xbojanic

PR #611, opened 2024-05-14
Fix Sole Run of `test_charge_deposition.py`

ax3l:
Ensure this test can still be run like a script, not only with PyTest. Execute:
```console
python tests/python/test_charge_deposition.py
```

Related to #610


cemitch99 (approved):
Tried it, and this worked for me also.

PR #612, opened 2024-05-14
Python: Deposit Charge

ax3l:
Add a Python function `sim.deposit_charge()` that deposits the current particle beam into x,y,z `rho`, independent of space charge logic and not applying any kicks to the beam.

Useful for debugging and experimenting with new algorithms.


cemitch99 (commented):
Thanks for this.

PR #616, opened 2024-05-17
Fix Py Extensions & Install `.pyi` Files

ax3l:
- The python runtime extension files did shadow the actual pybind classes we declare for `ImpactXParIter` and a few others, because the module name was the same.
- `.pyi` stub files were not yet installed.
- add a test


RTSandberg (commented):
This fixes my issue with the ``ImpactXParIter`` name collision and so turning into a module rather than the class I was expecting

PR #619, opened 2024-05-21
set num threads to avoid hanging

RTSandberg:
As documented in PyAMReX [issue 322](https://github.com/AMReX-Codes/pyamrex/issues/322), the multithreaded PyTorch defaults interfere with MPI-enabled AMReX and these sometimes result in programs that hang.  This PR updates the 15-LPA-stage surrogate model example to avoid this hanging.

### DIFF
--- a/examples/pytorch_surrogate_model/run_ml_surrogate_15_stage.py
+++ b/examples/pytorch_surrogate_model/run_ml_surrogate_15_stage.py
@@ -94,10 +94,18 @@ def download_and_unzip(url, data_dir):
 data_url = "https://zenodo.org/records/10810754/files/models.zip?download=1"
 download_and_unzip(data_url, "models.zip")
 
+# it was found that the PyTorch multithreaded defaults interfere with MPI-enabled AMReX
+# when initializing the models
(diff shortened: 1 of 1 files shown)


examples/pytorch_surrogate_model/run_ml_surrogate_15_stage.py
```diff
@@ -94,10 +94,18 @@ def download_and_unzip(url, data_dir):
 data_url = "https://zenodo.org/records/10810754/files/models.zip?download=1"
 download_and_unzip(data_url, "models.zip")
 
+# it was found that the PyTorch multithreaded defaults interfere with MPI-enabled AMReX
+# when initializing the models
+# so we manually set the number of threads
```
ax3l:
```suggestion
# It was found that the PyTorch multithreaded defaults interfere with MPI-enabled AMReX
# when initializing the models: https://github.com/AMReX-Codes/pyamrex/issues/322
# So we manually set the number of threads to serial (1).
```

ax3l (approved):
Thank you! :tada:

PR #630, opened 2024-06-04
Correct the sign of `p_t` in the documentation

RemiLehe:
After discussions with @cemitch99, it seems that the sign of `p_t` in the documentation was incorrect.


cemitch99 (approved):
Looks good, thanks.

cemitch99 (commented):
Looks good.  Remi and I discussed already.

PR #631, opened 2024-06-05
Escape code in BeamMonitor when openPMD is off

n01r:
This fixes an issue where the code could not be compiled with the following combination of build options.
```
cmake -S . -B build -DImpactX_MPI=OFF -DImpactX_PYTHON=ON -DImpactX_COMPUTE=NOACC -DImpactX_OPENPMD=OFF
```

This was first encountered by @proy30. Thanks for finding this!


proy30 (approved):

ax3l (approved):
Thank you! :)

PR #635, opened 2024-06-13
Doc: Corrected typo in the section of RFCavity

tanweihou:
Corrected the typo, it should be `sin_coefficients` instead of `cos_coefficients`.


ax3l (approved):
Thanks a lot, good catch! :)

cemitch99 (commented):
Great.  Thanks for catching the typo.

PR #642, opened 2024-06-26
update python usage doc, field names and default values

moellep:
Also removed doc for ChrUniformAcc which was replaced with ChrAcc.

### DIFF
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -417,7 +417,7 @@ This module provides particle beam distributions that can be used to initialize 
 
    A 6D Waterbag distribution.
 
-.. py:class:: impactx.distribution.Thermal(k, kT, kT_halo, normalize, normalize_halo, halo)
+.. py:class:: impactx.distribution.Thermal(k, kT, kT_halo, normalize, normalize_halo, halo=0.0)
 
    A 6D stationary thermal or bithermal distribution.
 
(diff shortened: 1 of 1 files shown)


ax3l:
Hi @moellep, thank you for your first PR to ImpactX! :tada: 

Please excuse the slow reaction - I was on vacation the last two weeks and we'll take a look now :)

docs/source/usage/python.rst
```diff
@@ -807,7 +807,7 @@ This module provides elements for the accelerator lattice.
    :param dy: vertical translation error in m
    :param rotation: rotation error in the transverse plane [degrees]
 
-.. py:class:: impactx.elements.ShortRF(V, freq, phase, dx=0, dy=0, rotation=0)
+.. py:class:: impactx.elements.ShortRF(V, freq, phase=-90.0, dx=0, dy=0, rotation=0)
```
ax3l, line 810: We have no default parameter for the phase.
```suggestion
.. py:class:: impactx.elements.ShortRF(V, freq, phase, dx=0, dy=0, rotation=0)
```

ax3l: Oh, we have in the inputs file but not in Python. Can you please add the default of `-90.0` in `src/python/elements.cpp`?

ax3l: I was wrong, it's added there already.

ax3l (approved):
Thank you so much for this!

PR #643, opened 2024-07-16
Release 24.07

ax3l:
Prepare the July release of ImpactX :tada:


cemitch99 (approved):
Looks good.

ax3l:
The failing Windows test appeared without a change in ImpactX in CI recently. We need to track it down separately if this is a real issue or a CI issue.

PR #645, opened 2024-07-17
Update Quad to treat special case k = 0

cemitch99:
Updated the quadrupole element to handle the case of zero focusing strength.


examples/distgen/README.rst
```diff
@@ -71,9 +71,9 @@ We run the following script to analyze correctness:
 A Kapchinskij-Vladimirskij (K-V) distribution
```
ax3l: The changes in this file should be reset.

src/particles/elements/Quad.H
```diff
@@ -128,6 +128,7 @@ namespace impactx
 
                tout = t + (slice_ds/betgam2)*pt;
                // ptout = pt;
+            } else {
             }
```
ax3l, line 133:
```suggestion
                // nothing to do for zero focusing strength
            }
```

ax3l (approved):
Thank you! :)

ax3l (approved):

PR #646, opened 2024-07-18
Python: `ShortRF` Properties

ax3l:
Add element properties for `ShortRF`, so it can be modified after variable creation, too.

### DIFF
--- a/src/python/elements.cpp
+++ b/src/python/elements.cpp
@@ -969,6 +969,21 @@ void init_elements(py::module& m)
              py::arg("rotation") = 0,
              "A short RF cavity element."
         )
+        .def_property("V",
+            [](ShortRF & short_rf) { return short_rf.m_V; },
+            [](ShortRF & short_rf, amrex::ParticleReal V) { short_rf.m_V = V; },
+            "Normalized RF voltage V = maximum energy gain/(m*c^2)"
+        )
+        .def_property("freq",
+            [](ShortRF & short_rf) { return short_rf.m_freq; },
+            [](ShortRF & short_rf, int freq) { short_rf.m_freq = freq; },
+            "RF frequency in Hz"
+        )
+        .def_property("phase",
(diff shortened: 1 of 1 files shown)


ax3l:
@EZoni volunteered to complete the other leftover elements in a separate PR :pray:

cemitch99 (approved):
Looks great, thanks.

EZoni (approved):

cemitch99 (approved):
Looks great.

src/python/elements.cpp
```diff
@@ -969,6 +969,21 @@ void init_elements(py::module& m)
...
+            "Normalized RF voltage V = maximum energy gain/(m*c^2)"
+        )
+        .def_property("freq",
+            [](ShortRF & short_rf) { return short_rf.m_freq; },
+            [](ShortRF & short_rf, int freq) { short_rf.m_freq = freq; },
+            "RF frequency in Hz"
+        )
+        .def_property("phase",
+            [](ShortRF & short_rf) { return short_rf.m_phase; },
+            [](ShortRF & short_rf, int phase) { short_rf.m_phase = phase; },
```
EZoni, line 984: Actually I just noticed this, should `freq` and `phase` be `amrex::ParticleReal` too, as opposed to `int`? If so, I can fix it in the follow-up PR.

cemitch99: Thanks for catching this!  Those should definitely be amrex::ParticleReal.

ax3l: Ooops, yes please. Copy pasted this, dang!

EZoni: @ax3l 
I will fix it in the upcoming PR. Question, which I guess is pybind11-related: why does the compiler not complain about/catch this discrepancy between types?

ax3l: I am surprised as well, should issue a narrowing warning.
https://godbolt.org/z/f6bE3xPPE

Thx for the fix in #647

PR #647, opened 2024-07-18
Python: add elements properties

EZoni:
Add missing elements properties (set, get), see #646 for reference.

I found that there are two naming conventions for units arguments and/or member variables, namely `unit` and `units`. Is that intentional or shall we clean it up (here or in a separate PR)?

### DIFF
--- a/src/python/elements.cpp
+++ b/src/python/elements.cpp
@@ -509,6 +509,26 @@ void init_elements(py::module& m)
              py::arg("rotation") = 0,
              "Edge focusing associated with bend entry or exit."
         )
+        .def_property("psi",
+            [](DipEdge & dip_edge) { return dip_edge.m_psi; },
+            [](DipEdge & dip_edge, amrex::ParticleReal psi) { dip_edge.m_psi = psi; },
+            "Pole face angle in rad"
+        )
+        .def_property("rc",
+            [](DipEdge & dip_edge) { return dip_edge.m_rc; },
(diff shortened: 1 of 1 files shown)


ax3l:
> I found that there are two naming conventions for units arguments and/or member variables, namely unit and units. Is that intentional or shall we clean it up (here or in a separate PR)?

I would propose to generally prefer singular unless it appears very awkward. @cemitch99 shall we unify to that?

cemitch99:
Yes, agreed.  The singular is will work fine.

ax3l (approved):
Thank you! :)

src/python/elements.cpp
```diff
@@ -976,12 +1094,12 @@ void init_elements(py::module& m)
...
         .def_property("freq",
             [](ShortRF & short_rf) { return short_rf.m_freq; },
-            [](ShortRF & short_rf, int freq) { short_rf.m_freq = freq; },
+            [](ShortRF & short_rf, amrex::ParticleReal freq) { short_rf.m_freq = freq; },
             "RF frequency in Hz"
         )
         .def_property("phase",
             [](ShortRF & short_rf) { return short_rf.m_phase; },
-            [](ShortRF & short_rf, int phase) { short_rf.m_phase = phase; },
+            [](ShortRF & short_rf, amrex::ParticleReal phase) { short_rf.m_phase = phase; },
```
ax3l, line 1102: Fix for #646

PR #655, opened 2024-08-01
CMake: Simplify Python `add_impactx_test`

ax3l:
We do not need to ask the developer but can find out if a test is a Python test, based on the input file ending.


EZoni (approved):
Looks good to me, thanks!

cemitch99 (approved):
Looks great to me.  This should help streamline the addition of new tests.

PR #627, opened 2024-05-31
1D CSR Model

xbojanic:
This PR implements a 1D CSR model.  Currently, this is the ultrarelativistic steady-state wakefield model (eg, eq 19 of 
E. L. Saldin et al, NIMA 398, p. 373-394 (1997), [DOI:10.1016/S0168-9002(97)00822-X](https://doi.org/10.1016/S0168-9002(97)00822-X)).

This includes:

- [x] wake function calculation
  - [x] integrate out singularity value
- [x] beam profile convolution
- [x] apply the force "kick" to the particle momentum.
- [x] benchmark example
- [x] GPU support
- [x] MPI support
- [x] user-facing documentation & controls: inputs & Python

Close #452

PR #683, opened 2024-08-16
Update ABLASTR & pyAMReX

ax3l:
To latest commit on `development` each.

This adds https://github.com/ECP-WarpX/WarpX/pull/5146, which we need for #627


cemitch99 (approved):
Looks good to me.

PR #696, opened 2024-09-05
Add Python CSR wake plotting test.

cemitch99:
This adds a Python plotting script for the CSR wakefield in tests/python, as a follow-up to [PR#627](url).  The original author is `xbojanic`.


cemitch99 (commented):
Locally, this gives me:

NO mpi4py load
Segmentation fault: 11

even if I comment lines 141-144.

ax3l:
Can you try if the standalone run of the test via `pytest` works locally?
https://github.com/AMReX-Codes/pyamrex?tab=readme-ov-file#test

ax3l (approved):

cemitch99:
Yes, this test passes if I run it locally via pytest, as expected.  The saved png files may be deleted on completion--is there a way to run via pytest and also view the generated png files?

ax3l:
Good question. I have to google that for pytest.

PR #704, opened 2024-09-17
`load_inputs_file`: Simplify File Check

ax3l:
AMReX will now check this in a user-friendly and performant manner.

- [x] wait for next update that pulls in https://github.com/AMReX-Codes/amrex/pull/4156


n01r (commented):
Gotcha, looks good

PR #711, opened 2024-09-23
Fix main() Profile Region

AlexanderSinn:
Address new errors with AMReX development of the kind:
```
189: 0::Assertion `m_profiler.m_do_profiling == false' failed, file "Src/Base/AMReX_Arena.cpp", line 123 !!!
```
and
```
5: terminate called after throwing an instance of 'std::runtime_error'
5:   what():  Assertion `static_cast<int>(ttstack.size()) == global_depth' failed, file "Src/Base/AMReX_TinyProfiler.cpp", line 174, Msg: "TinyProfiler sections must be nested with respect to each other"
```

### DIFF
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -54,6 +54,9 @@ namespace impactx {
             // this one last
             amr_data.reset();
 
+            if (amrex::Initialized())
+                amrex::Finalize();
+
             // only finalize once
             m_grids_initialized = false;
         }
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -54,9 +54,6 @@ namespace impactx {
             // this one last
             amr_data.reset();
 
-            if (amrex::Initialized())
-                amrex::Finalize();
-
             // only finalize once
             m_grids_initialized = false;
         }
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -29,16 +29,15 @@ int main(int argc, char* argv[])
     impactx::initialization::default_init_AMReX(argc, argv);
 
     {
-        BL_PROFILE("main()");
+        BL_PROFILE_VAR("main()", pmain);
         impactx::ImpactX impactX;
         impactX.init_grids();
         impactX.initBeamDistributionFromInputs();
         impactX.initLatticeElementsFromInputs();
         impactX.evolve();
+        BL_PROFILE_VAR_STOP(pmain);
         impactX.finalize();
     }
-
-    amrex::Finalize();
 
 #if defined(AMREX_USE_MPI)
     AMREX_ALWAYS_ASSERT(MPI_SUCCESS == MPI_Finalize());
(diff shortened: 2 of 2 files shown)


src/main.cpp
```diff
@@ -28,16 +28,17 @@ int main(int argc, char* argv[])
     // it here so users can pass command line arguments
     impactx::initialization::default_init_AMReX(argc, argv);
 
-    BL_PROFILE_VAR("main()", pmain);
     {
+        BL_PROFILE("main()");
```
WeiqunZhang: I think this is the right fix.

src/main.cpp
```diff
@@ -28,16 +28,17 @@ int main(int argc, char* argv[])
...
         impactx::ImpactX impactX;
         impactX.init_grids();
         impactX.initBeamDistributionFromInputs();
         impactX.initLatticeElementsFromInputs();
         impactX.evolve();
         impactX.finalize();
     }
-    BL_PROFILE_VAR_STOP(pmain);
+
+    amrex::Finalize();
```
ax3l: The `main()` is only used in executable mode, not in Python mode.

src/ImpactX.cpp
```diff
@@ -54,9 +54,6 @@ namespace impactx {
             // this one last
             amr_data.reset();
 
-            if (amrex::Initialized())
```
ax3l: We need this for Python module usage.

src/main.cpp
```diff
@@ -28,16 +28,17 @@ int main(int argc, char* argv[])
...
 
-    BL_PROFILE_VAR("main()", pmain);
     {
+        BL_PROFILE("main()");
         impactx::ImpactX impactX;
         impactX.init_grids();
         impactX.initBeamDistributionFromInputs();
         impactX.initLatticeElementsFromInputs();
         impactX.evolve();
         impactX.finalize();
```
ax3l, line 39: We could end the profiling before this line is called.

src/main.cpp
AlexanderSinn: Ah thanks

src/main.cpp
AlexanderSinn: Yes

src/main.cpp
```diff
@@ -28,16 +28,16 @@ int main(int argc, char* argv[])
...
 
-    BL_PROFILE_VAR("main()", pmain);
     {
+        BL_PROFILE_VAR("main()", pmain);
         impactx::ImpactX impactX;
         impactX.init_grids();
         impactX.initBeamDistributionFromInputs();
         impactX.initLatticeElementsFromInputs();
         impactX.evolve();
+        BL_PROFILE_VAR_STOP(pmain);
```
ax3l, line 38: Or this:
```suggestion
        impactx::ImpactX impactX;
        {
            BL_PROFILE("main()");
            impactX.init_grids();
            impactX.initBeamDistributionFromInputs();
            impactX.initLatticeElementsFromInputs();
            impactX.evolve();
        }
```

ax3l: nah, not better

ax3l (approved):
Thank you, Alex! :sparkles:

Issue #662, opened 2024-08-08
Allow user output at specific turn numbers


cemitch99:
To avoid large output to h5 files and long processing times, it would be good to allow the user to specify at a monitor specific turn indices, or a range of turn indices, at which output is produced.

ax3l:
I'll move the WarpX "Intervals Parser" to ABLASTR so we can reuse this in ImpactX.
https://warpx.readthedocs.io/en/latest/usage/parameters.html#intervals-parser

We can target this for 24.10 since the WarpX/ABLASTR release for 24.09 is alread out.

ax3l:
First implementation in #713 for our `24.10` release.

PR #713, opened 2024-09-24
BeamMonitor: Enable Filtering by Cycle (Turn)

ax3l:
For modeling of periodic lattices, we can now sub-select the output of particles at the `BeamMonitor`/`beam_monitor` element to every N-th period (turn or cycle).

This is first, simple (equidistant) implementation of #662.

## To Do

- [x] implement
- [x] document
- [x] add example
- [x] finalize user-facing naming of filter

### DIFF
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -452,7 +452,7 @@ Lattice Elements
             * ``<element_name>.cycle_intervals`` (``int``, default value: ``1``)
 
                 For periodic lattice, only output every N cycles (turns).
-                By default, diagnostics is performed every cycle.
+                By default, diagnostics are returned every cycle.
 
             * ``<element_name>.nonlinear_lens_invariants`` (``boolean``, default value: ``false``)
 
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -449,9 +449,9 @@ Lattice Elements
                 openPMD `iteration encoding <https://openpmd-api.readthedocs.io/en/0.14.0/usage/concepts.html#iteration-and-series>`__: (v)ariable based, (f)ile based, (g)roup based (default)
                 variable based is an `experimental feature with ADIOS2 <https://openpmd-api.readthedocs.io/en/0.14.0/backends/adios2.html#experimental-new-adios2-schema>`__.
 
-            * ``<element_name>.cycle_intervals`` (``int``, default value: ``1``)
+            * ``<element_name>.period_sample_intervals`` (``int``, default value: ``1``)
 
-                For periodic lattice, only output every N cycles (turns).
+                for periodic lattice, only output every Nth period (turn).
                 By default, diagnostics are returned every cycle.
 
             * ``<element_name>.nonlinear_lens_invariants`` (``boolean``, default value: ``false``)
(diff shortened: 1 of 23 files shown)


docs/source/usage/parameters.rst
```diff
@@ -449,6 +449,11 @@ Lattice Elements
                 openPMD `iteration encoding <https://openpmd-api.readthedocs.io/en/0.14.0/usage/concepts.html#iteration-and-series>`__: (v)ariable based, (f)ile based, (g)roup based (default)
                 variable based is an `experimental feature with ADIOS2 <https://openpmd-api.readthedocs.io/en/0.14.0/backends/adios2.html#experimental-new-adios2-schema>`__.
 
+            * ``<element_name>.cycle_intervals`` (``int``, default value: ``1``)
+
+                For periodic lattice, only output every N cycles (turns).
```
cemitch99: Elsewhere, we use the term "periods" instead of "cycles".  For example, we set lattice.periods = N.  I think this terminology is more standard, and it might be good to use this for consistency.

docs/source/usage/parameters.rst
```diff
@@ -449,6 +449,11 @@ Lattice Elements
...
+                By default, diagnostics is performed every cycle.
```
```suggestion
                By default, diagnostics are returned every cycle.
```

docs/source/usage/python.rst
```diff
@@ -623,7 +623,7 @@ This module provides elements for the accelerator lattice.
    :param dy: vertical translation error in m
    :param rotation: rotation error in the transverse plane [degrees]
 
-.. py:class:: impactx.elements.BeamMonitor(name, backend="default", encoding="g")
+.. py:class:: impactx.elements.BeamMonitor(name, backend="default", encoding="g", cycle_intervals=1)
```
cemitch99: It might be good to brainstorm on the name `cycle_intervals`.

examples/fodo_channel/README.rst
```diff
@@ -0,0 +1,74 @@
...
+.. dropdown:: Script ``plot_fodo.py``
+
+   .. literalinclude:: plot_fodo.py
+      :language: python3
+      :caption: You can copy this file from ``examples/fodo/plot_fodo.py``.
+
+.. figure:: https://gist.githubusercontent.com/ax3l/8ae7dcb9e07c361e002fa56d6b16cb16/raw/cc952670bb946cd7a62282bc7aa3f03f3d5faa16/fodo_channel.png
+   :alt: preserved emittance in the FODO channel.
+
+   FODO transversal emittance evolution (preserved)
```
cemitch99:
```suggestion
   FODO transverse emittance evolution (preserved)
```

docs/source/usage/parameters.rst
ax3l: Good point, currently based on
https://github.com/ECP-WarpX/impactx/blob/3a8ab8dc981aafccc9ab3de167e222d986f8b374/src/ImpactX.cpp#L183-L187

cemitch99: For the user-facing name, what about "period_sample_interval"?

ax3l: That sounds good. I would write as in warpx plural, `period_sample_intervals`, because in a follow-up we will enable users to write multiple intervals here: https://warpx.readthedocs.io/en/latest/usage/parameters.html#intervals-parser

docs/source/usage/python.rst
ax3l: Renamed to `period_sample_intervals`

ax3l:
@cemitch99 this is ready for review & merge now :)

cemitch99 (approved):
Looks great.  This is a useful feature to have, thanks.

src/particles/elements/diagnostics/openPMD.cpp
```diff
@@ -179,8 +179,10 @@ namespace detail
         else if ( "f" == encoding )
             series_encoding = openPMD::IterationEncoding::fileBased;
 
-        // legacy options from other diagnostics
         amrex::ParmParse pp_diag("diag");
+        // turn filter
+        pp_diag.queryAdd("period_sample_intervals", m_period_sample_intervals);
```
ax3l, line 184: Bug, removed in #1592

PR #718, opened 2024-09-27
[Dashboard] Improved Simulation Output Logging

proy30:
PR improves existing simulation logging by displaying simulation diagnostics in the Trame xterm.

**Note**: The output is currently streamed to the Trame xterm only **after** the simulation has completed. The ideal behavior, where the contents are streamed progressively as the simulation runs, is still a work in progress.

[Wurlitzer ](https://pypi.org/project/wurlitzer/) library utilized for this functionality.

Before HGMoRa0dnb
After chrome_odyapCQFWS


ax3l:
Thanks! I think we can alternatively solve this in pyAMReX modifying AMReX' `amrex::system::osout` and `amrex::system::oserr`. That would then also be useful for general pyAMReX/WarpX/ImpactX usage in Jupyter.

- https://pybind11.readthedocs.io/en/stable/advanced/pycpp/utilities.html#capturing-standard-output-from-ostream

ax3l (approved):
I took a look at the package and its implementation and it is ok to use this as a work-around for now :)

PR #725, opened 2024-10-03
Relocated simulation file in dashboard

proy30:
simulation file is now stored in `simulation.py` rather than `.Analyze.plot_PhaseSpaceProjections.phaseSpace`

**before:**
dashboard/
└── Analyze/
    └── plot_PhaseSpaceProjections/
        └── phaseSpace.py

**after:**
dashboard/
└── simulation.py


ax3l (approved):
Thank you, nice!

src/python/impactx/dashboard/Analyze/plotsMain.py
```diff
@@ -14,10 +14,10 @@
 from trame.widgets import matplotlib, plotly, vuetify
 from wurlitzer import pipes
 
+from ..simulation import run_simulation
```
ax3l, line 17: It is still a bit unstructured that `run_simulation` is called from an analysis sub-folder.

PR #730, opened 2024-10-09
Update zero-strength quad case = drift.

cemitch99:
A thick quadrupole with zero focusing strength requires special treatment in the limit k=0, and the element should be equivalent to a drift of the same length.

- [x] update Quad.H
- [x] update ChrQuad.H


ax3l (approved):
Thank you!

From a performance perspective, keeping element code small (register usage) would be another possibility by changing under the hood user-specified quads that have no field strengths to drift elements in lattices.

PR #738, opened 2024-10-11
`PYBIND11_FINDPYTHON=ON`

ax3l:
Reuse our `find_package(Python ...)` call and use new CMake logic in pybind11.
https://pybind11.readthedocs.io/en/stable/compiling.html#modules-with-cmake
https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-version-selection

X-ref: https://github.com/openPMD/openPMD-api/pull/1677#issuecomment-2407767112

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,7 +136,7 @@ impactx_make_third_party_includes_system(WarpX::ablastr_3d ablastr_3d)
 
 # Python
 if(ImpactX_PYTHON)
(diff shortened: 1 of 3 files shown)


EZoni (approved):

EZoni (approved):

EZoni:
@ax3l I think you're tracking whether this works or not on your end, so I let you merge or auto-merge.

PR #714, opened 2024-09-24
[Draft] Transport and Covariance Matrices

ax3l:
- [x] general structure
  - [x] use `amrex::SmallMatrix`
- [x] distribution init of covariance matrix
- [ ] return transport matrices in each element
- [x] user-defined, thin linear element (kick) #538 #703
- [x] push for user-defined linear element
- [ ] push for covariance matrix
  - [x] `ImpactX::evolve` -> `ImpactX::track_particles` #741
  - [ ] add: `ImpactX::track_envelope`

PR #741, opened 2024-10-11
Introduce `sim.track_particles()`

ax3l:
We will have multiple ways to evolve the sim soon (particles, only ref particle, only linear push of covariance matrix #714), so it is time to rename our central loop.

For now, `sim.evolve()` just calls the `sim.track_particles()` logic.


cemitch99 (approved):
Got it.

PR #750, opened 2024-10-23
CI: fix `AppleClang` workflow

EZoni:
We should move `CXXFLAGS: "-Werror -Wno-error=pass-failed"` to when ImpactX builds. Currently it is picked up by `pip`. It didn't fail before probably because there was cached version of Python stuff. Now there is probably a new version of something that requires rebuilding some packages.


cemitch99 (approved):
Look great, thanks for the fix.

PR #767, opened 2024-12-06
Draft of Python bindings for SmallMatrix objects

cemitch99:
This is a draft of possible bindings to support type conversions from Python `np.array` object to `amrex::SmallMatrix` object.  To support PR #743.


ax3l:
Implemented via https://github.com/AMReX-Codes/pyamrex/pull/399 :)

PR #771, opened 2024-12-12
Release: 24.12

cemitch99:
Prepare the December release of ImpactX.

Delay updating the AMReX and pyAMReX dependencies for now, pending:  
https://github.com/ECP-WarpX/impactx/pull/760
https://github.com/ECP-WarpX/impactx/issues/769


proy30 (approved):

EZoni (approved):

ax3l:
Hi, just FYI: CF release fails
https://github.com/conda-forge/impactx-feedstock/pull/41
because dependencies where not updated in this PR for pyAMReX & AMReX. See #733 for a template (or other earlier releases). The files that also needed a bump were:
- cmake/dependencies/ABLASTR.cmake
- cmake/dependencies/pyAMReX.cmake

ax3l:
We still need a `new_version.py` script or so to automate this step, so we can have a simple way to release (copied from WarpX so far in the docs but not yet implemented): https://impactx.readthedocs.io/en/latest/maintenance/release.html

cemitch99:
Hm.  We _intentionally_ had to skip updating the pyAMReX & AMReX dependencies for this release, because of some issues with Windows + GPU memory management on the AMReX side.  However, all the ImpactX tests pass in CI using the existing (previous) dependencies.

I didn't realize updating these dependencies was mandatory.  (Maybe these feedstock tests should be run before the release is published?)  Is there some way to keep the old dependencies on the `conda-forge` side?

ax3l:
Yes, we mostly keep the versions in sync to avoid subtle binary (ABI) incompatibilities that we have no CI and developer time for.

> some issues with Windows + GPU memory management on the AMReX side

Can you link those just for reference? Curious.

Note that we do not yet build packages for GPU on CF releases.

>  Is there some way to keep the old dependencies on the conda-forge side?

Yes, one can temporarily patch [ImpactX](https://github.com/ECP-WarpX/impactx/blob/24.12/cmake/dependencies/pyAMReX.cmake#L62) or [pyAMReX](https://github.com/AMReX-Codes/pyamrex/blob/b7ce2882c82a8c10e6831d47c1ea5fc4232591da/CMakeLists.txt#L236-L239) on CF, i.e., [this line](https://github.com/AMReX-Codes/pyamrex/blob/b7ce2882c82a8c10e6831d47c1ea5fc4232591da/CMakeLists.txt#L236-L239) is intentionally set to [COMPATIBILITY SameMinorVersion](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#generating-a-package-version-file) and currently triggers the CF build release error in ImpactX 24.12 for CF:
```
...
CMake Error at cmake/dependencies/pyAMReX.cmake:62 (find_package):
  Could not find a configuration file for package "pyAMReX" that is
  compatible with requested version "24.10".

  The following configuration files were considered but not accepted:

    /home/conda/feedstock_root/build_artifacts/impactx_1734118488323/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/cmake/pyAMReX/pyAMReXConfig.cmake, version: 24.12

Call Stack (most recent call first):
  cmake/dependencies/pyAMReX.cmake:81 (find_pyamrex)
  CMakeLists.txt:150 (include)


-- Configuring incomplete, errors occurred!
```

I recommend just patching the ImpactX line in https://github.com/ECP-WarpX/impactx/blob/24.12/cmake/dependencies/pyAMReX.cmake#L62

ax3l:
>  (Maybe these feedstock tests should be run before the release is published?)

That's a bit of a chicken-egg problem. We could do pre-releases to run though package managers (Spack, CF, etc.) and we do have general tests that build with the same compilers, but the exact release has to happen at some point.

All in all no big deal, we just need the release scripts I mentioned above to automate for us so we have an easier time and don´t need to remember such details :)

PR #772, opened 2024-12-17
Add defaults.py, Custom Step Values, and fix code in dashboard

proy30:
This PR accomplishes the following:

- **Relocates Hard-Coded Default Values to `defaults.py`**
  - Centralizes default values in a single file for easier access and modification.
- **Introduces Custom Step Count Values for Inputs**
  - Previously, all parameters used a default step count of 1, which was often suboptimal for scientific inputs. 
  - Now, it is possible for one parameter (ex: beta) to increment by 0.1 and another parameter (mlmg_relative_tolerance) to increment by 1e-12.

Additionally:
- **Adds missing __init __ files for CSR and Space Charge folders**
- **Fix bug which did not allow users to be able to run a simulation**
   - "name" parameter in a lattice element always held an error_message.
- **Fix matplotlib errors**
    - Two errors would appear (1) when simulation is ran (2) when phase space plot is selected to view:
          - 1.) Error: Skip state value for 'simulation_data' since its content is not serializable
          - 2.) JS Error => ReferenceError: trame__matplotlib_2 is not defined

### DIFF
--- a/src/python/impactx/dashboard/Analyze/plotsMain.py
+++ b/src/python/impactx/dashboard/Analyze/plotsMain.py
@@ -101,6 +101,7 @@ state.active_plot = None
 state.filtered_data = []
 state.all_data = []
 state.all_headers = []
+state.phase_space_png = None
 
 # -----------------------------------------------------------------------------
 # Functions to update table/plot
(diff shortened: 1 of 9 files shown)


cemitch99 (approved):
Looks good to me.

ax3l (approved):
Thank you, LGTM :+1:

PR #778, opened 2025-01-04
Add UI functionality to reset inputs

proy30:
Adds reset buttons to clear inputs - both section-specific (distribution, lattice, etc.) and a global reset for all dashboard inputs.

In addition, defaults.py is updated.


ax3l (approved):
Nice, very good!

PR #780, opened 2025-01-06
Condense repetitive dashboard code

proy30:
Currently, the dashboard contains many repeated logic, especially for UI components. This PR condenses some of the logic so that future input implementations can be easily placed.

Functionality for the dashboard remains the same.


src/python/impactx/dashboard/Input/csrConfiguration/csrMain.py
```diff
@@ -1,5 +1,6 @@
 from trame.widgets import vuetify
 
+from ...Input.trameFunctions import TrameFunctions
```
proy30: The class holding key input vuetify components (text fields, select boxes, routes, dialog tabs) is named TrameFunctions. I think a class rename is needed, maybe DashboardComponents?

proy30:
A larger, but necessary dashboard PR. There's still more logic to refine and optimize, but this is a solid starting point.

ax3l (approved):
LGTM, thank you!

PR #786, opened 2025-01-07
Python: Mixin Elements Submodule

ax3l:
Move the mixin elements (non-physical, mostly helper methods and attributes) to `impactx.elements.mixin` to avoid that they are listed by default and, e.g., picked up in the GUI as physical elements.

### DIFF
--- a/src/python/elements.cpp
+++ b/src/python/elements.cpp
@@ -135,14 +135,19 @@ namespace
 
 void init_elements(py::module& m)
 {
-    py::module_ const me = m.def_submodule(
+    py::module_ me = m.def_submodule(
         "elements",
         "Accelerator lattice elements in ImpactX"
     );
 
     // mixin classes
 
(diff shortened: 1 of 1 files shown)


src/python/elements.cpp
```diff
@@ -135,14 +135,19 @@ namespace
...
 
     // mixin classes
 
-    py::class_<elements::Named>(me, "Named")
+    py::module_ const mx = me.def_submodule(
+        "mixin",
+        "Mixin classes for accelerator lattice elements in ImpactX"
+    );
+
+    py::class_<elements::Named>(mx, "Named")
```
ax3l, line 150: @proy30 just FYI: the mixin classes (`Named`, `Thick`, `Thin`, `Alignment`) should not be selectable as "physical elements" in the accelerator lattice by the user in the GUI. This change should do that automatically now.

proy30 (approved):

PR #787, opened 2025-01-08
Python 3.13 Support, 3.8 EOL

ax3l:
Add support for Python 3.13.
Remove 3.8 because it is EOL as of Oct 2024.

Bump to pybind11 2.13.0+, which add Python 3.13 support in CI.

Same as:
- https://github.com/ECP-WarpX/WarpX/pull/5361
- https://github.com/AMReX-Codes/pyamrex/pull/374

Note that we remove Python 3.8 only from documentation for now, but do not hard enforce that it is not used (because old Ubuntu versions and CI still use it). Enforcement would go in `CMakeLists.txt` (`find_package(Python 3.9 ...)`) and in `setup.py` (`python_requires=">=3.8"`).

### DIFF
--- a/cmake/dependencies/pybind11.cmake
+++ b/cmake/dependencies/pybind11.cmake
@@ -37,7 +37,7 @@ function(find_pybind11)
             mark_as_advanced(FETCHCONTENT_UPDATES_DISCONNECTED_FETCHEDpybind11)
         endif()
     else()
-        find_package(pybind11 2.12.0 CONFIG REQUIRED)
+        find_package(pybind11 2.13.0 CONFIG REQUIRED)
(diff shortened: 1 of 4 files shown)


cmake/dependencies/pybind11.cmake
EZoni, line 1: Is there a reason why we have `2.13.0` in one line and `v2.13.6` in another line or is it just a typo?

ax3l: The one line is the minimum requirement for `find_package` (any newer than `2.13.0` will work), the other is the exact version we download on the fly (latest patch level, because great).

cemitch99 (approved):
This looks good to me.

PR #789, opened 2025-01-08
Address possible NaN in ExactSbend & ChrUniformAcc

### LINKED ISSUE #599: Exact sbend returns NaN for certain beam/element input values
The nonlinear map for the exact sbend element requires extracting a square root, which can yield NaN for certain initial values providing a negative radicand.  An error flag or switch should be added to avoid this.  Observed by @SchroederSa

cemitch99:
This PR addresses the special case when the square root appearing in the nonlinear map for the ExactSbend element becomes negative -- due, for example, to numerical roundoff.

A similar restriction of the domain of map definition appears for the ChrAcc element, and this is also addressed here.  A modified cyclotron example is added with increased energy spread to illustrate the possibility of particle loss.

Fix #599 .

- [x] treat the special case of negative sqrt
- [x] mark particles as lost if outside the domain of map definition
- [x] find a problematic example to verify that Issue #599 is resolved
- [x] double-check the nonlinear map implementation
- [x] fix possible NaN in ChrAcc element also
- [x] add modified cyclotron example as benchmark test
- [x] choose values for analysis script

Follow-up:  Check for any additional (nonlinear) elements that require further restrictions on the domain of definition to avoid nan's.


cemitch99:
Found two examples that illustrate the original problem and this fix.  Just trying to decide what works best for testing.

examples/cyclotron/README.rst
```diff
@@ -50,3 +50,52 @@ Visualize
 .. note::
 
    TODO :)
+
+
+.. _examples-cyclotron-loss:
+
+Cyclotron with Dynamical Losses
+================================
```
ax3l:
```suggestion
Cyclotron with Dynamical Losses
===============================
```

ax3l (approved):
That looks great, thank you for handling these corner cases!

examples/cyclotron/README.rst
```diff
@@ -50,3 +50,52 @@ Visualize
...
+
+Cyclotron with Dynamical Losses
+===============================
+
+This is identical to the example above, except that the initial energy spread of the beam is increased.  The value is so large that some particles are not successfully transported.
+These particles are considered "lost".
+
+In this test, the initial values of :math:`\sigma_x`, :math:`\sigma_y`, :math:`\sigma_t`, :math:`\epsilon_x`, :math:`\epsilon_y`, and :math:`\epsilon_t` must agree with nominal values.
+
+In addition, the fraction of charge that is lost must agree with the nominal value of 7.7754%, to within a specified tolerance.
```
SchroederSa, line 65: It's a bit unclear where the number 7.7754% is coming from.

examples/cyclotron/run_cyclotron_loss.py
```diff
@@ -0,0 +1,69 @@
...
+    muxpx=-0.0,
+    muypy=0.0,
+    mutpt=0.0,
+)
+sim.add_particles(bunch_charge_C, distr, npart)
+
+# add beam diagnostics
+monitor = elements.BeamMonitor("monitor", backend="h5")
+
+# design the accelerator lattice)
```
SchroederSa, line 49: Little typo: ')' at the end of the line

src/particles/elements/ChrUniformAcc.H
```diff
@@ -123,57 +123,66 @@ namespace impactx
...
-
-            // assign updated momenta
-            px = pxout;
-            py = pyout;
-            pt = ptout;
+            amrex::ParticleReal const pti_tot2 = std::pow(pti_tot,2);
+            amrex::ParticleReal const ptf_tot2 = std::pow(ptf_tot,2);
+
+            // check whether particle lies within the domain of map definition
+            if (pti_tot2 <= 1_prt || ptf_tot2 <= 1_prt) {
```
SchroederSa, line 130: Just a question: Are there coding practices on what to put in an 'if' clause and what in a 'else' clause? I would have naively thought the if(XXX) is testing the good / default case and the 'else:' handling the exception.
Thanks!

examples/cyclotron/run_cyclotron_loss.py
```diff
@@ -0,0 +1,69 @@
...
+
+# domain decomposition & space charge mesh
+sim.init_grids()
+
+# load initial beam
+kin_energy_MeV = 4.0e-3  # reference energy
+bunch_charge_C = 1.0e-9  # used with space charge
+npart = 10000  # number of macro particles
+
+#   reference particle
```
SchroederSa, line 28: There are an irregular number of spaces in line 28 and 32 after the '#  '

SchroederSa (commented):
Thanks for taking care of this! I just added some meticulous hints to typos.

PR #790, opened 2025-01-09
Release 25.01

ax3l:
Prepare the January release of ImpactX :tada:


ax3l:
AMReX Windows FFTW: will be addressed in the weekly update https://github.com/AMReX-Codes/amrex/pull/4282
Compiles on CF & Spack also with the current `25.01` release of AMReX, so we can ignore it for this release.

ax3l:
@WeiqunZhang interesting, when we update AMReX to `25.01`, our ASAN plugin shows a memory leak in `amrex::FFT::OpenBCSolver`

[job-logs.zip](https://github.com/user-attachments/files/18364698/job-logs.zip)

cmake/dependencies/ABLASTR.cmake
```diff
@@ -161,15 +161,15 @@ set(ImpactX_openpmd_src ""
...
+set(ImpactX_ablastr_branch "25.01"
     CACHE STRING
     "Repository branch for ImpactX_ablastr_repo if(ImpactX_ablastr_internal)")
 
 # AMReX is transitively pulled through ABLASTR
 set(ImpactX_amrex_repo "https://github.com/AMReX-Codes/amrex.git"
     CACHE STRING
     "Repository URI to pull and build AMReX from if(ImpactX_amrex_internal)")
-set(ImpactX_amrex_branch "e64ffef57a7608d1d60f9abe738cc634e9c1272e"
+set(ImpactX_amrex_branch ""
```
cemitch99, line 173: To clarify for my understanding:  without a specification here, what version of AMReX will be used?

ax3l: Without specifying it here, we pick the version set in ABLASTR (WarpX):
https://github.com/ECP-WarpX/WarpX/blob/development/cmake/dependencies/AMReX.cmake#L293-L301

I usually only overwrite it here if we temporarily need a newer version than what is set there.

WeiqunZhang:
https://www.fftw.org/faq/section3.html#leaks

> Question 3.16. FFTW seems to have a memory leak.
> After you create a plan, FFTW caches the information required to quickly recreate the plan. (See [Q3.9 `Can I save FFTW's > plans?'](https://www.fftw.org/faq/section3.html#savePlans)) It also maintains a small amount of other persistent memory. You can deallocate all of FFTW's internally allocated memory, if you wish, by calling fftw_cleanup(), as documented in the manual.

We could try to call fftw_cleanup.

ax3l:
Interesting, I never saw this issue in other FFTW runs. Sure, let's add this to AMReX finalization?

WeiqunZhang:
Could we try https://github.com/AMReX-Codes/amrex/pull/4289 to see if it fixes the problem? If it does, I am actually not sure we should fix it.

ax3l:
I'll try.

I was wondering why `linac-segment.run`  is the only test that shows the memory leak, and the reason might be that all other FFT tests in ImpactX are set to not use MPI.

WeiqunZhang:
The issue might be batched fftw plans. When there is no MPI, we use fftw to do the 3d fft. But with MPI, we fftw to do batched 1d ffts.

WeiqunZhang:
In that test, we call both `fftw_plan_dft_?2?` and `fftwf_plan_many_dft_?2?`. But the log shows it only complains about the latter.

ax3l:
Correct, same with valgrind.

WeiqunZhang:
Okay, I found the issue. This is a real memory leak. Let me think about how to fix it.

WeiqunZhang:
https://github.com/AMReX-Codes/amrex/pull/4290

ax3l:
Thank you - awesome, that fixes the leaks we see here :tada:

ax3l:
Testing all new AMReX bugfixes in #791, but will merge this release PR first.

PR #793, opened 2025-01-09
Remove Managed Memory Workarounds

ax3l:
Two changes:
- the AMReX default for `the_arena_is_managed` is false for a while, so we do not need to "overwrite" it anymore https://github.com/AMReX-Codes/amrex/pull/3438
- our initial implementation of particle loss in apertures was reworked for a while and does not require managed GPU memory anymore: remove work-arounds in inputs #531


cemitch99 (approved):
Ok, great.

atmyers (approved):

PR #795, opened 2025-01-10
Fix Lost Particle w/ Runtime Attr

### LINKED ISSUE #762: Monitor+Nonlinear invariants+Aperture yields SegFault
The test script below yields a SegFault.  However, the script runs successfully in any of the following cases:

1) If the flag "nonlinear_lens_invariants=True" is changed to "nonlinear_lens_invariants=False".
2) If the Aperture element is removed or commented.
3) If the first Monitor element is removed or commented.

The following is a Python script.  (The change .py -> .txt was made to allow attachment to this Issue.)
[run_aperture_test_segfault.txt](https://github.com/user-attachments/files/17894412/run_aperture_test_segfault.txt)

ax3l:
For some processes and/or diagnostics, we add extra runtime attributes to our beam (particle container). The logic that collected "lost" particles in the beamline, i.e., as marked as lost in apertures, did not account for extra runtime attributes and thus got lost in bookkeeping.

This fixes the collection logic to be more robust and also copy any extra runtime attributes over to the "lost" particle recording.

This was first seen with an input that used the nonlinear lens (NLL), NLL-invariant diagnostics and an aperture at the same time. Fix #762

- [ ] add a test to the NLL / IOTA examples (i.e., make sure we have an aperture that removes particles active?)

### DIFF
--- a/src/particles/CollectLost.cpp
+++ b/src/particles/CollectLost.cpp
@@ -55,8 +55,26 @@ namespace impactx
         BL_PROFILE("impactX::collect_lost_particles");
 
         using SrcData = ImpactXParticleContainer::ParticleTileType::ConstParticleTileDataType;
-
         ImpactXParticleContainer& dest = *source.GetLostParticleContainer();
+
+        // Check destination has the same attributes as source + "s_lost"
+        for (auto & name : source.GetRealSoANames())
+        {
+            if (!dest.HasRealComp(name)) {
+                dest.AddRealComp(name);
+            }
+        }
+        for (auto & name : source.GetIntSoANames())
+        {
+            if (!dest.HasIntComp(name)) {
+                dest.AddIntComp(name);
+            }
+        }
+        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(source.GetRealSoANames().size() + 1 == dest.GetRealSoANames().size(),
+                                         "Source and destination have different Real attributes!");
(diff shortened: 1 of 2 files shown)


src/particles/CollectLost.cpp
```diff
@@ -55,8 +55,26 @@ namespace impactx
...
         using SrcData = ImpactXParticleContainer::ParticleTileType::ConstParticleTileDataType;
-
         ImpactXParticleContainer& dest = *source.GetLostParticleContainer();
+
+        // Check destination has the same attributes as source + "s_lost"
+        for (auto & name : source.GetRealSoANames())
+        {
+            if (!dest.HasRealComp(name)) {
+                dest.AddRealComp(name);
+            }
```
ax3l: @atmyers question for you: we use `amrex::filterAndTransformParticles` below.
I am a bit afraid that when the order of attributes added in `dest` is not 100% the same as in `source` that this gets confused.

In particular, the attributes in `source` are something like:
- x,y,t,px,py,pt,I,H
and in `dest`
- x,y,t,px,py,pt,s,I,H

`s`, `I` and `H` are runtime components, all are `Real`.

Will that lead to broken/mismatched or partial copies from `source` to `dest`?

ax3l: I tried to add a few asserts, but maybe we need to generalize with new helper methods/functions for copies that help with the matching a bit, to simplify complex bookkeeping.

PR #804, opened 2025-01-18
Fix Aperture Backwards Compat.

### LINKED ISSUE #803: Using bool variable with queryAdd for elements appearing more than once can yield false -> true
For a lattice with an element (named, _eg_, "`ap1`") that appears more than once in the lattice, the use of lines such as:

`bool has_old_xmax = pp_element.queryAdd("xmax", aperture_x);`
`amrex::Print() << "has_old_xmax" << " " << has_old_xmax << std::endl;`

in `InitElement.cpp` shows that the value of `has_old_xmax` can change from `false` to `true`, even if "xmax" does not appear anywhere.

This appears in the context of PR #802.  It has no affect on Python-based input.

ax3l:
Fix #803:
Do not add the old parameter to `ParmParse`. Just check if it is there and only add `aperture_x/y`.

Follow-up to a bug introduced in #788

### DIFF
--- a/src/initialization/InitElement.cpp
+++ b/src/initialization/InitElement.cpp
@@ -433,8 +433,8 @@ namespace detail
             // pp_element.getWithParser("aperture_x", aperture_x);
             // pp_element.getWithParser("aperture_y", aperture_y);
             // Backwards compatibility to ImpactX <= 25.01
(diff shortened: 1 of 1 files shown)


cemitch99 (approved):
This works.

PR #806, opened 2025-01-18
Refactor Unit Handling for kin_energy

proy30:
This PR refactors the implementation for handling kinetic energy unit changes. While initially intended to address broken functionality introduced in [#780](https://github.com/ECP-WarpX/impactx/pull/780), the scope has evolved to implement a more intuitive approach for unit handling that no longer depends on the previous logic.

Previously, changing units would update both the UI display and backend values to represent the same physical quantity (e.g., 5 MeV changed to GeV would result in 0.005 on both the UI and backend). This behavior has now been redesigned to better align with user expectations. With this update, the backend value (in MeV) is updated to reflect the new unit, while the UI number remains constant (e.g., 5 MeV → 5 GeV).

### DIFF
--- a/src/python/impactx/dashboard/Input/defaults.py
+++ b/src/python/impactx/dashboard/Input/defaults.py
@@ -16,7 +16,7 @@ class DashboardDefaults:
         "charge_qe": -1,
         "mass_MeV": 0.51099895,
         "npart": 1000,
-        "kin_energy": 2e3,
+        "kin_energy_on_ui": 2e3,
         "kin_energy_MeV": 2e3,
         "kin_energy_unit": "MeV",
         "bunch_charge_C": 1e-9,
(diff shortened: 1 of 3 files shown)


proy30:
before:

https://github.com/user-attachments/assets/3f3d9c38-95c5-4a27-b5c0-e61d0871fe5e

after (print statement only used in testing):

https://github.com/user-attachments/assets/159db5f1-b718-4c21-8392-afa11c9274f2

ax3l:
@cemitch99 Oh, did we want to change this? I liked that it flipped so nicely :)

ax3l:
Ok. The rationale is to keep it simple and consistent, that is ok :)

src/python/impactx/dashboard/Input/defaults.py
```diff
@@ -16,7 +16,7 @@ class DashboardDefaults:
         "charge_qe": -1,
         "mass_MeV": 0.51099895,
         "npart": 1000,
-        "kin_energy": 2e3,
+        "kin_energy_on_ui": 2e3,
```
ax3l, line 19: Do you want to add the suffix `_eV` to the variable name for clarity?

ax3l (approved):

PR #809, opened 2025-01-23
Fix reset input capabilities

proy30:
Referencing wrong variable names


cemitch99 (approved):

EZoni (approved):

EZoni:
The "Source / style" failures may be fixed by https://github.com/AMReX-Codes/amrex/pull/4308?

ax3l:
We hoped so, but does not seem to fix it just yet.

PR #812, opened 2025-01-25
Profiler: Charge Deposition

ax3l:
Add missing annotations for the tiny profiler sections in space charge kicks: charge deposition was not captured.


ax3l:
Hm, maybe I did not add it because `ablastr::particles::deposit_charge::ChargeDeposition` should have covered it anyway...

ax3l:
@cemitch99 , we could look at your 1B particle example again to see what we might have missed. I cannot spot directly another section.

cemitch99 (approved):

PR #814, opened 2025-01-27
Track Covariance Map

ax3l:
Add a new runtime mode, tracking of 2nd moments in a covariance matrix.

- [x] generalize initialization #815
- [x] add map initialization to each element
- [x] add maps for core elements (drift, quad, bend, shortrf)
- [x] push loop
- [x] diagnostics
- [x] example(s): input file
- [x] example(s): Python
- [x] example(s): analysis / NaN pandas issue
- [x] documentation

Follow-up:  add the implementation of space charge kicks (in envelope mode), following Issue #826


src/ImpactX.cpp
```diff
@@ -336,4 +336,35 @@ namespace impactx {
...
+        {
+            std::visit([&ref, &cm](auto&& element){
+                // push reference particle in global coordinates
+                {
+                    BL_PROFILE("impactx::Push::RefPart");
+                    element(ref);
+                }
+
+                // push Covariance Matrix
+                cm = cm * element.transport_map(ref);
```
ax3l: Correct would be: `R_T * cm * R`

cemitch99: Transpose is on the right:  `R * cm * R_T`

ax3l: Thanks! Dang, we had a 50% chance to get it right by guessing :D

ax3l: fixed

src/particles/elements/mixin/lineartransport.H
```diff
@@ -24,27 +24,35 @@ namespace impactx::elements::mixin
...
+        {
+            static_assert(
+                std::is_base_of_v<LinearTransport, T_Element>,
+                "LinearTransport can only be used as a mixin class!"
+            );
+
+            // small trick to force every derived class has to implement a method transport_map
+            // (w/o using a purely virtual function)
+            T_Element& element = *static_cast<T_Element*>(this);
+            cm *= element.transport_map(cm);
```
cemitch99: Is this logic right?  The function `transport_map` in each element takes the reference particle data as input, but this appears to call `transport_map` with the covariance matrix `cm` as input.  Also, if the intent is to push the covariance matrix by the matrix `R`, then the math should look like `cm = R * cm * R^T`, which appears also `ImpactX.cpp`.

cemitch99:
Regarding user control of the run mode, I propose `algo.envelope_mode = true/false`, with false as the default (particle tracking).

ax3l:
We could also make it more general: `algo.method = `
- `envelope`
- `reference_particle`
- `tracking` (default)

That way we have one flag for all planned modes and we can continue to add modes beyond 3 in the future :D (e.g., ensemble/batch variants of these modes or so)

src/particles/elements/mixin/lineartransport.H
ax3l: Oh, that is outdated. I only fixed it in `ImpactX.cpp`.

cemitch99:
What about `algo.mode = track_envelope, track_ref_part, track_particles`, or similar?  This could reflect the name of the function used in the push loop.  I think "method" is maybe not descriptive enough, and could apply to any algorithm.

ax3l:
Perfect

ax3l:
Oh, or instead of mode we name it `algo.track = envelope, ref_part, particles`.

cemitch99:
That works, too.  FYI:  I used `algo.mode = track_envelopes` in the diagnostics PR, but it's a trivial change.

ax3l:
I will fix the Doxygen warning/error later, please ignore for now.

ax3l:
@cemitch99 I started a new, general user introduction:
https://impactx--814.org.readthedocs.build/en/814/usage/how_to_run.html

ax3l:
I also documented the new APIs for the new run modes. Now we only need to add an example.

src/particles/diagnostics/ReducedBeamCharacteristics.cpp
```diff
@@ -403,7 +406,183 @@ namespace impactx::diagnostics
...
+           std::tuple<amrex::ParticleReal, amrex::ParticleReal, amrex::ParticleReal> emittances = Eigenemittances(Sigma);
+           emittance_1 = std::get<0>(emittances);
+           emittance_2 = std::get<1>(emittances);
+           emittance_3 = std::get<2>(emittances);
+        }
+
+        auto const nan = std::numeric_limits<amrex::ParticleReal>::quiet_NaN();
+
+        std::unordered_map<std::string, amrex::ParticleReal> data;
+        data["x_mean"] = nan;
```
ax3l, line 549: We can check if most of those are truly undefined or instead are just "trivial" values (like `0`). I would plot the trivial ones as well, because it makes the columns available more consistent and one can reuse plot scripts for envelope and tracking sims that way.

src/particles/diagnostics/DiagnosticOutput.cpp
```diff
@@ -9,226 +9,221 @@
...
-        } // if( otype == OutputType::PrintReducedBeamCharacteristics)
+            file_handler << "\n";
+        }
+    }
+
+    void
+    prepare_header (
+        amrex::AllPrintToFile & file_handler,
+        OutputType otype,
+        bool has_charge,
```
ax3l: Bug: remove `has_charge` and write all columns unconditionally.

cemitch99:
Note:  The NaNs appearing in `reduced_beam_characteristics.*` appear to cause a problem when using the strategy from PR #822 to write the CI analysis scripts.  Need to find a workaround or replace the NaNs.

docs/source/usage/how_to_run.rst
```diff
@@ -3,87 +3,89 @@
 Run ImpactX
 ===========
 
-In order to run a new simulation:
 
-#. create a **new directory**, where the simulation will be run
-#. make sure the ImpactX **executable** is either copied into this directory or in your ``PATH`` `environment variable <https://en.wikipedia.org/wiki/PATH_(variable)>`__
-#. add an **inputs file** and on :ref:`HPC systems <install-hpc>` a **submission script** to the directory
-#. run
+When to run: Physics Models
```
cemitch99: It's great to have this summary here, but I don't think the "when", "what", "how", headings are very helpful.  For example, when I read "When to run", my first response is:  "When I need to model something."  I suggest more direct, uniformly-phrased questions like:  "How to select a run mode", "How to select a user interface", "How to select a computing system", and "How to run the code".  I'll add some possible suggestions.

cemitch99:
```suggestion
How to select a tracking mode
```

docs/source/usage/how_to_run.rst
```diff
@@ -3,87 +3,89 @@
...
+---------------------------
 
-1. Run Directory
-----------------
+ImpactX most powerful and central physics model is its symplectic particle tracker with collective effects (space charge, CSR, wakefields, etc.).
```
cemitch99:
```suggestion
ImpactX can be run using any of three distinct tracking modes.  ImpactX's most powerful tracking mode makes use of symplectic particle tracking with collective effects included (space charge, CSR, wakefields, etc.).
```

cemitch99: I would not call these "physics models".  They really are modes that differ due to what is being tracked--for example, for linear beamline elements (default) the physical model is the same in all three cases.

docs/source/usage/how_to_run.rst
```diff
@@ -3,87 +3,89 @@
...
+Additionally, ImpactX provides two simplified physics models to aid scientists through every step, from beamline inception to operation:
```
cemitch99:
```suggestion
Additionally, ImpactX provides two simplified tracking modes to aid scientists through every step, from beamline inception to operation:
```

docs/source/usage/how_to_run.rst
```diff
@@ -3,87 +3,89 @@
...
+only tracking of the reference particle orbit, or tracking of the beam envelope (covariance matrix) through linearized transport maps.
```
cemitch99:
```suggestion
only tracking of the reference particle orbit, or tracking of the beam envelope (6x6 covariance matrix) through linearized transport maps.
```

docs/source/usage/how_to_run.rst
```diff
@@ -3,87 +3,89 @@
...
+================== =============== =============== ==================
+Particle Tracking  Full Dynamics   Most general    Supported
+Reference Tracking Early Design    Reference orbit No
+Envelope Tracking  Rapid Scans     Linearized      Not yet
+================== =============== =============== ==================
 
-.. code-block:: bash
 
-   mkdir -p <run_directory>
+What to run: User Interfaces
```
cemitch99:
```suggestion
How to select a user interface
```

docs/source/usage/how_to_run.rst
```diff
@@ -3,87 +3,89 @@
...
+=========== ================= ========================= =========== =========== ============== ===========
+Python      Beginner-Advanced Automation, AI/ML         Yes         Yes         Yes            Yes
+Application Advanced          Minimal requirements      No          No          Terminal       Yes
+Dashboard   Beginner          Learning, Control-Systems Yes         No          Yes            Not yet
+=========== ================= ========================= =========== =========== ============== ===========
 
-   impactx<TAB>
 
-Hitting the ``<TAB>`` key will suggest available ImpactX executables as found in your ``PATH`` `environment variable <https://en.wikipedia.org/wiki/PATH_(variable)>`__.
+How to run
```
cemitch99, line 43: I propose moving the "How to run" section to the end of this page, and specifying that the steps described are given for a local computer.  For details on how to proceed on HPC machines, we can add a link to the existing documentation.

docs/source/usage/how_to_run.rst
```diff
@@ -3,87 +3,89 @@
...
+      .. code-block:: bash
 
-   cd <run_directory>
+         impactx-dashboard
 
-   # run with an inputs file:
-   mpirun -np <n_ranks> ./impactx <input_file>
 
-or
+Where to run
```
cemitch99:
```suggestion
How to select a computing system
```

docs/source/usage/parameters.rst
```diff
@@ -11,6 +11,20 @@ This documents how to use ImpactX with an input file (``impactx input_file.in``)
    A `parser <https://amrex-codes.github.io/amrex/docs_html/Basics.html#parser>`__) is used for the right-hand-side of all input parameters that consist of one or more integers or floats, so expressions like ``beam.kin_energy = "2.+1."``, ``beam.lambdaY = beam.lambdaX`` and/or using user-defined constants are accepted.
 
 
+.. _running-cpp-parameters-mode:
+
+Physics Models
```
```suggestion
Tracking Modes
```

docs/source/usage/parameters.rst
```diff
@@ -11,6 +11,20 @@ This documents how to use ImpactX with an input file (``impactx input_file.in``)
...
+--------------
+
+* ``algo.track`` (``string``)
+
+  Physical tracking model to run:
```
```suggestion
  Mode that specifies how the beam is tracked:
```

docs/source/usage/how_to_run.rst
ax3l: Thanks! Yes, I am very flexible on these headings, thanks for the suggestions!

docs/source/usage/how_to_run.rst
ax3l: Agreed, thx!

docs/source/usage/parameters.rst
ax3l: Fun! I actually called them `modes`  in the internal anchor (2 lines above) and then used for some reason `Model` in the visible text xD

docs/source/usage/how_to_run.rst
ax3l: Agreed. Let us quickly connect if you would do another change here to discuss.

examples/fodo/README.rst
```diff
@@ -76,3 +76,65 @@ You can run the following script to visualize the beam evolution over time:
...
+* :math:`\beta_\mathrm{y} = 2.82161941` m
+* :math:`\alpha_\mathrm{y} = 1.59050035`
+
+We use a 2 GeV electron beam with initial unnormalized rms emittance of 2 nm.
+
+The second moments of the particle distribution after the FODO cell should coincide with the second moments of the particle distribution before the FODO cell, to within the level expected
+due to$
+
+In this test, the initial and final values of :math:`\lambda_x`, :math:`\lambda_y`, :math:`\lambda_t`, :math:`\epsilon_x`, :math:`\epsilon_y`, and :math:`\epsilon_t` must agree with nominal
+valu$
```
ax3l: Copy paste from a cropped terminal :)

cemitch99 (approved):

docs/source/usage/how_to_run.rst
ax3l: Resolved offline: ok for now, we will do a theory section update as a follow-up and will refine docs further.

src/python/elements.cpp
```diff
@@ -72,6 +72,22 @@ namespace
...
 
+    /** Registers the mixin LinearTransport::operator method
+     */
+    template<typename T_PyClass>
+    void register_envelope_push (T_PyClass & cl)
+    {
+        using Element = typename T_PyClass::type;  // py::class<T, options...>
+
+        cl.def("push",
+               [](Element & el, Map6x6 & cm, RefPart const & AMREX_RESTRICT ref) {
```
```suggestion
               [](Element & el, Map6x6 & cm, RefPart const & ref) {
```

src/particles/diagnostics/DiagnosticOutput.cpp
```diff
@@ -9,226 +9,216 @@
...
+                         << "emittance_x" << " " << "emittance_y" << " " << "emittance_t" << " "
+                         << "alpha_x" << " " << "alpha_y" << " " << "alpha_t" << " "
+                         << "beta_x" << " " << "beta_y" << " " << "beta_t" << " "
+                         << "dispersion_x" << " " << "dispersion_px" << " "
+                         << "dispersion_y" << " " << "dispersion_py" << " "
+                         << "emittance_xn" << " " << "emittance_yn" << " " << "emittance_tn";
+            if (compute_eigenemittances)
+            {
+                file_handler << " "
+                             << "emittance_xn" << " " << "emittance_yn" << " " << "emittance_tn";
```
ax3l, line 70: Copy-paste bug: Fix in #864

PR #815, opened 2025-01-27
Reorder Distribution Init

ax3l:
Restructure for reuse in new tracking modes.

### DIFF
--- a/src/initialization/InitElement.cpp
+++ b/src/initialization/InitElement.cpp
@@ -529,7 +529,7 @@ namespace detail
             amrex::ParticleReal ds = 0.0;
             pp_element.queryAdd("ds", ds);
 
-            elements::mixin::LinearTransport::Map6x6 transport_map = elements::mixin::LinearTransport::Map6x6::Identity();
+            Map6x6 transport_map = Map6x6::Identity();
 
             // safe to ParmParse inputs for reproducibility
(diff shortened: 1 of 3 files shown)


src/initialization/InitElement.cpp
```diff
@@ -529,7 +529,7 @@ namespace detail
             amrex::ParticleReal ds = 0.0;
             pp_element.queryAdd("ds", ds);
 
-            elements::mixin::LinearTransport::Map6x6 transport_map = elements::mixin::LinearTransport::Map6x6::Identity();
+            Map6x6 transport_map = Map6x6::Identity();
```
n01r: `Map6x6` is currently not declared anywhere

n01r (changes requested):
There seems to be a declaration or maybe include missing

src/initialization/InitElement.cpp
n01r: Maybe `LinearMap::Map6x6`?

ax3l: This is from another PR, will revert here.

```suggestion
            elements::mixin::LinearTransport::Map6x6 transport_map = elements::mixin::LinearTransport::Map6x6::Identity();
```

n01r (approved):

PR #818, opened 2025-01-29
Fix ConstF special k=0 case to address NaN's

### LINKED ISSUE #116: constf.kt = 0: handle special case
@cemitch99 found in #114 that if we set the constant focusing element in `kt` to zero that we get a division by zero somewhere.

cemitch99:
This PR adds a treatment of the special limiting cases kx = 0, ky = 0, or kt = 0 in the `ConstF` element, in order to avoid a division by zero that occurs in these cases.

This fixes #116


ax3l (approved):
Looks great, thank you!

PR #822, opened 2025-02-03
Triplet Example Plot: Pandas

ax3l:
Use pandas to select a reduced beam diags column from the CSV file instead of a hard-coded, numerical column number.

This is more robust and the reason we have CSV column headers, I think I forgot to modernize this file a while ago.


cemitch99 (approved):
Great update.  Tested - this works no problem.

PR #831, opened 2025-02-07
Mode: Reference Orbit Tracking

### LINKED ISSUE #622: Run Mode: Only Reference Particle
For a couple of workflows, e.g.,:
- converting lattice files that use focusing strengths (relative to the magnetic rigidity)
- determining phases to RF cavities (phase scan)

a first run of the reference particle as single particle would be helpful to determine properties like reference beam energy over s.

Maybe we can make this convenient, e.g., similar to `sim.evolve()` a first, optional `sim.reference_evolve()` that just pushes one particle.

- [ASTRA](https://www.desy.de/~mpyflo/Astra_manual/Astra-Manual_V3.2.pdf) calls this "auto_phase". The option "phase_scan" then uses information like this to adjust arrival times in cavities or other optimizations (?).
- Elegant calls this "[fiducialization](https://ops.aps.anl.gov/manuals/elegant_latest/elegantse5.html)" (which is their way to calculate a reference orbit from bunch properties).

ax3l:
Add a simulation mode that only tracks the reference orbit. Close #622

- [x] implementation
- [x] documentation
- [ ] example(s): follow-up PR(s)

### DIFF
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -233,6 +233,10 @@ Collective Effects & Overall Simulation Parameters
 
       Run the envelope tracking simulation loop.
 
+   .. py:method:: track_reference()
+
+      Run the reference orbit tracking simulation loop.
+
    .. py:method:: resize_mesh()
 
       Resize the mesh :py:attr:`~domain` based on the :py:attr:`~dynamic_size` and related parameters.
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -21,8 +21,8 @@ Tracking Modes
   Mode that specifies how the beam is tracked:
 
   * ``particles`` (default): symplectic particle tracking
-  * ``reference_orbit``: *not yet implemented*, only tracking of the reference particle orbit
   * ``envelope``: beam envelop (covariance matrix) tracking, through linearized transport maps
+  * ``reference_orbit``: only tracking of the reference particle orbit
 
 
 .. _running-cpp-parameters-particle:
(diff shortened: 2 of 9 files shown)


cemitch99 (approved):
Beautiful.  I also tested this on ~5 examples of different types--both app and Python input--and everything works.

ax3l:
Yay!! Do you want to push the examples into this PR or do you want to do a follow-up? :)

ax3l:
(I can also add the examples, I just want to sync with you which ones to pick first.)

cemitch99:
Hm.  I'm not sure yet what examples would be the most useful for CI and for illustration.  It might be better to come up with something new and interesting that uses this feature.  It's ok with me if we do this as a follow-up, if you're comfortable with that.

Otherwise, we're just repeating existing examples in a different mode.

ax3l:
Agreed, sounds good to me.

docs/source/usage/python.rst
```diff
@@ -233,6 +233,10 @@ Collective Effects & Overall Simulation Parameters
 
       Run the envelope tracking simulation loop.
 
+   .. py:method:: track_reference()
```
ax3l, line 236: Opps, `ref` argument missing.

ax3l: Will push a fix

cemitch99: Oh, yes I ran this, but I missed that the argument was missing in the doc here.

ax3l: I changed the API after the first doc draft and forgot to add this. Fixed now via dfe55da5c4a0344d42675c20ef6fd018242dcd94 :)

PR #832, opened 2025-02-07
Fix: SEdge in Envelope Mode

ax3l:
Assign the `sedge = s` to the correct reference particle variable in envelope tracking. Follow-up to #814


cemitch99 (approved):
Ok.  This should only affect soft-edge elements in envelope mode, none of which are yet implemented.

ax3l:
Correct :+1:

PR #833, opened 2025-02-07
Python: Do not Overwrite `PC::Redistribute`

ax3l:
This is actually already exposed in pyAMReX. The binding here overwrote it and, problematically, did not define the proper default arguments that go into `Redistribute(...)`. Just removing it thus solves this and removes the sloppy duplication that I added by accident.


cemitch99 (approved):
Looks good.

PR #836, opened 2025-02-09
Dashboard: Add documentation sidebar

proy30:
Documentation appears in a sidebar instead of opening in a new tab.

https://github.com/user-attachments/assets/ecaa1c0e-c388-490f-ae68-b7cda24e71e2

After https://github.com/ECP-WarpX/impactx/pull/834 is merged, need to make some minor changes

### DIFF
--- a/src/python/impactx/dashboard/Input/components.py
+++ b/src/python/impactx/dashboard/Input/components.py
@@ -41,32 +41,10 @@ class CardComponents:
         :param section_name: The name for the input section.
         """
 
-        def open_documentation():
-            """
-            Retrieves the documentation link with the provided section_name
-            and opens the documentation sidebar on the dashoard.
-            """
-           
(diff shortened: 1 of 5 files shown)


proy30:
Docmentation links _may_ need an update.

proy30:
Are tooltips needed after this addition given that the documentation provides the same information as tooltips would? @ax3l

ax3l:
Niceee!

I think I would still keep the tooltips for the individual fields, because it makes it easy to see what's going on and is a best practice. We can consider cropping the tooltips to the first line / first sentence we encounter in a doc string.

ax3l:
@proy30 small conflict to resolve pls, due to tooltip merge :pray:

ax3l (approved):
Splendid :tada:

PR #837, opened 2025-02-09
Dashboard: Simplify imports

proy30:
Centralized some dashboard imports. Most notably the Vuetify library, which is a step that simplifies the migration to Vuetify3 and future releases.


ax3l (approved):

ax3l:
👍

PR #840, opened 2025-02-10
Fix `beam.units = static`

ax3l:
Currently, we only support (and document correctly) "static" units, but the user was able to also pass "dynamic", upon which we did print "Dynamic units" but then still used static ones.

This now errors out correctly for anything else than "static" until we implement another variant.


cemitch99 (approved):
Yes, thanks for catching this.

PR #843, opened 2025-02-11
Dashboard: Add tooltips

### LINKED ISSUE #722: [Dashboard] Tooltips
Use Python doc strings to add tooltips (hover over) elements. Especially useful for input.

There is `.__doc__` that can be queried for this.

proy30:
### Replacing https://github.com/ECP-WarpX/impactx/pull/759, resolves https://github.com/ECP-WarpX/impactx/issues/722
The core functionality works by taking a list of class names where the docstrings are located, scanning through each property and method that matches a key in the DEFAULTS dictionary, and then outputting a dictionary with the appropriate tooltip.

For example, in `defaults.py` we can have something like:
```python
INPUT_PARAMETERS = {
    "charge_qe": -1,
}
```
When we call:
```py
get_docstrings([RefPart, ImpactX], INPUT_PARAMETERS )
```
the function iterates through each property and method  in the two classes until it finds:
```py
@property
def charge_qe(self) -> float:
    """
    Get reference particle charge (positive elementary charge)
    """
```
It then appends the extracted docstring to the output dictionary, resulting in something like:
```py
TOOLTIPS = {
    "charge_qe": "Get reference particle charge (positive elementary charge)",
}
```
**Note:** The tooltips are not added to the distribution/lattice configuration sections. A separate PR is needed to update defaults.py with the correct defaults to enable tooltips for those sections.

### DIFF
--- a/src/python/impactx/dashboard/Input/components.py
+++ b/src/python/impactx/dashboard/Input/components.py
@@ -1,6 +1,7 @@
 from typing import Optional
 
-from .. import setup_server, vuetify
+from .. import html, setup_server, vuetify
+from .defaults import TooltipDefaults
 from .generalFunctions import generalFunctions
 
 server, state, ctrl = setup_server()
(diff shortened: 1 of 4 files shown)


ax3l (approved):
Awesome, thank you! :)

PR #850, opened 2025-02-12
Make Linear Elements Fast

### LINKED ISSUE #849: Optimize Lattice: Pre-Calc Constants
We should add an additional step before we push particles where we calculate constants that only depend on the element configuration and reference particle. This will speed up particle tracking through elements like the `Quad` by roughly 4-5x and bring them on-par with `LinearElement`, without loosing generalizability (e.g., particle loss in a thick element's aperture).

ax3l:
Make linear elements as fast as possible by:
- calculating constants that depend on the element config and reference particle properties before pushing all particles,
- instead of using a Matrix-Vector calculation (as in LinearElement), we rely on the fact that those matrices are sparse and keep the multiply-adds explicit.

## Example: 1 CPU core run of a single `Quad`, 100M particles.

- System: Laptop: 12th Gen Intel(R) Core(TM) i9-12900H, performance core w/ performance power mode on.
- CXXFLAGS: `export CXXFLAGS="-Ofast -march=native"`
- CMake: default (+Python + FFT) via `-DImpactX_PYTHON=ON -DImpactX_FFT=ON`
- threading: off via `export OMP_NUM_THREADS=1`

Execution (best of 5):
```
./quad_impactx.py &
taskset -cp 6 $!
```
[quad_impactx.py.txt](https://github.com/user-attachments/files/18949267/quad_impactx.py.txt)

### Development
```
Execution time ImpactX: 5.1419 seconds (51.42ns / particle)
```

### This PR (11x faster)
```
Execution time ImpactX: 0.4675 seconds (4.67ns / particle)
```

Close #849

- [x] decide on general structure
- [x] go through all 40+ elements...

### DIFF
--- a/src/elements/CFbend.H
+++ b/src/elements/CFbend.H
@@ -81,7 +81,68 @@ namespace impactx::elements
         /** Push all particles */
         using BeamOptic::operator();
 
+        /** Compute and cache the constants for the push.
+         *
+         * In particular, used to pre-compute and cache variables that are
+         * independent of the individually tracked particle.
+         *
+         * @param refpart reference particle
+         */
+        void compute_constants (RefPart const & refpart)
+        {
+            using namespace amrex::literals; // for _rt and _prt
+
+            Alignment::compute_constants(refpart);
+
+            // length of the current slice
+            amrex::ParticleReal const slice_ds = m_ds / nslice();
+
+            // find beta*gamma^2, beta
+            amrex::ParticleReal const betgam2 = std::pow(refpart.pt, 2) - 1_prt;
+            amrex::ParticleReal const bet = refpart.beta();
+            amrex::ParticleReal const ibetgam2 = 1_prt / betgam2;
+            amrex::ParticleReal const b2rc2 = std::pow(bet, 2) * std::pow(m_rc, 2);
+
+            // update horizontal and longitudinal phase space variables
+            amrex::ParticleReal const gx = m_k + std::pow(m_rc,-2);
+            amrex::ParticleReal const omega_x = std::sqrt(std::abs(gx));
+
+            // update vertical phase space variables
+            amrex::ParticleReal const gy = -m_k;
+            amrex::ParticleReal const omega_y = std::sqrt(std::abs(gy));
+
+            // trigonometry
+            auto const [sinx, cosx] = amrex::Math::sincos(omega_x * slice_ds);
+            amrex::ParticleReal const sinhx = std::sinh(omega_x * slice_ds);
+            amrex::ParticleReal const coshx = std::cosh(omega_x * slice_ds);
+            auto const [siny, cosy] = amrex::Math::sincos(omega_y * slice_ds);
+            amrex::ParticleReal const sinhy = std::sinh(omega_y * slice_ds);
+            amrex::ParticleReal const coshy = std::cosh(omega_y * slice_ds);
+
+            amrex::ParticleReal const igbrc = 1_prt / (gx * bet * m_rc);
+            amrex::ParticleReal const iobrc = 1_prt / (omega_x * bet * m_rc);
+            amrex::ParticleReal const igobr = 1_prt / (gx * omega_x * b2rc2);
+
+            m_R11 = gx > 0_prt ? cosx : coshx;
+            m_R12 = gx > 0_prt ? sinx / omega_x : sinhx / omega_x;
+            m_R16 = gx > 0_prt ? -(1_prt - cosx) * igbrc : -(1_prt - coshx) * igbrc;
+            m_R21 = gx > 0_prt ? -omega_x * sinx : omega_x * sinhx;
+            m_R22 = gx > 0_prt ? cosx : coshx;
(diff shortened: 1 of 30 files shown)


src/elements/Quad.H
```diff
@@ -96,24 +118,14 @@ namespace impactx::elements
             amrex::ParticleReal & AMREX_RESTRICT py,
             amrex::ParticleReal & AMREX_RESTRICT pt,
             uint64_t & AMREX_RESTRICT idcpu,
-            RefPart const & refpart
+            [[maybe_unused]] RefPart const & refpart
```
ax3l: Idea: I would remove this fully from the interface to prevent that we can write inefficient code by accident.

ax3l: Discussed: this will not work to be fully removed, since the `refpart` for elements that integrate through maps is non-const.

src/elements/ConstF.H
```diff
@@ -117,23 +155,15 @@ namespace impactx::elements
...
 
-            yout = std::cos(m_ky*slice_ds)*y + sincy*py;
-            pyout = -m_ky * std::sin(m_ky*slice_ds)*y + std::cos(m_ky*slice_ds)*py;
+            yout = m_cos_kyds * y + m_sincy * py;
+            pyout = m_const_y * y + m_cos_kyds * py;
 
-            tout = std::cos(m_kt*slice_ds)*t + sinct/(betgam2)*pt;
-            ptout = -(m_kt*betgam2) * std::sin(m_kt*slice_ds)*t + std::cos(m_kt*slice_ds)*pt;
+            tout = m_cos_ktds * t + m_const_pt * pt;
+            ptout = m_const_y * t + m_cos_ktds * pt;
```
ax3l: We could actually give those members proper `m_Rij` names...

cemitch99: We will need to return the linear map for envelope tracking anyway, in a separate function within `ConstF`.  Those matrix element values could be used here.

ax3l: Thinking about it, I would actually do it the other way around: using the constants in the matrix. Can discuss tomorrow for details.

src/elements/CFbend.H
```diff
@@ -119,67 +168,44 @@ namespace impactx::elements
...
                 // advance position and momentum (defocusing)
-                x = coshx*xout + sinhx/omegax*px - (1.0_prt - coshx)/(gx*bet*m_rc)*pt;
-                pxout = omegax*sinhx*xout + coshx*px - sinhx/(omegax*bet*m_rc)*pt;
+                x = m_coshx * xout + m_sinhx / m_omega_x * px - (1.0_prt - m_coshx) * m_igbrc * pt;
+                pxout = m_omega_x * m_sinhx * xout + m_coshx * px - m_sinhx * m_iobrc * pt;
 
-                t = sinhx/(omegax*bet*m_rc)*xout + (1.0_prt - coshx)/(gx*bet*m_rc)*px
-                    + tout + r56*pt;
+                t = m_sinhx * m_iobrc * xout + (1.0_prt - m_coshx) * m_igbrc * px
+                    + tout + r56 * pt;
```
ax3l: Something is still broken for the longitudinal variable, but I cannot spot it right now.

ax3l: Ah, found it. Used `m_gx` before assignment.

ax3l:
Other speeds tests:

## CFBend: 23x Faster

```
CFBend dev:
Execution time ImpactX: 6.4189 seconds (64.18939502000285ns / particle)

CFBend pr:
Execution time ImpactX: 0.2792 seconds (2.7923822899902007ns / particle)
```

## ConstF: 11x Faster

```
ConstF dev:
Execution time ImpactX: 3.2141 seconds (32.141436460005934ns / particle)

ConstF pr:
Execution time ImpactX: 0.2992 seconds (2.9918171600002097ns / particle)
```

ax3l:
I did a comparison of ImpactX in FP64 (!) with [Cheetah using FP32 (default)](https://github.com/desy-ml/cheetah), using both codes in serial, and ImpactX is:
- 2.5x faster for `Quad`
- 2.96x faster for a simple 3D space charge test (1 slice).

Note for the Quad: Additionally to both code's `Quad` having alignment/offset support, ImpactX also supports apertures. Alone providing this feature causes an overhead, but we successfully optimize it (mostly) out now, for the benchmark where thick element apertures are not used.

Note for linear vs. nonlinear tracking: I also compared against the BMAD-X based `Quad`, which is a model for the non-linear elements in Cheetah (linear elements use matrices). The PyTorch implementation of pushing particles is significantly slower (`Drift`: 316ns / particle)

For reproducibility (currently private repo): https://github.com/BLAST-ImpactX/impactx-benchmarks/commit/19db0213670a81c34a9983c46d74a3214221815b

Update: I overlooked that Cheetah only computes in FP32 by default. We have the same option in ImpactX, but I used our default, FP64. Repeating the tests adds roughly another factor 2 speedup for ImpactX.

cemitch99 (approved):
I spot-checked the math for several elements, and I found no errors yet.  The variable names look ok to me.  As I implement the linear matrix return function element-by-element, there will be another opportunity to triple-check the math.

PR #854, opened 2025-02-20
Fix `t` in `CFbend`

ax3l:
This fixes the longitudinal component in `CFbend` in one of the cases.


cemitch99 (approved):
Yes, thanks for catching the typo here.

PR #862, opened 2025-02-25
Split up particles over multiple tiles for OpenMP

### LINKED ISSUE #847: Particle Push: OpenMP Not Working if no Collective Effects
Tested with 1 MPI rank:
It looks like setting `OMP_NUM_THREADS` to values greater than 1 neither accelerates tracking nor space charge calculations right now. This is a defect.

The init logic shows we use multiple threads, but tracking takes as long with 1 as with 4 threads as of `25.02`.

Reproducer:
- any test w/o collective effects on CPU, e.g., [FODO Cell](https://impactx.readthedocs.io/en/latest/usage/examples/fodo/README.html)

Reproducer (private link):
- https://github.com/ECP-WarpX/impactx-benchmarks/blob/main/spacecharge_impactx.py
- https://github.com/ECP-WarpX/impactx-benchmarks/blob/main/drift_impactx.py
But any other example (disable diagnostics to reduce noise) shows this as well.

atmyers:
I now get, on [this](https://github.com/BLAST-ImpactX/impactx-benchmarks/blob/main/drift_impactx.py) example, the following timings for `track_particles` on 1, 2, and 4 threads:

`1.893`
`1.017`
`0.5939`

Fix #847

### DIFF
--- a/src/particles/ImpactXParticleContainer.cpp
+++ b/src/particles/ImpactXParticleContainer.cpp
@@ -142,24 +142,24 @@ namespace impactx
             }
         }
 
-	int nthreads = 1;
+    int nthreads = 1;
 #if defined(AMREX_USE_OMP)
-	nthreads = omp_get_max_threads();
+    nthreads = omp_get_max_threads();
 #endif
 
-	const auto& ba = ParticleBoxArray(lid);
-	auto n_logical =  numTilesInBox(ba[gid], true, tile_size);
+    const auto& ba = ParticleBoxArray(lid);
+    auto n_logical =  numTilesInBox(ba[gid], true, tile_size);
 
-	if (n_logical < nthreads ) {
-	    amrex::Print() << "Too few tiles for the number of OpenMP threads. Parallelization will be poor. \n";
-	}
+    if (n_logical < nthreads ) {
+        amrex::Print() << "Too few tiles for the number of OpenMP threads. Parallelization will be poor. \n";
+    }
 
-	for (int ithr = 0; ithr < nthreads; ++ithr) {
-	    DefineAndReturnParticleTile(lid, gid, ithr);
-	}
+    for (int ithr = 0; ithr < nthreads; ++ithr) {
+        DefineAndReturnParticleTile(lid, gid, ithr);
+    }
 
-	int pid = ParticleType::NextID();
-	ParticleType::NextID(pid+np);
+    int pid = ParticleType::NextID();
+    ParticleType::NextID(pid+np);
         AMREX_ALWAYS_ASSERT_WITH_MESSAGE(
         static_cast<amrex::Long>(pid) + static_cast<amrex::Long>(np) < amrex::LongParticleIds::LastParticleID,
             "ERROR: overflow on particle id numbers");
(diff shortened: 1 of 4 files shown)


src/particles/ImpactXParticleContainer.cpp
```diff
@@ -141,25 +141,52 @@ namespace impactx
...
-        {
-            pid = ParticleType::NextID();
-            ParticleType::NextID(pid+np);
-        }
+
+    const auto& ba = ParticleBoxArray(lid);
+    auto n_logical =  numTilesInBox(ba[gid], true, tile_size);
+
+    if (n_logical < nthreads ) {
+        amrex::Print() << "Too few tiles for the number of OpenMP threads. Parallelization will be poor. \n";
```
ax3l: To avoid stdout flooding, we can use `#include <ablastr/warn_manager/WarnManager.H>`, e.g.,
https://github.com/BLAST-ImpactX/impactx/blob/61b13e453ea5e953b8a5f985539d3e6833cf088f/src/diagnostics/CovarianceMatrixMath.H#L65-L72

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -141,25 +141,52 @@ namespace impactx
...
+        int tid = omp_get_thread_num();
+        int nr = np / nthreads;
+        int nlft = np - nr*nthreads;
+
+        int num_to_add = 0;
+        int my_index =0;
+        if (tid < nlft) { // get nr+1 items
+        my_index = tid * (nr+1);
+        num_to_add = nr+1;
+        } else {         // get nr items
```
ax3l: Can you please add inline comments on the meaning of the temporary variables (tid, nr, nlft, my_index), maybe spell them out for clarity, and a bit more details on the if/else? :pray:

ax3l (commented):
Yay, thank you for the fix! :tada:

ax3l:
Oh, CI indicates we did not initialize the right number of total particles yet (or forgot them in output)?

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -181,25 +208,27 @@ namespace impactx
...
+        amrex::ParallelFor(num_to_add,
+            [=] AMREX_GPU_DEVICE (int i) noexcept
         {
-            idcpu_arr[old_np+i] = amrex::SetParticleIDandCPU(pid + i, cpuid);
+            idcpu_arr[old_np+i] = amrex::SetParticleIDandCPU(pid + my_index + i, cpuid);
 
-            x_arr[old_np+i] = x_ptr[i];
-            y_arr[old_np+i] = y_ptr[i];
-            t_arr[old_np+i] = t_ptr[i];
+            x_arr[old_np+i] = x_ptr[my_index+i];
```
ax3l: Does `my_index` only need to be added to the index on the LHS? e.g.,
```suggestion
            x_arr[old_np+my_index+i] = x_ptr[i];
```

The tile (`x_ptr`) is shorter now but the `x_arr` is still the one for the whole process?

atmyers:
Only the `run_fodo_programmable` fails now. Are we sure that example is thread-safe?

ax3l:
Yay! :rocket: :sparkles: 

Yes, I think it is actually not thread safe... Let's set those two controls to `False`:
https://github.com/BLAST-ImpactX/impactx/blob/77980670e525bf07e58fcfb1409feb46c175d3d3/examples/fodo_programmable/run_fodo_programmable.py#L120
https://github.com/BLAST-ImpactX/impactx/blob/77980670e525bf07e58fcfb1409feb46c175d3d3/examples/fodo_programmable/run_fodo_programmable.py#L129

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -107,6 +108,49 @@ namespace impactx
...
+
+        const auto& ba = ParticleBoxArray(lid);
+        auto n_logical = numTilesInBox(ba[gid], true, tile_size);
+
+        int ntry = 0;
+        while ((n_logical < nthreads) && (ntry++ < 6)) {
+            int idim = (ntry % 2) + 1;  // alternate between 1 and 2
+            tile_size[idim] /= 2;
+            n_logical = numTilesInBox(ba[gid], true, tile_size);
+        }
```
ax3l: @atmyers can you please add more inline comments to this block (line 133-141)?

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -141,61 +185,93 @@ namespace impactx
...
+        int nthreads = 1;
+#if defined(AMREX_USE_OMP)
+        nthreads = omp_get_max_threads();
 #endif
-        {
-            pid = ParticleType::NextID();
-            ParticleType::NextID(pid+np);
+
+        // split up particles over nthreads tiles
+        AMREX_ALWAYS_ASSERT(numTilesInBox(ParticleBoxArray(lid)[gid], true, tile_size) >= nthreads);
```
ax3l: @atmyers can you please add a message to this assert?

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -141,61 +185,93 @@ namespace impactx
...
-        });
+            int tid = 1;
+#if defined(AMREX_USE_OMP)
+            tid = omp_get_thread_num();
+#endif
+
+            // we split up the np particles onto multiple tiles.
+            // some tiles will get nr and some will get nlft.
+            int nr = np / nthreads;
+            int nlft = np - nr*nthreads;
```
ax3l: @atmyers can you please spell out `nr` and `nlft`?
Do they stand for `n_right` and `n_left`?

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -141,61 +185,93 @@ namespace impactx
...
+
+            int num_to_add = 0; /* how many particles this tile will get*/
+            int my_index = 0;
+
+            if (tid < nlft) { // get nr+1 items
+                my_index = tid * (nr + 1);
+                num_to_add = nr + 1;
+            } else {         // get nr items
```
ax3l: @atmyers can you please add an inline comment here? Who do some get `+1`? Due to rounding into `nthreads` chunks?

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -141,61 +185,93 @@ namespace impactx
...
+            tid = omp_get_thread_num();
+#endif
+
+            // we split up the np particles onto multiple tiles.
+            // some tiles will get nr and some will get nlft.
+            int nr = np / nthreads;
+            int nlft = np - nr*nthreads;
+
+            int num_to_add = 0; /* how many particles this tile will get*/
+            int my_index = 0;
```
ax3l: @atmyers let us pick a more descriptive variable name. Does `my_index` mean `my_thread_index` or `my_tile_index` or something else?

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -107,6 +108,49 @@ namespace impactx
...
+
+        int ntry = 0;
+        while ((n_logical < nthreads) && (ntry++ < 6)) {
+            int idim = (ntry % 2) + 1;  // alternate between 1 and 2
+            tile_size[idim] /= 2;
+            n_logical = numTilesInBox(ba[gid], true, tile_size);
+        }
+
+        if (n_logical < nthreads) {
+            amrex::Abort("ImpactParticleContainer::prepare() "
```
ax3l, line 156: @atmyers I added more more concrete guidance to users here, can you please double check it?

In which situations would we not be able to find enough tiles? Is there other guidance we should give or can we make it a `high` warning and "just" be less parallel (i.e., less parallely efficient)?

atmyers: Currently, the code will not just be efficient, it will give incorrect results if the number of tiles according to the `tile_size` is fewer than the number of threads. The issue is that the copyParticles routine in AMReX does not copy the ones that aren't on a valid tile. This resulted in some of the particles not getting written out some examples (since we us copyParticles to a pinned_pc in IO). 

This can be changed in AMReX, I'd suggest changing this to a warning rather than an Abort at that point.

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -107,6 +108,49 @@ namespace impactx
...
+        nthreads = omp_get_max_threads();
+#endif
+
+        const auto& ba = ParticleBoxArray(lid);
+        auto n_logical = numTilesInBox(ba[gid], true, tile_size);
+
+        int ntry = 0;
+        while ((n_logical < nthreads) && (ntry++ < 6)) {
+            int idim = (ntry % 2) + 1;  // alternate between 1 and 2
+            tile_size[idim] /= 2;
```
atmyers, line 147:
```suggestion
            tile_size[idim] /= 2;
            AMREX_ALWAYS_ASSERT(tile_size[idim] > 0);
```

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -107,6 +108,50 @@ namespace impactx
...
+            } else {
+                gid = *it;
+            }
+        }
+
+        int nthreads = 1;
+#if defined(AMREX_USE_OMP)
+        nthreads = omp_get_max_threads();
+#endif
+
```
atmyers, line 132:
```suggestion
    /**
        When running without space charge and OpenMP parallelization, we need to make sure we have enough tiles on level 0, grid 0 to thread over the available tiles. We try to set the tile_size appropriately here. The tiles start as (long, 8, 8) in (x, y, z). So we try to reduce the tile size in the y and z directions alternately until there are enough tiles for the number of threads. 
    */
```

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -141,61 +188,93 @@ namespace impactx
...
+        int nthreads = 1;
+#if defined(AMREX_USE_OMP)
+        nthreads = omp_get_max_threads();
 #endif
-        {
-            pid = ParticleType::NextID();
-            ParticleType::NextID(pid+np);
+
+        // split up particles over nthreads tiles
+        AMREX_ALWAYS_ASSERT(numTilesInBox(ParticleBoxArray(lid)[gid], true, tile_size) >= nthreads);
```
atmyers:
```suggestion
        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(numTilesInBox(ParticleBoxArray(lid)[gid], true, tile_size) >= nthreads, "Not enough tiles for the number of OpenMP threads - please try reducing particles.tile_size or increasing the number of cells in the domain.");
```

ax3l (approved):
Thank you, I understand the logic way better now :pray:

ax3l:
@atmyers I see segfaults on Windows right now, which is one of the few runners compiling with `-DImpactX_COMPUTE=NOACC`.

Can you compile in the same mode please and run valgrind on the failing example(s)? I suspect there is an actual bug somewhere.

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -141,61 +197,96 @@ namespace impactx
...
+
+            if (tid < n_leftover) { // get n_regular+1 items
+                my_offset = tid * (n_regular + 1);
+                num_to_add = n_regular + 1;
+            } else {         // get n_regular items
+                my_offset = tid * n_regular + n_leftover;
+                num_to_add = n_regular;
+            }
+
+            auto& particle_tile = ParticlesAt(lid, gid, tid);
```
ax3l, line 247: @atmyers this line throws if compiled with `-DImpactX_COMPUTE=NOACC`

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -141,61 +197,96 @@ namespace impactx
...
-            y_arr[old_np+i] = y_ptr[i];
-            t_arr[old_np+i] = t_ptr[i];
-
-            px_arr[old_np+i] = px_ptr[i];
-            py_arr[old_np+i] = py_ptr[i];
-            pt_arr[old_np+i] = pt_ptr[i];
-            qm_arr[old_np+i] = qm;
-            w_arr[old_np+i]  = bchchg/ablastr::constant::SI::q_e/np;
-        });
+            int tid = 1;
```
ax3l: Is the thread index zero-indexed?
```suggestion
            int tid = 0;
```

src/particles/ImpactXParticleContainer.cpp
ax3l: Fixed by defaulting `tid` to `0` not `1`.

src/particles/ImpactXParticleContainer.cpp
ax3l: Pushed this change.

PR #864, opened 2025-02-26
Fix eigenemittance column labels (CSV)

cemitch99:
This fixes an error introduced in labeling of the reduced diagnostics output file (CSV only) in the case when `eigenemittances = true`.


ax3l:
Thank you! Bug introduced in #814

PR #866, opened 2025-03-02
Add 3D space charge for envelope tracking (uncoupled)

cemitch99:
This PR adds support for 3D space charge when tracking bunches in envelope mode.  It currently assumes that the spatial block of the 6x6 covariance matrix is diagonal (``<xy> = <yt> = <tx> = 0``), which is sufficient for many purposes.

- [x] add initial elliptic integral and envelope push functions
- [x] clean up elliptic integral function
- [x] handle user-facing inputs
- [x] double-check factors of beta*gamma
- [x] add warnings for the case of a coupled covariance matrix
- [x] add return maps R needed for space charge examples
- [x] add benchmark examples
- [x] update documentation

[EnvelopeTracking_FODO_RF_Benchmark.pdf](https://github.com/user-attachments/files/19059500/EnvelopeTracking_FODO_RF_Benchmark.pdf)

Follow-up:  Add diagonalization of the 3x3 spatial block to correctly handle the fully-coupled case.


src/elements/ConstF.H
```diff
@@ -220,15 +220,51 @@ namespace impactx::elements
...
+            amrex::ParticleReal const_y = -m_ky * sin_kyds;
+            auto const [sin_ktds, cos_ktds] = amrex::Math::sincos(m_kt * slice_ds);
+            amrex::ParticleReal const_t = -m_kt * betgam2 * sin_ktds;
+
+            // intermediate quantities - to avoid division by zero
+            amrex::ParticleReal sincx = m_kx > 0 ? std::sin(m_kx * slice_ds) / m_kx : slice_ds;
+            amrex::ParticleReal sincy = m_ky > 0 ? std::sin(m_ky * slice_ds) / m_ky : slice_ds;
+            amrex::ParticleReal const sinct = m_kt > 0 ? std::sin(m_kt * slice_ds) / m_kt : slice_ds;
+            amrex::ParticleReal const_pt = sinct / betgam2;
+
```
cemitch99, line 251: The above lines can eventually replaced using a call to `compute_constants`.

ax3l: Yes, at the same time we do not have the same situation here where we call `transport_map` steps * N_p times as we do for particle tracking. So we could check.

We could generally also call `compute_constants` in the constructor, I just realized. That way, unless we change attributes again at runtime we are also well-defined upfront.

examples/CMakeLists.txt
```diff
@@ -580,6 +594,21 @@ add_impactx_test(kurth_10nC_periodic.py
     OFF  # no plot script yet
 )
 
+# Kurth Beam in a Periodic Channel with Envelope Tracking ######################
+#
+w/ space charge
```
```suggestion
# with space charge
```

src/envelope/spacecharge/Elliptic_Integral.H
```diff
@@ -0,0 +1,163 @@
+/* Copyright 2022-2023 The Regents of the University of California, through Lawrence
+ *           Berkeley National Laboratory (subject to receipt of any required
+ *           approvals from the U.S. Dept. of Energy). All rights reserved.
+ *
+ * This file is part of ImpactX.
+ *
+ * Authors: Chad Mitchell, Axel Huebl
```
ax3l, line 7:
```suggestion
 * Authors: Chad Mitchell, Axel Huebl, John Burkardt
```

src/envelope/spacecharge/Elliptic_Integral.H
```diff
@@ -0,0 +1,163 @@
...
+#include <AMReX_Print.H>
+
+#include <cmath>
+
+namespace impactx::envelope::spacecharge
+{
+
+    /** Function to return the Carlson elliptic integral denoted
+     *  by RD(x,y,z).  This is a slight modification of
+     *  the C++ function written by John Burkardt (GNU LGPL license):
```
ax3l: Since we do not literally copy the source code, I would rephrase this as:
```suggestion
     *  by RD(x,y,z).  This is a reimplementation of
     *  the C++ function written by John Burkardt (GNU LGPL license):
```
and avoid license implications from copyleft.

src/envelope/spacecharge/Elliptic_Integral.H
```diff
@@ -0,0 +1,163 @@
...
+     * @returns the elliptic integral RD(x,y,z)
+     */
+    amrex::ParticleReal
+    Elliptic_RD (
+        amrex::ParticleReal x,
+        amrex::ParticleReal y,
+        amrex::ParticleReal z,
+        amrex::ParticleReal errtol
+    )
+    {
```
ax3l: We can move the body for this function in a `src/envelope/spacecharge/Elliptic_Integral.cpp` file, so we can compile it in parallel.

src/envelope/spacecharge/Elliptic_Integral.H
```diff
@@ -0,0 +1,163 @@
...
+        amrex::ParticleReal c3;
+        amrex::ParticleReal c4;
+        amrex::ParticleReal ea;
+        amrex::ParticleReal eb;
+        amrex::ParticleReal ec;
+        amrex::ParticleReal ed;
+        amrex::ParticleReal ef;
+        amrex::ParticleReal epslon;
+        amrex::ParticleReal lamda;
+        amrex::ParticleReal const lolim = 6.0E-51;
```
ax3l: `lolim` and `uplim` should be modified for the value of `amrex::ParticleReal`, which can be `float` and `double`.

Our [build options](https://impactx.readthedocs.io/en/latest/install/cmake.html#build-options) for the non-default single precision mode are:
```console
cmake --fresh -S . -B build -DImpactX_PRECISION=SINGLE
cmake --build build -j 6
```
Single precision support (where sufficient) is really beneficial for GPU and CPU runs alike, often with a close to 2x speedup. We underutilize it a bit so far, but for instance Cheetah (and PyTorch) default to single precision (32bit floating point numbers).

For the single precision (float) case here, the current limits are out of the possible value range (-3.4e+38 to 3.4e+38), so the integrator will not work.

A typical implementation for single precision could be:
```suggestion
        amrex::ParticleReal const lolim = std::is_same_v<amrex::ParticleReal, float> ? 1.0E-30 : 6.0E-51;
```
or so, with a sensible value for the `float` case (I made one up right now).

Please include [<type_traits>](https://en.cppreference.com/w/cpp/types/is_same) for this.

There are also [<numeric_limits>](https://en.cppreference.com/w/cpp/types/numeric_limits) we can rely on, e.g., before subnormal numbers start in floats, but I assume the numbers here are picked more carefully than just "close to floating point range", so we can also set them manually for the two cases we have.

ax3l (commented):
This looks great! A small comment on the support for single precision is left.

src/envelope/spacecharge/Elliptic_Integral.H
cemitch99: Done.

src/envelope/spacecharge/Elliptic_Integral.H
cemitch99: Done.

src/envelope/spacecharge/Elliptic_Integral.cpp
```diff
@@ -0,0 +1,139 @@
...
+          value = 0.0_prt;
+        }
+
+        xn = x;
+        yn = y;
+        zn = z;
+        sigma = 0.0_prt;
+        power4 = 1.0_prt;
+
+        int max_iterations = 100;
```
ax3l, line 90: Similar to the MLMG, we could make this (in a follow-up PR) a ParmParse option.

src/envelope/spacecharge/EnvelopeSpaceChargePush.cpp
```diff
@@ -67,5 +68,66 @@ namespace impactx::envelope::spacecharge
...
+        amrex::ParticleReal const charge = refpart.charge;
+        amrex::ParticleReal const pt_ref = refpart.pt;
+        amrex::ParticleReal const betgam2 = std::pow(pt_ref, 2) - 1_prt;
+
+        // evaluate the 3D space charge intensity parameter from bunch charge
+        amrex::ParticleReal const rcN = std::abs(charge * bunch_charge) / (4_prt * pi * ep0 * mass * std::pow(c,2));
+        amrex::ParticleReal const coeff = ds * rcN / betgam2 * (1_prt/(5_prt * std::sqrt(5_prt)));
+
+        // set parameters for elliptic integrals
+        amrex::ParticleReal const errtol = 1.0e-3;
```
ax3l, line 101: Similar to MLMG options, we could in a follow-up PR make this a ParmParse option.

ax3l (approved):
Wonderful, thank you!

PR #870, opened 2025-03-04
BeamMonitor: No BP5 Group Based

### LINKED ISSUE #868: Group Based Encoding Causes Issues with BP5
See:
- https://github.com/openPMD/openPMD-api/discussions/1724
- https://github.com/BLAST-ImpactX/impactx/blob/875cb37052d788497cfd138739bf2e42ab83a742/src/elements/diagnostics/BeamMonitor.H#L84

We probably need to work on closing steps, changing encoding defaults, etc.

ax3l:
Default to BP4 for now with group based encoding.
Disallow BP5 with group based encoding.

Follow-up: default to BP5 with variable based encoding, once support is final in reading (openPMD-api).

Fix #868

cc @franzpoeschel

### DIFF
--- a/src/elements/diagnostics/BeamMonitor.cpp
+++ b/src/elements/diagnostics/BeamMonitor.cpp
@@ -126,11 +126,11 @@ namespace detail {
     {
 #ifdef ImpactX_USE_OPENPMD
         // pick first available backend if default is chosen
-        if( m_OpenPMDFileType == "default" )
+        if (m_OpenPMDFileType == "default")
 #   if openPMD_HAVE_ADIOS2==1
-        m_OpenPMDFileType = "bp";
+        m_OpenPMDFileType = "bp4";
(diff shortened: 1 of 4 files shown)


src/elements/diagnostics/BeamMonitor.cpp
```diff
@@ -126,11 +126,11 @@ namespace detail {
     {
 #ifdef ImpactX_USE_OPENPMD
         // pick first available backend if default is chosen
-        if( m_OpenPMDFileType == "default" )
+        if (m_OpenPMDFileType == "default")
 #   if openPMD_HAVE_ADIOS2==1
-        m_OpenPMDFileType = "bp";
+        m_OpenPMDFileType = "bp4";
```
ax3l, line 131: We want to change this default to bp5 as soon as variable based encoding works well for our (reader) tools.

PR #871, opened 2025-03-04
Python: Set `max_grid_size` w/ BF

ax3l:
When setting the `amr.blocking_factor_x/y/z` per dimension, also set the `amr.max_grid_size_x/y/z` to the same number in Python. Otherwise, often more grids are created than intended.

- [ ] update docs?
- [ ] update inputs files to also set `amr.max_grid_size`, too?


ax3l:
This part is generally still underdocumented in ImpactX (both for inputs and Python).

PR #872, opened 2025-03-04
[WIP] Write `impactx_used_inputs`

ax3l:
Similar to [WarpX](https://warpx.readthedocs.io/en/latest/usage/parameters.html#overall-simulation-parameters), write out the values of `amrex::ParmParse` that we use to initialize/control the simulation.

- [ ] Python control
- [x] ABLASTR: implement support to disable & quiet print (besides writing to `/dev/null`) https://github.com/ECP-WarpX/WarpX/pull/5733
- [ ] docs

### DIFF
--- a/src/ImpactX.cpp
+++ b/src/ImpactX.cpp
@@ -14,6 +14,8 @@
 #include "diagnostics/DiagnosticOutput.H"
 #include "particles/wakefields/HandleWakefield.H"
 
+#include <ablastr/utils/UsedInputsFile.H>
+
 #include <AMReX.H>
 #include <AMReX_AmrParGDB.H>
 #include <AMReX_BLProfiler.H>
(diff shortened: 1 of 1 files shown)


src/ImpactX.cpp
```diff
@@ -118,6 +120,15 @@ namespace impactx {
             }
         }
 
+        // write parameters
+        {
+            std::string filename = "impactx_used_inputs";
+            amrex::ParmParse pp_impactx("impactx");
+            pp_impactx.queryAdd("used_inputs_file", filename);
+
+            ablastr::utils::write_used_inputs_file(filename);
```
ax3l, line 129:
```suggestion
            ablastr::utils::write_used_inputs_file(filename, verbose);
```

ax3l: Note: to use https://github.com/AMReX-Codes/amrex/pull/4652 now

PR #874, opened 2025-03-06
Correct units in thin multipole documentation.

cemitch99:
Corrects an error in the documentation of units for the "integrated multipole strength" input field of the thin multipole element.


ax3l (approved):
Thank you!

PR #829, opened 2025-02-07
[WIP] Dashboard: Variable Referencing in Lattice Configuration

proy30:
Allows dashboard users to define variable names, assign numeric values, and reference these variables in lattice configurations.

**Variable Naming Requirements:**
* The variable name must begin with an alphabetical character

**Variable Value Requirements:**
* The value must be numeric.

**To Do:**
- [X] Implement core functionality
- [ ] Resolve UI reactivity issues
   - The current nested structure severely affects reactivity.

PR #876, opened 2025-03-07
Dashboard: Add impactx examples

proy30:
Dashboard users can now pick and choose which ImpactX example to load on the dashboard. Serves as a great user feature and a good testing functionality to evaluate the performance of the dashboard parser and it's ability to handle various types of inputs.

While users can now select any impactx example, the dashboard parser needs updating for more flexibility in reading different types of file formats and the addition of variable referencing for the lattice configuration (https://github.com/BLAST-ImpactX/impactx/pull/829) needs to be completed for better usage. 

TODO:
- [x] Merge #930 
- [x] Merge #981
- [ ] Merge https://github.com/BLAST-ImpactX/impactx/pull/1140

### DIFF
--- a/src/python/impactx/dashboard/Toolbar/examples_loader/loader.py
+++ b/src/python/impactx/dashboard/Toolbar/examples_loader/loader.py
@@ -25,7 +25,6 @@ DASHBOARD_EXAMPLES = {
     # "iota_lattice/run_iotalattice.py",
     # "cyclotron/run_cyclotron.py",
     # "dogleg/run_dogleg.py",
-
 }
 
 
--- a/src/python/impactx/dashboard/Toolbar/examples_loader/loader.py
+++ b/src/python/impactx/dashboard/Toolbar/examples_loader/loader.py
@@ -23,7 +23,7 @@ DASHBOARD_EXAMPLES = {
     "kurth/run_kurth_10nC_periodic.py",
     "expanding_beam/run_expanding_fft.py",
     "expanding_beam/run_expanding_envelope.py",
-    "iota_lattice/run_iotalattice.py",
+    # "iota_lattice/run_iotalattice.py",
     "cyclotron/run_cyclotron.py",
     "dogleg/run_dogleg.py",
 }
--- a/src/python/impactx/dashboard/Toolbar/examples_loader/loader.py
+++ b/src/python/impactx/dashboard/Toolbar/examples_loader/loader.py
@@ -1,0 +1,92 @@
+"""
+This file is part of ImpactX
+Copyright 2025 ImpactX contributors
+Authors: Parthib Roy
+License: BSD-3-Clause-LBNL
+"""
+
+from pathlib import Path
+
+from ... import state
+from ...Input.utils import GeneralFunctions
+from ..file_imports.ui_populator import populate_impactx_simulation_file_to_ui
+
+
+state.impactx_example_list = []
+
+
+DASHBOARD_EXAMPLES = {
+    "fodo/run_fodo.py",
+    "chicane/run_chicane_csr.py",
+    "fodo_space_charge/run_fodo_envelope_sc.py",
+    "apochromatic/run_apochromatic.py",
+    "kurth/run_kurth_10nC_periodic.py",
+    "expanding_beam/run_expanding_fft.py",
+    "expanding_beam/run_expanding_envelope.py",
+    "iota_lattice/run_iotalattice.py",
+    "cyclotron/run_cyclotron.py",
+    "dogleg/run_dogleg.py",
+}
+
+
+class DashboardExamplesLoader:
+    @state.change("impactx_example")
+    def on_selected_impactx_example_change(**kwargs):
+        if state.impactx_example is None:
+            GeneralFunctions.reset_inputs("all")
+        if state.impactx_example in state.impactx_example_list:
+            example_script = DashboardExamplesLoader._get_example_content(
+                state.impactx_example
+            )
+            if example_script:
+                populate_impactx_simulation_file_to_ui(example_script)
+
+    @staticmethod
+    def get_impactx_path() -> Path:
+        """
+        Helper method to find the impactx/examples parent directory.
+        For now, just utilized to load the file names of the impactx examples
+        and retrieve the script in string format.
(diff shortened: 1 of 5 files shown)


proy30:
https://github.com/user-attachments/assets/cfeb8ec1-10f9-495b-a4e3-a023920c97f9

proy30:
Another touchup PR could focus on giving users better selection options.. following something similar to:

https://impactx.readthedocs.io/en/latest/usage/examples.html

For example in the dropdown:

Space Charge
- [All examples of space charge examples]

CSR
-  [All examples of CSR]

etc..


I notice an issue right away with an example such as `cfchannel/run_cfchannel_10nC_fft.py`. Would be nice if `sc` was included in the file name to indicate this is a space charge example, as would making parsing easier.

ax3l:
This is really cool, but since examples can use arbitrary syntax my feeling is that this can be hard to maintain... e.g., the `BeamMonitor` in the examples is usually a separate variable, and we can write arbitrary functions to create lattices or particle input...

One way that would be easier to maintain, but is different than your current approach, would be to only read examples from their "inputs" (executable/app) files, because that is a simple key-value syntax that can easily be parsed.

To discuss :)

ax3l:
Discussed today: simplifying to an allow-list of explicitly working examples (e.g., the most common 5-20 that fully work) would be sufficient and create no confusion.

ax3l:
@proy30 if you rebase this PR, is it ready to review/merge? :)

proy30:
@ax3l The current dashboard parser isn’t reliable enough to accurately load all of the initial set of ImpactX examples we selected for the dashboard.

I worked on a branch over a month (or two) ago that improves things — it successfully imports most of those examples (only `iotalattice.py` I believe is still failing), but it’s not fully robust. From what I’ve seen online, using an `ast`-based approach (https://docs.python.org/3/library/ast.html) would be a better long-term solution, but it would require a full refactor of what I have, which I haven’t had time to do yet.

We *could* merge the parser improvements I have now and move forward with this PR. I have to rebase it and touch up on a couple of points and then make a PR for it. Just a heads-up that we would definitely need a follow-up PR to update the parser again.

ax3l:
Ok. I mean what we should generally do is some try-catch parsing that simply skips examples from the list in the GUI that throw errors in parsing. Can you add that logic?

Then we can see if we restructure a few examples that fail and others that are not parsable easily we simply skip in the GUI.

cemitch99 (commented):
This looks great, thanks.  Are you still waiting on #1140 ?

proy30:
Yes I am @cemitch99. a note is that some of the examples that are in this PR might not be supported even with the changes in #1140. I will update the list of supported examples in #1140.

1140 will be complete within the next week.

Edit:

The core funcitonality of this PR (allowing the use of examples) is what makes the PR "ready". Now it is on #1140 to make the parsing correct.

src/python/impactx/dashboard/Toolbar/examples_loader/loader.py
```diff
@@ -0,0 +1,91 @@
...
+    "fodo/run_fodo.py",
+    "chicane/run_chicane_csr.py",
+    "fodo_space_charge/run_fodo_envelope_sc.py",
+    "apochromatic/run_apochromatic.py",
+    "kurth/run_kurth_10nC_periodic.py",
+    "expanding_beam/run_expanding_fft.py",
+    "expanding_beam/run_expanding_envelope.py",
+    # "iota_lattice/run_iotalattice.py",
+    "cyclotron/run_cyclotron.py",
+    "dogleg/run_dogleg.py",
```
ax3l: Let's comment them in again with the parser PR, currently only FODO works:
```suggestion
    "fodo/run_fodo.py",
    # "chicane/run_chicane_csr.py",
    # "fodo_space_charge/run_fodo_envelope_sc.py",
    # "apochromatic/run_apochromatic.py",
    # "kurth/run_kurth_10nC_periodic.py",
    # "expanding_beam/run_expanding_fft.py",
    # "expanding_beam/run_expanding_envelope.py",
    # "iota_lattice/run_iotalattice.py",
    # "cyclotron/run_cyclotron.py",
    # "dogleg/run_dogleg.py",

```

ax3l (approved):

cemitch99 (approved):

src/python/impactx/dashboard/Toolbar/examples_loader/loader.py
```diff
@@ -0,0 +1,91 @@
...
+    def load_impactx_examples() -> None:
+        """
+        Loads only the ImpactX example files defined in DASHBOARD_EXAMPLES
+        to state.impact_example_list.
+        """
+
+        state.impactx_example_list.clear()
+
+        impactx_directory = DashboardExamplesLoader.get_impactx_path()
+        impactx_examples_directory = impactx_directory / "examples"
```
ax3l, line 85: There is an issue here, where a `pip install`-ed ImpactX returns `None` for `impactx_directory`.
https://github.com/BLAST-ImpactX/impactx/pull/1140#pullrequestreview-3255528217

We do not currently package the examples dir and this needs to be added. Then, the path finding here needs to be generalized.

src/python/impactx/dashboard/Toolbar/examples_loader/loader.py
```diff
@@ -0,0 +1,91 @@
...
+    @staticmethod
+    def get_impactx_path() -> Path:
+        """
+        Helper method to find the impactx/examples parent directory.
+        For now, just utilized to load the file names of the impactx examples
+        and retrieve the script in string format.
+        Potentially, this could be used to locate other impactx directories.
+        """
+
+        current_directory = Path(__file__).resolve()
```
ax3l, line 52: For a pip-installed ImpactX, this reads something like:
```
/home/axel/micromamba/envs/warpx-cpu-mpich-dev/lib/python3.12/site-packages/impactx/dashboard/Toolbar/examples_loader/loader.py
```

The result of this function is `None`.

src/python/impactx/dashboard/Toolbar/examples_loader/loader.py
ax3l: Fix in #1179

PR #891, opened 2025-03-12
Fix numpy ODR

ax3l:
Pybind headers need to be included consistently in all translation units to avoid one-definition-rule violations.
- https://pybind11.readthedocs.io/en/stable/advanced/cast/index.html
- https://pybind11.readthedocs.io/en/stable/advanced/cast/overview.html#conversion-table
- https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html#arrays

This might help with the internal compiler error I see on MSVC 64bit on https://github.com/BLAST-ImpactX/impactx-wheels

### DIFF
--- a/src/python/pyImpactX.H
+++ b/src/python/pyImpactX.H
@@ -10,9 +10,10 @@
 #define IMPACTX_PYIMPACTX_H
 
 #include <pybind11/pybind11.h>
+#include <pybind11/functional.h>
+#include <pybind11/numpy.h>
 #include <pybind11/stl.h>
 #include <pybind11/stl_bind.h>
-#include <pybind11/functional.h>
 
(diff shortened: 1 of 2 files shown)


src/python/pyImpactX.H
```diff
@@ -10,9 +10,10 @@
 #define IMPACTX_PYIMPACTX_H
 
 #include <pybind11/pybind11.h>
+#include <pybind11/functional.h>
+#include <pybind11/numpy.h>
```
ax3l: I am not even sure we still use this. Could have been left over from development.

This could be tried by commenting, once the CI passes:
```suggestion
// include <pybind11/numpy.h>  // not yet used
```

PR #896, opened 2025-03-15
Python: TinyProfiler Control

### LINKED ISSUE #894: Python: Disable Tiny Profiler Output
Question by @carlandreaslindstrom: How can I disable the verbose tables at the end of a Python simulation?

### Quick Work-Around

This can be turned off: before the simulation is run, to silence the output from Tiny Profiler that creates these tables when a simulation concludes:
```python
import amrex.space3d as amr

# add before the simulation setup
pp_prof = amr.ParmParse("tiny_profiler")
pp_prof.add("enabled", 0)
```

(ref: new feature that [was added here](https://github.com/AMReX-Codes/amrex/pull/4100))

### Complete Answer

I will add a `sim...` option for this to turn if off, so it is less awkward to control than the work-around above.

ax3l:
Add Python properties to control the tiny profiler at runtime.

Close #894

### DIFF
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -232,6 +232,23 @@ Collective Effects & Overall Simulation Parameters
       Controls how much information is printed to the terminal, when running ImpactX.
       ``0`` for silent, higher is more verbose. Default is ``1``.
 
+   .. py:property:: tiny_profiler
+
+      This parameter can be used to disable tiny profiling including CArena memory profiling at runtime.
+      Default is ``True``.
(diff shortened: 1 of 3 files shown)


ax3l:
@carlandreaslindstrom this adds
```python
sim.tiny_profiler = False
```
to ImpactX 25.03+, to disable the tables at the end of a simulation.

If you want to have the simulation even more quiet, add also
```python
sim.verbose = 0
```

EZoni (approved):

PR #899, opened 2025-03-18
[WIP] CI: ImpactX Single Precision

ax3l:
Add a runner for single precision tests in CI.

- [ ] make s-to-t / t-to-s transform more robust?
- [ ] make reduced beam characteristics more robust?
- [ ] adjust tolerances

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -180,9 +180,48 @@ jobs:
     - name: validate created openPMD files
       run: find build -name *.h5 | xargs -n1 -I{} openPMD_check_h5 -i {}
 
-    - name: build ImpactX (SP) w/o running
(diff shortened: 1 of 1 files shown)


ax3l:
Looking at the t-to-s / s-to-t test in `tests/python/test_transformation.py`, which underlies the space charge implementation, I think that either that transformation or the `ReducedBeamDiagnostics` that we use to validate it need improvements to deal better with single precision.

We might be able to re-formulate/taylor the expressions in the transformations where we have terms like `-1.0 + 1.000001` first:
- https://github.com/BLAST-ImpactX/impactx/blob/development/src/particles/transformation/ToFixedT.H
- https://github.com/BLAST-ImpactX/impactx/blob/development/src/particles/transformation/ToFixedS.H

PR #902, opened 2025-03-18
PyTorch Example: HDF5

ax3l:
In CI/Examples we use `.h5` for the `BeamMonitor`, because it is easier to more install/broadly available in environments. Update this for our PyTorch example to be in sync with the other examples we have.


ax3l:
We can even relax this for all beam monitors and use a wildcard in analysis:
https://github.com/openPMD/openPMD-api/pull/1584

Will do in a follow-up PR: #904

PR #909, opened 2025-03-19
2D Space Charge in Particle Tracking

### LINKED ISSUE #401: 2D/2.5D Space Charge
We currently implement a 3D space charge solver, but for many cases a 2D (transverse) solver would be sufficient.

There are two options we are interested in, for different science cases:
- [ ] the AMReX HiPACE++ 2D solver (for 3D sims) by @WeiqunZhang - potentially best for EB setups
- [ ] @qianglbl 's 2D spectral symplectic solver - potentially best for open boundary scenarios

ax3l:
Add 2D Space Charge in Particle Tracking. First part of #401

## To Do

- [x] [turn off the s-to-t transformation](https://github.com/BLAST-ImpactX/impactx/issues/401#issuecomment-1641208672)
- [x] flatten charge beam (to an average surface density) 
  - AMReX: https://github.com/AMReX-Codes/amrex/pull/4384 https://github.com/AMReX-Codes/amrex/pull/4566
  - geom.domain() for level 0, minimalBox() for higher levels 
  - WarpX/ABLASTR: https://github.com/BLAST-WarpX/warpx/pull/6127
- [x] fix solver bugs upstream
  - [x] AMReX: https://github.com/AMReX-Codes/amrex/pull/4671
  - [x] AMReX: https://github.com/AMReX-Codes/amrex/pull/4703
  - [x] WarpX/ABLASTR: https://github.com/BLAST-WarpX/warpx/pull/6199
  - [x] pull in latest AMReX & ABLASTR after https://github.com/BLAST-WarpX/warpx/pull/6201 / #1157 / #1177
- [x] implement 2D (transverse) force from self fields
- [x] update gather
- [x] update push
- [x] CI test
- [x] user-facing documentation

For this to compile, you need to set:
`cmake --fresh -S . -B build -DImpactX_FFT=ON -DImpactX_ablastr_branch=a76eecf1553f5d0e72ec948c9a67db4111a1544f` (WarpX commit of the merged PR https://github.com/BLAST-WarpX/warpx/pull/6201 )

## Follow-up

- MR support (fix segfaults in OpenBC - backward_doit - ParallelCopy)
- https://github.com/BLAST-ImpactX/impactx/issues/1121
  - [ ] calculate 1D current along t (for a fixed s) -- take model from CSR? expose resolution as separate option?


src/particles/spacecharge/GatherAndPush.cpp
```diff
@@ -76,32 +80,66 @@ namespace impactx::particles::spacecharge
...
+                                x, y, z,
+                                scf_arr_x, scf_arr_y, scf_arr_z,
+                                invdr,
+                                prob_lo
+                            );
+
+                        // push momentum
+                        px += field_interp[0] * push_consts;
+                        py += field_interp[1] * push_consts;
+                        pz += field_interp[2] * push_consts;  // TODO: is this always += 0.0?
```
ax3l: Non-zero, because we will model a variation of the current density along t.

Might in practice often be neglected. We can start without long. kick and add it later.

src/particles/spacecharge/GatherAndPush.cpp
```diff
@@ -76,32 +80,66 @@ namespace impactx::particles::spacecharge
...
+                        amrex::ParticleReal & AMREX_RESTRICT x = part_x[i];
+                        amrex::ParticleReal & AMREX_RESTRICT y = part_y[i];
+                        amrex::ParticleReal z = 0.0_prt;  // flatten 3rd dimension
+                        amrex::ParticleReal & AMREX_RESTRICT px = part_px[i];
+                        amrex::ParticleReal & AMREX_RESTRICT py = part_py[i];
+                        amrex::ParticleReal & AMREX_RESTRICT pz = part_pz[i];
+
+                        // force gather
+                        amrex::GpuArray<amrex::Real, 3> const field_interp =
+                            ablastr::particles::doGatherVectorFieldNodal<2>(
```
ax3l, line 103: For this to compile, you need to set:
```
cmake --fresh -S . -B build -DImpactX_FFT=ON -DImpactX_ablastr_branch=738d7bd87a455a484297a45c1b52246a45af207f
```

cemitch99:
The test here currently fails with the simple error:

```
domain size in direction 2 is 1
blocking_factor is 8
amrex::Error::0::domain size not divisible by blocking_factor !!!
```

We should make this error conditional on the space charge solver.  This solver requires a grid size of 1 in z.

src/particles/spacecharge/GatherAndPush.cpp
```diff
@@ -76,32 +80,66 @@ namespace impactx::particles::spacecharge
...
+                                x, y, z,
+                                scf_arr_x, scf_arr_y, scf_arr_z,
+                                invdr,
+                                prob_lo
```
cemitch99:
```suggestion
                                prob_lo_2D
```

cemitch99 (commented):
Need to touch base on the scale/normalization of the fields returned by the 2D solve, in order to correctly set the value of `push_constants` in `GatherAndPush.cpp`.

src/particles/spacecharge/ForceFromSelfFields.cpp
```diff
@@ -52,11 +56,22 @@ namespace impactx::particles::spacecharge
...
                 auto scf_arr_z = space_charge_field[lev]["z"][mfi].array();
 
-                amrex::ParallelFor(bx, [=] AMREX_GPU_DEVICE (int i, int j, int k) noexcept {
-                    scf_arr_x(i, j, k) = inv2dr[0] * (phi_arr(i-1, j, k) - phi_arr(i+1, j, k));
-                    scf_arr_y(i, j, k) = inv2dr[1] * (phi_arr(i, j-1, k) - phi_arr(i, j+1, k));
-                    scf_arr_z(i, j, k) = inv2dr[2] * (phi_arr(i, j, k-1) - phi_arr(i, j, k+1));
-                });
+                if (space_charge == SpaceChargeAlgo::True_2D) {
+                    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(bx.size()[2] == 1,
+                                                     "2D space charge requires exactly 1 slice in z");
```
cemitch99, line 61: These lines cause a problem:  If run with `amr.n_cell = 32 32 1`, this error is still thrown.  I had to comment this in order to run tests.

ax3l: Can you print the value of
bx.size()[2]
?

cemitch99: Interesting.  Here is the input:
```
amr.n_cell = 16 16 1
amr.blocking_factor_x = 8
amr.blocking_factor_y = 8
amr.blocking_factor_z = 1
```

Here is the output:
`bx.size()[2] = 2`

cemitch99: In general, `bx.size()[2] = nz+1`, where `nz` is the third value following `amr.n_cell = ...`.

ax3l: Could be from the particle shape (guard?).

cemitch99, line 61:
```suggestion
                if (space_charge == SpaceChargeAlgo::True_2D) {
                    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(bx.size()[2]-1 == 1,
                                                     "2D space charge requires exactly 1 slice in z");
```

cemitch99:
Here is a more interesting observation.  I ran with an axisymmetric KV beam, expanding in free space, with 2D space charge for an identical number of grid points in x and y, but the envelopes show significantly different evolution.  This is evidence of x/y asymmetry in the underlying 2D Poisson solve.  The push constants need to be updated, but they are identical in x and y.
[Expanding_beam_2D_envelopes.pdf](https://github.com/user-attachments/files/22306865/Expanding_beam_2D_envelopes.pdf)

```ini
# Particle Beam(s)
beam.npart = 100000  # outside tests, use 1e5 or more
beam.units = static
beam.kin_energy = 250.0
beam.current = 1.0e-3
beam.particle = proton
beam.distribution = kvdist
beam.lambdaX = 4.472135955e-4
beam.lambdaY = beam.lambdaX
beam.lambdaT = 9.12241869e-7
beam.lambdaPx = 0.0
beam.lambdaPy = 0.0 
beam.lambdaPt = 0.0

# Beamline: lattice elements and segments
lattice.elements = monitor drift1 monitor
lattice.nslice = 40

drift1.type = drift
drift1.ds = 6.0

monitor.type = beam_monitor
monitor.backend = h5

# Algorithms
algo.particle_shape = 2
algo.space_charge = 2D
algo.poisson_solver = "fft"

amr.n_cell = 16 16 1
amr.blocking_factor_x = 8
amr.blocking_factor_y = 8
amr.blocking_factor_z = 1

geometry.prob_relative = 1.01
```

src/particles/spacecharge/ForceFromSelfFields.cpp
cemitch99: Revisit this to force `bx.size()[2] == 1.`

ax3l: fixed! forgot to make `phi` actually 2D between the functions. Had it originally in 3D before we had https://github.com/BLAST-WarpX/warpx/pull/6127

ax3l: fixed

src/particles/spacecharge/GatherAndPush.cpp
```diff
@@ -76,32 +81,67 @@ namespace impactx::particles::spacecharge
...
+                            ablastr::particles::doGatherVectorFieldNodal<2>(
+                                x, y, z,
+                                scf_arr_x, scf_arr_y, scf_arr_z,
+                                invdr,
+                                prob_lo_2D
+                            );
+
+                        // push momentum
+                        px += field_interp[0] * push_consts * beta_gamma * dr[2] / (c0_SI);
+                        py += field_interp[1] * push_consts * beta_gamma * dr[2] / (c0_SI);  //this should be field_interp[1]
```
cemitch99: The constants appearing in px and py are currently in draft form only, and will continue to be updated.

ax3l:
Yay, [now](https://github.com/BLAST-WarpX/warpx/pull/6201) we are getting there :)

Expanding beam example (with one kick):

## projected charge density (initial)
charge_slice

## SC potential (from initial density)
phi

## projected charge density after kick
charge_projected_after

src/particles/spacecharge/PoissonSolve.cpp
```diff
@@ -73,16 +79,41 @@ namespace impactx::particles::spacecharge
...
+            if (space_charge == SpaceChargeAlgo::True_2D) {
+                // 2D phi
+                auto & r2d = rho_2d[lev].second;
+                auto nGrow = phi[lev].nGrowVect();
+                nGrow[2] = 0;
+                phi.erase(lev);
+                phi.emplace(
+                    lev,
+                    amrex::MultiFab{r2d.boxArray(), r2d.DistributionMap(), r2d.nComp(), nGrow}
+                );
```
ax3l: Clean-up thought for later: From a book-keeping perspective, I am not yet sure if re-allocating (rho and) phi here to 2D is the most sensible thing to do or if we just add persistent `rho_2d` and `phi_2d` member variables alongside the simulation `rho` and `phi` fields (in `amr_data`) to avoid confusion.

ax3l:
@cemitch99 just a word of warning: I rebased this branch against `development` earlier today.
The easiest way to edit it locally is to remove the local branch you have (`git branch -D topic-space-charge-particles-2D`) and then check it out again:
```
git fetch ax3l
git checkout -b topic-space-charge-particles-2D ax3l/topic-space-charge-particles-2D
```
(assuming `git remote -v` shows my fork as `ax3l`).

src/particles/spacecharge/GatherAndPush.cpp
cemitch99: The above comment now appears to be obsolete.

src/particles/spacecharge/GatherAndPush.cpp
```diff
@@ -76,32 +82,67 @@ namespace impactx::particles::spacecharge
...
+                                x, y, z,
+                                scf_arr_x, scf_arr_y, scf_arr_z,
+                                invdr,
+                                prob_lo_2D
+                            );
+
+                        // push momentum
+                        px += field_interp[0] * push_consts * dr[2] / (30_prt * std::sqrt(beta_gamma) * c0_SI);
+                        py += field_interp[1] * push_consts * dr[2] / (30_prt * std::sqrt(beta_gamma) * c0_SI);  //this should be field_interp[1]
+                        pz += 0.0_rt;
```
cemitch99: The constants here are still in draft form, and have been determined empirically by testing scaling with various input parameters.  These will be updated.

cemitch99:
Somehow, these changes have broken the existing 3D space charge benchmark tests...!  Under investigation...

The error is significant, but appears to be localized to this branch.

src/initialization/AmrCoreData.cpp
```diff
@@ -145,9 +147,11 @@ namespace impactx::initialization
                 amrex::MultiFab{amrex::convert(cba, rho_nodal_flag), dm, num_components_rho, num_guards_rho, tag("rho")});
 
         // scalar potential
-        auto const phi_nodal_flag = rho_nodal_flag;
+        amrex::IntVect phi_nodal_flag = rho_nodal_flag;
         int const num_components_phi = 1;
-        int const num_guards_phi = num_guards_rho + 1; // todo: I think this just depends on max(MLMG, force calc)
+        amrex::IntVect num_guards_phi{num_guards_rho + 1}; // todo: I think this just depends on max(MLMG, force calc)
+        if (space_charge == SpaceChargeAlgo::True_2D) { num_guards_phi[2] = 0; phi_nodal_flag[2] = 0; }
+        num_guards_phi[2] = 0;
```
cemitch99:
```suggestion
```

cemitch99: This line, appearing outside the conditional, affects the 3D Poisson solve, and seems to be responsible for breaking those tests.  I suspect it is not needed here.

cemitch99:
Problem fixed.

cemitch99:
The potential output from the 2D Poisson solve does not appear to converge with  the number of cells `nx`, `ny` as specified in `sim.n_cell = [nx, ny, nz]`.  Below are results from `test_charge_deposition_2D.py`, obtained by varying these parameters only.  The effect is apparent in the dynamical expanding beam test as well.

Fig_16x16
Fig_32x32

WeiqunZhang:
https://github.com/AMReX-Codes/amrex/pull/4703

cemitch99:
Thanks @WeiqunZhang !  [PR#1177](https://github.com/BLAST-ImpactX/impactx/pull/1177) appears to address the issue of numerical convergence with respect to the transverse grid [`nx`,`ny`].  Note that this PR also affects the overall scaling of the potential.  The correct scale factor to use here is still not clear to me, although some of the parameter dependencies are known.

cemitch99:
The Python versions of these tests yield _slightly_ different results when compared to the tests using the C++ input files, and as a result are expected to fail here.  Some additional eyes locating the source of those differences would help.

cemitch99:
The difference between the Python and app/C++ benchmark tests may be due to a difference in grid construction, despite similar inputs.  Below is a diff of the log output:

> ==> log_app <==
> Initializing AMReX (25.10-19-g546a888ef041)...
> OMP initialized with 12 OMP threads
> AMReX (25.10-19-g546a888ef041) initialized
> 
> Grids Summary:
>   Level 0   **1 grids**  1024 cells  100 % of domain
> 
> Beam kinetic energy (MeV): 250
> Bunch charge (C): 0.15
> Particle type: proton
> 
> ==> log_py <==
> Initializing AMReX (25.10-19-g546a888ef041)...
> OMP initialized with 12 OMP threads
> AMReX (25.10-19-g546a888ef041) initialized
> 
> Grids Summary:
>   Level 0   **4 grids**  1024 cells  100 % of domain
>             smallest grid: 16 x 16 x 1  biggest grid: 16 x 16 x 1
>

examples/CMakeLists.txt
```diff
@@ -1675,3 +1675,35 @@ add_impactx_test(solenoid-softedge-solvable.py
...
+# with space charge
+add_impactx_test(expanding-fft-2d
+    examples/expanding_beam/input_expanding_fft_2D.in
+    ON   # ImpactX MPI-parallel
+    examples/expanding_beam/analysis_expanding_fft_2D.py
+    OFF  # no plot script yet
+)
+add_impactx_test(expanding-fft-2d.py
+    examples/expanding_beam/run_expanding_fft_2D.py
+    OFF   # ImpactX MPI-parallel
```
ax3l, line 1713: If we want 100% the same AMReX blocks ("grids output"):
```suggestion
    ON    # ImpactX MPI-parallel
```

examples/CMakeLists.txt
```diff
@@ -1675,3 +1675,35 @@ add_impactx_test(solenoid-softedge-solvable.py
...
+# with space charge
+add_impactx_test(fodo-2d-sc
+    examples/fodo_space_charge/input_fodo_2d_sc.in
+    ON   # ImpactX MPI-parallel
+    examples/fodo_space_charge/analysis_fodo_2d_sc.py
+    OFF  # no plot script yet
+)
+add_impactx_test(fodo-2d-sc.py
+    examples/fodo_space_charge/run_fodo_2d_sc.py
+    OFF   # ImpactX MPI-parallel
```
ax3l: If we want 100% the same AMReX blocks ("grids output"):
```suggestion
    ON    # ImpactX MPI-parallel
```

examples/fodo_space_charge/analysis_fodo_2d_sc.py
```diff
@@ -0,0 +1,98 @@
...
+print("")
+print("Final Beam:")
+sig_xf, sig_yf, sig_tf, emittance_xf, emittance_yf, emittance_tf = get_moments(final)
+print(f"  sigx={sig_xf:e} sigy={sig_yf:e} sigt={sig_tf:e}")
+print(
+    f"  emittance_x={emittance_xf:e} emittance_y={emittance_yf:e} emittance_t={emittance_tf:e}"
+)
+
+atol = 0.0  # ignored
+rtol = 3.6 * num_particles**-0.5  # from random sampling of a smooth distribution
```
cemitch99:
```suggestion
rtol = 3.7 * num_particles**-0.5  # from random sampling of a smooth distribution
```

cemitch99:
FYI:  I had to relax the tolerance a great deal for the Python tests, especially the FODO test, while the C++/app tests passed with _no problem_.  I'm still concerned that there may be something else going on specific to the Python test implementation or bindings.

cemitch99:
Screenshot 2025-10-17 at 3 11 30 PM
Note:  Something is still not quite right.

cemitch99:
@WeiqunZhang and @ax3l The test mentioned earlier today for this PR is located in: `examples/expanding_beam/`

C++ input file:  `input_expanding_fft_2D.in`
Python input file:  `run_expanding_fft_2D.py`

I have added a script for plotting the x-y beam size evolution called `plot_expanding_fft_2D.py`.

If the test works, the beam size at the end (at the maximum value of $s$ shown in the plot) should be 1 mm.

To see the convergence issue, I suggest running the Python test with npart = 1M for a few values of `sim.blocking_factor_x` and `sim.blocking_factor_y`.  You can see the effect this has on the beam moments by running the plot script in each case.

cemitch99:
@WeiqunZhang and @ax3l **_Note:_**  I tried the same test using 3D space charge (using `examples/expanding_beam/run_expanding_fft.py` after turning off mesh refinement), and this issue does not occur.

cemitch99:
@WeiqunZhang and @ax3l   Here is the latest update:  If I run the app example "input_expanding_fft_2D.in", then I always see that 1 grid is used, independently of the choice of blocking factor ("1 grids  1024 cells"), and the results are consistent:
ExpandingBeam_app_input
However, if I run the corresponding Python example "run_expanding_fft_2D.py", then as I change the blocking factor, the number of grids varies (bf = 32, grids = 1; bf = 16, grids = 4; bf = 8, grids = 16; bf = 4, grids = 64; bf = 2, grids = 256).  The results are consistent for bf >= 8, but disagree for smaller bf.  Here, the number of cells is [32,32,1] in each case.
ExpandingBeam_py_input


This issue has now been fixed.  Thanks @WeiqunZhang !

src/particles/spacecharge/PoissonSolve.cpp
WeiqunZhang:
```suggestion
```
`phi` should not be reset. `rho_2d[lev].second` has its own BoxArray and DistributionMapping that can be very different from ImpacX's.

WeiqunZhang: Since phi is truly 2D, there is no need to rebuild phi for the solver.

WeiqunZhang: Or are there cases phi is actually 3d? If so, rebuild phi using r2d's boxarray is still wrong because it makes it 2d.

src/particles/spacecharge/PoissonSolve.cpp
```diff
@@ -73,16 +79,41 @@ namespace impactx::particles::spacecharge
         pp_algo.queryAddWithParser("mlmg_max_iters", mlmg_max_iters);
         pp_algo.queryAddWithParser("mlmg_verbosity", mlmg_verbosity);
 
+        // flatten rho to 2D
+        std::unordered_map<int, std::pair<amrex::MultiFab, amrex::MultiFab>> rho_2d;  // pair: local & unique boxes
+        if (space_charge == SpaceChargeAlgo::True_2D) {
```
WeiqunZhang, line 87: What's the meaning of True_2D? If it means there is only one cell in z-direction, there is no need to call flatten_charge_to_2D, because it's already 2D.

cemitch99:
For this PR, we need to resolve the following error resulting in failing tests:

> amrex::Abort::1::Particle container needs to have at least one grid. !!!
> SIGABRT
> 
> Grids Summary:
>   Level 0   1 grids  1024 cells  100 % of domain

examples/fodo_space_charge/run_fodo_2d_sc.py
```diff
@@ -0,0 +1,81 @@
...
+
+from impactx import ImpactX, distribution, elements, twiss
+
+sim = ImpactX()
+
+# set numerical parameters and IO control
+sim.max_level = 0
+sim.n_cell = [32, 32, 1]
+sim.blocking_factor_x = [32]
+sim.blocking_factor_y = [32]
```
cemitch99:
```suggestion
sim.blocking_factor_x = [16]
sim.blocking_factor_y = [16]
```

cemitch99:
Fixed.

ax3l:
@WeiqunZhang can you confirm what the last AMReX patch is that we need? Are there any newer than AMReX 25.11?

ax3l:
Slack:
> I don't think that PR needs anything beyond 25.11.

cemitch99:
To further validate this PR, here is a figure illustrating the computed beam moments for the example `input_fodo_2d_sc.in` (modified to use 1 M particles and a [64,64] grid), compared against the result using envelope tracking.  This is a high-current case, with strong asymmetry in X and Y.
FODO_2D_SpaceCharge_Example

cemitch99 (approved):

examples/expanding_beam/README.rst
```diff
@@ -1,7 +1,8 @@
+
 .. _examples-expanding:
 
-Expanding Beam in Free Space
-============================
+Expanding Beam in Free Space with 3D Space Charge
+==================================================
```
ax3l:
```suggestion
Expanding Beam in Free Space with 3D Space Charge
=================================================
```

examples/expanding_beam/README.rst
```diff
@@ -69,3 +70,58 @@ We run the following script to analyze correctness:
    .. literalinclude:: analysis_expanding.py
       :language: python3
       :caption: You can copy this file from ``examples/expanding/analysis_expanding.py``.
+
+
+
+.. _examples-expanding-fft-2d:
+
+Expanding Beam in Free Space with 2D Space Charge
+==================================================
```
ax3l:
```suggestion
Expanding Beam in Free Space with 2D Space Charge
=================================================
```

src/initialization/InitDistribution.cpp
```diff
@@ -531,14 +531,22 @@ namespace impactx
         amrex::ParmParse pp_algo("algo");
         std::string track = "particles";
         pp_algo.queryAdd("track", track);
+        //std::string space_charge = "space_charge";
+        //pp_algo.queryAdd("space_charge", space_charge);
+
```
```suggestion
```

src/initialization/InitDistribution.cpp
```diff
@@ -586,7 +594,7 @@
 
             amrex::ParticleReal intensity = 0.0; // bunch charge (C) for 3D model, beam current (A) for 2D model
 
-            auto space_charge = get_space_charge_algo();
+            //auto space_charge = get_space_charge_algo();
```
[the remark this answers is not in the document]
ax3l:
```suggestion
```

src/initialization/AmrCoreData.cpp
```diff
@@ -145,25 +147,40 @@ namespace impactx::initialization
...
-        int const num_guards_phi = num_guards_rho + 1; // todo: I think this just depends on max(MLMG, force calc)
+        amrex::IntVect num_guards_phi{num_guards_rho + 1}; // todo: I think this just depends on max(MLMG, force calc)
+        amrex::BoxArray phi_ba = cba;
+        if (space_charge == SpaceChargeAlgo::True_2D) {
+            num_guards_phi[2] = 0;
+            amrex::BoxList bl(amrex::IndexType{rho_nodal_flag});
+            bl.reserve(cba.size());
+            for (int i = 0, N = cba.size(); i < N; ++i) {
+                amrex::Box b = cba[i];
+                b.convert(rho_nodal_flag).setRange(2,0); // Flatten box in z-direction
```
ax3l, line 159: I am not particularly proud of the logic here and in PoissonSolve on how and when we allocate 2D fields. But I am ok to go ahead for now and clean it up in another go.

Issue #794, opened 2025-01-10
Allow for "reverse" back-tracking


cemitch99:
Allow users to track particles backward with respect to the independent variable `s`.  This requires verifying that the tracking methods implemented are fully reversible.  Ideally, this should include space charge as well.

This is useful, for example, to determine an ideal "initial" distribution needed to obtain a target "final distribution" downstream.

It can also be used as a method of chaos detection for studying dynamic aperture.

PR #920, opened 2025-03-26
Implement Strang splitting with space charge

### LINKED ISSUE #846: Element Slicing: Strang Splitting
@RemiLehe suggested today we could use a weighted approach to `ds` slices (Strang Splitting) when we slice our elements (for collective effects), to increase the numerical order by one.

I think we could even use a generator for the slices we track in an element, to implement various splitting methods.

cc @cemitch99 

Example:
https://github.com/desy-ml/cheetah/blob/v0.7.0/tests/test_space_charge_kick.py#L46-L64

cemitch99:
This PR modifies the main tracking loop with collective effects to use a second-order Strang splitting.
- increases the accuracy of the space charge calculation by moving from first-order to second-order with respect to the slice step size, currently set by `nslice`
- ensures that integration in `s` is time-symmetric, which is needed for reversible back-tracking

Close #846 
- [x] simple implementation for the case that nslice is an even integer
- [x] implement and test in the envelope tracking loop
- [ ] implement and test in the particle tracking loop
- [ ] modify implementation for the case of general nslice
- [ ] add benchmark test(s)

A follow-up PR will address Issue #794

### DIFF
--- a/examples/expanding_beam/analysis_expanding_envelope.py
+++ b/examples/expanding_beam/analysis_expanding_envelope.py
@@ -104,9 +104,9 @@ print(f"  rtol={rtol} (ignored: atol~={atol})")
 assert np.allclose(
     [sig_xf, sig_yf, sig_tf],
     [
-        9.029112e-04,
-        9.029112e-04,
-        1.841402e-06,
+        8.94427191e-04,
+        8.94427191e-04,
+        1.824483738e-06,
     ],
     rtol=rtol,
     atol=atol,
--- a/examples/kurth/analysis_kurth_10nC_periodic_envelope.py
+++ b/examples/kurth/analysis_kurth_10nC_periodic_envelope.py
@@ -98,7 +98,7 @@ print(
 )
 
 atol = 0.0  # ignored
-rtol = 2.0e-3  # from random sampling of a smooth distribution
+rtol = 3.5e-3  # from random sampling of a smooth distribution
 print(f"  rtol={rtol} (ignored: atol~={atol})")
 
 assert np.allclose(
--- a/src/tracking/envelope.cpp
+++ b/src/tracking/envelope.cpp
@@ -137,7 +137,8 @@ namespace impactx
                 }, element_variant);
 
                 // sub-steps for space charge within the element
-                for (int slice_step = 0; slice_step < nslice; ++slice_step)
+                int nsteps = std::floor(nslice/2.0);
+                for (int slice_step = 0; slice_step < nsteps; ++slice_step)
                 {
                     BL_PROFILE("ImpactX::track_envelope::slice_step");
                     step++;
@@ -147,14 +148,27 @@ namespace impactx
                                        << " slice_step=" << slice_step << "\n";
                     }
 
+                    std::visit([&ref, &cm](auto&& element)
+                    {
+                        // push reference particle in global coordinates
+                        {
+                            BL_PROFILE("impactx::Push::RefPart");
+                            element(ref);
+                        }  
+                    
+                        // push Covariance Matrix in external fields
+                        element(cm, ref);
+                                       
+                    }, element_variant);
+
                     if (space_charge == SpaceChargeAlgo::True_2D)
                     {
                         // push Covariance Matrix in 2D space charge fields
-                        envelope::spacecharge::space_charge2D_push(ref, cm, intensity, slice_ds);
+                        envelope::spacecharge::space_charge2D_push(ref, cm, intensity, 2*slice_ds);
                     } else if (space_charge == SpaceChargeAlgo::True_3D)
                     {
                         // push Covariance Matrix in 3D space charge fields
-                        envelope::spacecharge::space_charge3D_push(ref, cm, intensity, slice_ds);
+                        envelope::spacecharge::space_charge3D_push(ref, cm, intensity, 2*slice_ds);
                     } else {
                         amrex::Print() << "Warning: Space charge is off by default." << "\n";
                     }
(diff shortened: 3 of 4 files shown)


examples/expanding_beam/analysis_expanding_envelope.py
```diff
@@ -104,9 +104,9 @@ def read_time_series(file_pattern):
 assert np.allclose(
     [sig_xf, sig_yf, sig_tf],
     [
-        9.029112e-04,
-        9.029112e-04,
-        1.841402e-06,
+        8.94427191e-04,
```
cemitch99, line 107: Here the targets were tuned to analytically-determined values, instead of being set to coincide with a low-resolution run.

examples/kurth/analysis_kurth_10nC_periodic_envelope.py
```diff
@@ -98,7 +98,7 @@ def read_time_series(file_pattern):
 )
 
 atol = 0.0  # ignored
-rtol = 2.0e-3  # from random sampling of a smooth distribution
+rtol = 3.5e-3  # from random sampling of a smooth distribution
 print(f"  rtol={rtol} (ignored: atol~={atol})")
```
cemitch99, line 102: The original tolerance was too tight.

src/tracking/envelope.cpp
```diff
@@ -137,7 +137,8 @@ namespace impactx
                 }, element_variant);
 
                 // sub-steps for space charge within the element
-                for (int slice_step = 0; slice_step < nslice; ++slice_step)
+                int nsteps = std::floor(nslice/2.0);
```
cemitch99:
```suggestion
                int nsteps = std::round(nslice/2.0);
```

cemitch99:
The attached result illustrates the improvement in numerical convergence obtained when running the expanding beam [test](https://impactx.readthedocs.io/en/latest/usage/examples/expanding_beam/README.html) with envelope tracking (`input_expanding_envelope.in`).

[ImpactX_StrangSplitting.pdf](https://github.com/user-attachments/files/19511092/ImpactX_StrangSplitting.pdf)

src/tracking/envelope.cpp
```diff
@@ -137,7 +137,8 @@ namespace impactx
                 }, element_variant);
 
                 // sub-steps for space charge within the element
-                for (int slice_step = 0; slice_step < nslice; ++slice_step)
+                int nsteps = std::round(nslice/2.0);
+                for (int slice_step = 0; slice_step < nsteps; ++slice_step)
```
ax3l, line 141: I think we could even do this: replace the for loop with an iterator ("generator") of slices.

The generator we can implement similar to different integrators in separate files and that way we can provide multiple strang splitting options.

The code here would then read something like:
```c++
auto s_slicer = StrangSplitter(nslice, order, nsteps);
for (auto & slice : s_slicer)
{
    int slice_step = slice.slice_step;
    amrex::Real slice_ds = slice.slice_ds;

    // ...
}
```

That way, no strang splitting constants/math need to be calculated here and we can reuse this in envelope and particle tracking.

cemitch99: Good idea.  Let's touch base on the details when convenient.

src/tracking/envelope.cpp
```diff
@@ -147,14 +148,27 @@ namespace impactx
                                        << " slice_step=" << slice_step << "\n";
                     }
 
+                    std::visit([&ref, &cm](auto&& element)
```
ax3l, line 151: Is the reference particle pushed twice now? o.0

Happens again in line 176 below

ax3l: I think this is responsible for the failing tests

cemitch99: Clarified in person:
- also includes envelope push
- note the difference in weights (half step for push, full step for SC & CSR, another half step for push)

cemitch99: We have to make sure that thin elements, which only go through this loop body once, are still only kicked once though.

src/tracking/envelope.cpp
ax3l: Discussed today: does not work so easily, because  for higher-order schemes we also need to split the weights and number of calls to pushes inside the slice loop body. Will revert the commit that added the strang splitter.

ax3l: Removed my change from the branch.

ax3l:
A small comment: this PR attempted a map/2 + collective kick + map/2 strang split, because intuitively this keeps the more costly collective kick centered and avoids doing +1 kick per element on its slices.

But, this only holds for `Thick`-derived, constant-field, non-s-dependent elements as a performance optimization. For thick elements that have s-dependent fields (RFCavity, SoftQuad, SoftSol), for potentially-thick but not easily splittable elements (LinearMap/SpinMap/Programmable) and for all Thin elements, a Kick/2 + map + kick/2 is the self-consistent fallback to use.

Under that viewpoint, I personally would recommend to do optimizations for number of collective kicks done later on: it complicates the code for a subset of the elements, and alternatives to optimize, like merging the half-kick at the last nslice with the next half-kick of the next element could be explored too (both optimizations are a bit messy in book-keeping).

Even with kick/2 + map + kick/2: in practice, because of the lower nslice needed now it might not even always pay off to carve out that one extra collective kick per element... (to be measured). Of course, for low numbers of nslice the cost is big, a regime we can now enter more with strang splitting...

cemitch99:
Hm; I don't understand your last comment.  The idea, as it is implemented in IMPACT-Z, is to use map/2 + collective kick + map/2 for _all thick_ elements.  If there is an s-dependence in the field, then the calculation of map is applied by numerical integration via splitting into substeps, whose number is set by the `mapsteps` parameter.

PR #921, opened 2025-03-27
Dashboard: Variable Referencing in Lattice Configuration 

proy30:
Supersedes #829 due to changes in a different branch.

--------------
This PR allows dashboard users to define variable names, assign numeric values, and reference these variables in lattice configurations.

- This PR validates a variable name by using Python’s built-in isidentifier() method and the keyword library to ensure it starts correctly, contains only allowed characters, isn’t a reserved word, and is unique among all variables.

### Clarification on `ui_input` vs `sim_input` behavior introduced in this PR

- If the user inputs `x` in `parameter1`, and `x` **is defined** as a variable with value `50`:
  - `ui_input = 'x'`
  - `sim_input = 50`

- If the user inputs `x` in `parameter1`, and `x` **is not defined**:
  - `ui_input = 'x'`
  - `sim_input = 'x'` *(not `''`)*  
    The simulation is disabled in this case if the `param_type` is not `str`.

- If the user inputs `7` directly in `parameter1`:
  - `ui_input = 7`
  - `sim_input = 7`


proy30:
https://github.com/user-attachments/assets/c48bb282-2c6e-4ef9-b983-cf13aa61d045

cemitch99 (approved):
It looks good to me.  Thanks a lot.

src/python/impactx/dashboard/Input/latticeConfiguration/latticeMain.py
```diff
@@ -170,17 +154,74 @@ def on_add_lattice_element_click():
...
+    :param ui_value: The value present on the UI end..
+    :param parameter_type: The lattice element parameters type.
+    """
+
+    lattice_variable, variable_index = (
+        LatticeVariableHandler.determine_if_existing_variable(ui_value)
+    )
+    potentially_lattice_variable = LatticeVariableHandler.is_valid_variable_name(
+        ui_value
+    )
```
EZoni: @proy30 

I'm starting to review this PR, I will add single comments here and there as I go through it.

My understanding, based on the implementation of `is_valid_variable_name`, is that `ui_value` here is a `str` that represents the name of the "helper" variable used to encode the value of a parameter that ImpactX and/or the dashboard know about. Is that right? If this is the case, I wonder if `value` is the right name for this variable. If the "helper" variable is called `n` and its value is set to `10` (and, say, used to set the value of several parameters that ImpactX and/or the dashboard know about), then what does `ui_value` represent in this case, the name `n` or the value `10`?

src/python/impactx/dashboard/Input/generalFunctions.py
EZoni, line 1: Looking generally at the function `update_simulation_validation_status`, is `error_details`, which in this PR you append with what is returned by `has_error_in_variables`, used elsewhere other than for disabling the simulation button? If not, would it make sense to maybe print `error_details` to standard output or something similar? Or did you see this more like an internal debugging tool for developers only?

src/python/impactx/dashboard/Input/latticeConfiguration/variable_handler.py
```diff
@@ -0,0 +1,270 @@
...
+        """
+        Adds a new variable to the dashboard's variable
+        with empty values and updates UI.
+        Stored in a state which contains a list with dictionaries.
+        """
+
+        new_variable = {key: "" for key in state.variables[0]}
+        state.variables.append(new_variable)
+        state.dirty("variables")
+        LatticeVariableHandler.update_delete_availability()
```
EZoni: Could this manual call be removed if `update_delete_availability` is decorated with `@state.change("variables")`?
```suggestion
```

src/python/impactx/dashboard/Input/latticeConfiguration/variable_handler.py
```diff
@@ -0,0 +1,270 @@
...
+        """
+        Deletes the variable defined by the user
+        provided the index
+
+        :param index: The index of the variable
+        """
+
+        state.variables.pop(index)
+        state.dirty("variables")
+        LatticeVariableHandler.update_delete_availability()
```
EZoni: Could this manual call be removed if `update_delete_availability` is decorated with `@state.change("variables")`?
```suggestion
```

src/python/impactx/dashboard/Input/latticeConfiguration/variable_handler.py
```diff
@@ -0,0 +1,270 @@
...
+    def on_reset_variables() -> None:
+        """
+        Resets the dashboard's variables to default.
+        """
+
+        state.variables = [
+            {"name": init_value, "value": init_value, "error_message": init_value}
+        ]
+        state.dirty("variables")
+        LatticeVariableHandler.update_delete_availability()
```
EZoni: Could this manual call be removed if `update_delete_availability` is decorated with `@state.change("variables")`?
```suggestion
```

src/python/impactx/dashboard/Input/latticeConfiguration/latticeMain.py
proy30: In your example, `ui_value` would be `n` and `sim_value` would be `10`.

That said, `ui_value` isn’t always a `str` (variable name). If a user directly enters `7` into a text field, then `ui_value` would be `7`, since that’s what’s shown in the UI.

So, whatever is visible to or entered by the user — whether it’s a variable name or a number — ends up as the `ui_value`. The `sim_value` is the actual value used in the simulation (either the variable’s value or the raw input).

As for naming, I’m open to suggestions. I went with `ui_value` since it can hold both a number or a string, but maybe something like `ui_input` could be clearer?

src/python/impactx/dashboard/Input/latticeConfiguration/variable_handler.py
proy30: I originally had this handled manually instead of using `@state.change("variables")` because that decorator triggers on **any** change to the `variables` list — including edits to names or values — not just when the number of variables changes.

I think the performance impact is minimal, so I’m not opposed to switching to the reactive approach for simplicity.

src/python/impactx/dashboard/Input/generalFunctions.py
proy30: The error details aren’t shown directly to the user on the dashboard since users already see error messages on the specific input fields.

We could refactor this to use a simple true/false flag instead, but I originally kept the detailed messages for debugging during the early stages of the dashboard and just never updated it after that.

I do know that `update_simulation_validation_status` is messy and have plans to refactor it to reduce the amount of manual print statements to `error_details`.

src/python/impactx/dashboard/Input/latticeConfiguration/variable_handler.py
EZoni: Understood, makes sense. Yes, I was thinking that moving the availability control to a function that listens to changes in `variables` (i.e., all changes), as you did in your last commit, may make things more robust (thinking of someone other than you who may make changes to the code in the future, without necessarily mastering the whole infrastructure in detail), as it reduces the need for ad-hoc manual calls in exchange for the overhead that you mentioned.

src/python/impactx/dashboard/Input/generalFunctions.py
EZoni: Understood, thanks. Sounds like work that could be done in future PRs, without urgency.

src/python/impactx/dashboard/Input/latticeConfiguration/latticeMain.py
EZoni: Understood, thanks. Just to clarify, before I possibly make naming suggestions, `ui_value` can be either a `str` or the numerical value of the quantity represented by the `str`, while `sim_value` is always a numerical value? Or can you have a situation where both are `str` (maybe what you referred to as "raw input" in your comment above)?

proy30: I was about to say that `sim_value` is always a numerical value — but after checking that’s not true. A user could type something like `x` into a parameter, even though `x` hasn’t been defined as a variable yet. In that case, `sim_value` would be `x`, a `str`.

Right now, the dashboard handles this by throwing an error message and disabling the simulation. We could also make it so that if `sim_value` is a `str`, we just set it to an empty string `""`.

**I believe I actually tried this approach in the beginning of this PR,** but something didn’t work as expected — I think some reactivity issue --  so I left it as-is, especially since we already block the sim when a non-numerical input is given where a number is expected.

With that said, `sim_value` might need a variable rename then.

EZoni: Okay, I think throwing an error and disabling the simulation is fine, especially if this is already the general behavior anyways. I'm thinking that maybe your suggestion of using `input` instead of `value` may be a good one, like `ui_input` and `sim_input`. What do you think?

proy30: Yes, I agree that is a better naming.

Also for clarification:

If user inputs `x` in `parameter1` and `x` is defined as a variable with value `50`:
`ui_value` = `'x'`
`sim_value` = `50`

If user inputs `x` in `parameter1` and `x` is not defined as a variable:
`ui_value` = `'x'`
`sim_value` = `'x'` instead of `sim_value` = `''`

If user directly inputs `7`  in `parameter1`:
`ui_value` = `7`
`sim_value` = `7`

EZoni: Perfect, would you maybe add this example that you just made to the PR description as well, so it's a little more visible if one looks back at the PR in the future? Thanks!

And/or add something similar as inline comment in the code as well.

EZoni (approved):

PR #926, opened 2025-04-02
Add an exact nonlinear quadrupole

cemitch99:
This PR implements a quadrupole model called `ExactQuad` that is based on the exact nonlinear Hamiltonian of an ideal hard-edge quadrupole.  It is needed for beams with a large energy spread and transverse divergence, and correctly handles relativistic kinematic effects.  It is analogous to the Elegant element `KQUAD`.  Symplectic integration is used to push between space charge kicks.

- [x] add basic element structure, parameters and user inputs
- [x] add support for symplectic integration of particle 6-vector
- [x] implement push for map1 (identical to a linear quadrupole)
- [x] implement push for map2 (the nonlinear kinematic correction terms)
- [x] add user option for integration order (2 / 4)
- [x] add Python bindings
- [x] add benchmark examples:  FODO with exact nonlinear elements, quad with H invariance
- [x] update documentation
- [x] add citation(s)

Follow-up:  a general combined-function multipole addressing Issue #929 .

### DIFF
--- a/docs/source/usage/examples.rst
+++ b/docs/source/usage/examples.rst
@@ -38,6 +38,7 @@ Single Particle Dynamics
    examples/linear_map/README.rst
    examples/scraping_beam/README.rst
    examples/reversibility/README.rst
+   examples/symplectic_integration/README.rst
 
 
 Collective Effects
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -498,7 +498,10 @@ This requires these additional parameters:
 ``quad_exact`` for a Quadrupole magnet using the exact relativistic Hamiltonian, including all kinematic nonlinearities.
 Particle tracking is performed using symplectic integration based on the Hamiltonian splitting :math:`H = H_1 + H_2`.
 Here :math:`H_1` is the Hamiltonian for a linear quadrupole (containing all terms quadratic in the phase space variables),
-and :math:`H_2` is the remainder (including the kinematic square root).  This suggested splitting is based on:
+and :math:`H_2` is the remainder (including the kinematic square root).  This suggested splitting appears for example in:
+
+* D. L. Bruhwiler et al, in Proc. of EPAC 98, pp. 1171-1173 (1998).
+* E. Forest, J. Phys. A: Math. Gen. 39, 5321 (2006).
 
 This requires these additional parameters:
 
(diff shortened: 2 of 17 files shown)


nikitakuklev:
Noticed this effort while starting my own implementation. It would be useful to copy some features of Elegant/Xsuite quad elements. Notably, H/V/K1S are good for combined function magnets.

An alternative could be making generic 'exact' multipole element (later).

Reference code:
* elegant [multipole_tracking2](https://svn.aps.anl.gov/AOP/oag/trunk/apps/src/elegant/src/multipole.c) 
* xsuite generic [multipole tracker](https://github.com/xsuite/xtrack/blob/b30be4f6aa327fd75767b732c7ce9e5cb73caaf6/xtrack/beam_elements/elements_src/track_magnet.h#L139)

cemitch99:
Thanks @nikitakuklev.  Here is my current thinking: For an ideal quadrupole, a skew gradient `K1s` can already be supported by specifying a rotation.  By `H`/`V`, do you refer to HKICK and VKICK in the Elegant KQUAD?  I need to investigate that functionality to better understand the use case.  Support for fringe field models is a larger effort that will need to be added later.

I planned to follow up with an `ExactMultipole` or `ThickMultipole` element, based on a similar splitting, which would support an array of multipoles (KNn, KNs), and an `ExactCFBend`, with a similar functionality for curved coordinates--a generalization of the existing linear `CFBend`.

Any additional comments about which features of the elegant/xsuite models are of highest interest to you are much appreciated!

nikitakuklev:
Yes, H/V corresponds to dipole kicks. One problem with tilting to get skew quad is that H/V are not H/V anymore (in the interpretation that tilt should change reference trajectory). It is a minor problem though since multipole will be available. 

It makes sense to have a general 'multipole_body_track' method that `ExactQuad` calls into if anything beyond K1 is defined, and that `ExactMultipole` always calls. Then, `ExactQuad` can have a special path for K1 only for performance reasons. 

We want to achieve parity between Xsuite/ImpactX for all basic types for building a simulation pipeline (see https://indico.cern.ch/event/1382428/contributions/6284332/). I will open an issue next week with a table of elements, to discuss where contributions would be welcome.

cemitch99:
Thanks @nikitakuklev.  Regarding `ExactQuad`, from a development perspective I find it helpful to add features with increasing generality, even if these are superseded later.  This makes basic models available as-needed, and it is helpful for benchmarking more complex models as they develop.  Having a table will be helpful and much appreciated.

PR #930, opened 2025-04-05
Dashboard: Improve Variable Handling Logic

proy30:
With the addition of [PR #921](https://github.com/BLAST-ImpactX/impactx/pull/921), the dashboard’s import parser must be updated to support loading variables from user-uploaded `.py` input files.

Only numerical (`int` and `float`) variable references are supported (ie. no math expression allowed or a reference to a function). 
Example: The dashboard will load in `x = 50` and `x = 3.14` but not `x = '50'` or `x=5 + 5` or `x=some_function()`

TODO:
- [x] Load only the variables references in the lattice configuration section
    - Since we give the option of only utilizing the variable handler in the lattice configuration section.
    - A variable defined in the imported `.py` file but not used specifically in the lattice configuration build is not loaded in.
- [x] Allow users to negate variable values
     - For example, if `ns = 50` allowing `-ns` would mean `-50`. Right now the dashboard does not support it.

**Changes after review (not necessarily all related to the reviewed comments)**
- [x] Fix issue where an error occurred when a `variable name` did not have a corresponding value 
    - In this case, we set the `variable value` to `None`
- [x] Convert the `variable value` immediately to a numeric value
    - The input was read as a `string` beforehand.
- [x] Update docstrings
- [x] Shorten `is_valid_name_for_user_input` to `is_valid_input_name`
- [x] Rename `lattice_params_bound_or_pending_variable` to `lattice_elements_using_variables`
     - The new name _can_ be misleading, but I think this is the closest it gets to condensing the meaning. Open to suggestions.
- [x] Add docstring to `parse_variables`

Merge after #971


src/python/impactx/dashboard/Input/latticeConfiguration/latticeMain.py
```diff
@@ -150,16 +150,27 @@ def process_if_variable(index, parameter_name, ui_input, parameter_type):
...
     )
 
     if lattice_variable or potentially_lattice_variable:
         if lattice_variable and variable_index is not None:
-            sim_input = state.variables[variable_index]["value"]
+            sim_value = state.variables[variable_index]["value"]
+            if is_negative:
+                sim_input = -float(sim_value)
+            else:
+                sim_input = sim_value
```
EZoni: Is this intentionally casting to `float`? For example, why would we be casting a negative `int` to a negative `float` in this context? 

It may be because state variables need to be serializable and you get runtime errors if you don't do the casting when the original variable is indeed a real (i.e., not an integer) - is that it?

src/python/impactx/dashboard/Input/latticeConfiguration/variable_handler.py
```diff
@@ -26,13 +26,14 @@ class LatticeVariableHandler:
...
-            )
-        LatticeVariableHandler.update_delete_availability()
+            index = lattice_element["index"]
+            if index < len(state.selected_lattice_list):
+                ctrl.updateLatticeElementParameters(
+                    lattice_element["index"],
+                    lattice_element["parameter_name"],
+                    lattice_element["ui_input"],
+                    lattice_element["parameter_type"],
+                )
```
EZoni: Can you briefly explain what `index` represents here? Maybe we can add a brief example as inline comment right above this block, to be decided.

src/python/impactx/dashboard/Toolbar/importParserHelper.py
```diff
@@ -153,7 +154,27 @@ def parse_lattice_elements(content: str) -> dict:
...
             dictionary["lattice_elements"].append(element)
 
+        dictionary["used_lattice_variables"] = used_variables
         return dictionary
+
+    @staticmethod
+    def parse_variables(content: str, used_vars: set) -> dict:
+        variables = {}
+        variables_regex = r"^\s*(\w+)\s*=\s*([^#\n]+)"
+        matches = re.findall(variables_regex, content, re.MULTILINE)
```
EZoni: Could you add an inline comment with an example of what string this would typically parse (maybe even with a brief explanation of the more undecipherable parts of `variables_regex` string - to be decided)?

EZoni (commented):
Thanks for the PR, @proy30. I left only a few comments.

src/python/impactx/dashboard/Toolbar/importParserHelper.py
proy30: Added docstring commit

src/python/impactx/dashboard/Input/latticeConfiguration/latticeMain.py
proy30: Thank you for pointing this out. It was initially done as a quick fix to convert `sim_value` to a numeric input, since it was provided as a string with its text field input. I added a comment to the description to now merge this PR after https://github.com/BLAST-ImpactX/impactx/pull/971 so that we can very simply convert the input to a numeric value.

Some additional commits were also made to simplify the readability.

src/python/impactx/dashboard/Input/latticeConfiguration/variable_handler.py
proy30: Thank you for pointing this out as well. I realized that the implementation of this was not quite right.

I added a commit to fix my original intention.

This is the comment I left in the commit:

`"Prevents cases similar to the following: Say you add three lattice elements, the first two reference a variable named ns, the third one references a variable named ns2 in the same parameter name. If we end up deleting the second lattice element, and update the value of the ns variable, the ns2 originally from index 2 becomes ns"`

I hope this makes sense.

EZoni:
@proy30 

I enabled auto-merge on #971, feel free to rebase here (if needed) once #971 is merged.

EZoni (approved):

PR #944, opened 2025-04-20
Dashboard: Correct import path for CLI entry point

proy30:
I noticed that the dashboard would not load properly when using the CLI entry point (`impactx-dashboard`) from the base directory. It failed with the following error:

```
ModuleNotFoundError: No module named 'distribution_input_helpers'
```

After testing in GitHub Codespaces, I found that adding the `impactx.` prefix resolved the issue on my end:

```python
from impactx.distribution_input_helpers import twiss
```

This change would allow the dashboard to run correctly when launched from the command line.


cemitch99 (approved):

cemitch99:
I tried this, observed the problem, and this provides a simple fix.  I will merge this right away.

PR #945, opened 2025-04-21
Change the default 3D Poisson solver to FFT.

cemitch99:
The 3D FFT-based Poisson solver based on an Integrated Green Function approach (with open boundary conditions) is robust and efficient.  Following a prior discussion with ax3l, this PR updates this Poisson solver to serve as default, with the MLMG solver as a user option.


proy30 (approved):

cemitch99:
Tests that use the MLMG solver implicitly (by default) now must be made explicit.  Otherwise, these tests use the FFT solver and fail because the large domain size results in poor grid resolution.

EZoni (approved):

PR #943, opened 2025-04-19
Dashboard: Add inputs for `tracking modes` and space charge dropdown for `2D/3D`

proy30:
Add new `tracking_mode` dropdown with following options:
- Particle Tracking (**default**) (user can also select `space charge` or `csr`)
- Envelope Tracking (user can select `space charge` while `csr` is disabled)
- Reference Tracking (both `space charge` and `csr` are disabled)

Additionally, the selection for space charge has changed from a checkbox to a dropdown, with the options depending on the users selection of `tracking mode`.:
- In particle tracking, the options for `space charge` are `off`, `3D`
- In envelope tracking, the options for `space charge` are `off`, `2D`, `3D`
- In reference tracking, `space charge` defaults to `false` and is set to `readonly`

Demo:
https://github.com/user-attachments/assets/8ec7f597-4dfe-44ab-b2c4-2e9e47e750c0

PR #949, opened 2025-04-22
Dashboard: Fix errors

proy30:
## Summary of Fixes and Changes

1. **Unconditional Space Charge Execution**  
   The issue originates from recently merged [#943](https://github.com/BLAST-ImpactX/impactx/pull/943), where simulations are running with space charge inputs **regardless of whether** `space_charge` is set to `false`.  
   - Issue is fixed by ensuring space charge inputs are only applied when explicitly enabled.

2. **Fix: `name` Parameter Type in Lattice Elements**  
   The `'name'` parameter always triggered a validation error due to its default type being `Optional[str]` and not `str`.  
   - This is solved by stripping the `Optional[...]` wrapper when parsing parameter types.
   - Dashboard now also properly validates the parameter by utilizing the existing `is_valid_variable_name` function. Renamed to `is_valid_name_for_user_input`.
3. **Update Default `blocking_factor` Value**  
   According to the [AMReX documentation](https://amrex-codes.github.io/amrex/docs_html/GridCreation.html) and [WarpX documentation](https://warpx.readthedocs.io/en/21.10/usage/workflows/parallelization.html), the default `blocking_factor` should be **8**, not 16. 
   - `Inputs/defaults.py` is updated to reflect the correct default.

4. **Fix: `blocking_factor` Input Format for Space Charge**  
   The dashboard currently sets `blocking_factor` as a single integer, which breaks space charge runs.  
   Per the documentation, it must be a **list of integers**.  
   ```python
   def blocking_factor_x(self, arg1: list[int]) -> None

5. **Fix: `isr` inputs**
    - Currently the documentation is missing for `isr`. Additionally, we add a check in the code to alert if a documentation is missing.
    - todo Additionally, `isr` is displaying for all three `tracking_mode`s. Add a check to disable on certain modes.


cemitch99 (approved):
Looks good.

PR #950, opened 2025-04-22
Dashboard: Add tooltips for Twiss distribution parameters

proy30:
The PR adds the tooltips for the twiss distribution parameters on the dashboard.

This is done by adding a helper function `get_tooltips_from_param`  in `Input/defaults_helper.py` which looks at a given function (in this case, `twiss` from`src\python\impactx\distribution_input_helpers.py`) and properly parses through the docstring by looking at each `:param` and its description to retrieve the tooltip.

### DIFF
--- a/src/python/impactx/dashboard/Input/defaults_helper.py
+++ b/src/python/impactx/dashboard/Input/defaults_helper.py
@@ -51,24 +51,36 @@ class InputDefaultsHelper:
         return docstrings
 
     @staticmethod
-    def get_tooltips_from_param(func: Callable) -> Dict[str, str]:
+    def get_tooltips_from_param(function: Callable) -> Dict[str, str]:
         """
         Extract all ':param name: description' entries from a function's docstring.
 
-        :param func: The function whose docstring you want to parse.
+        Example:
+            :param beta_x: Beta function value in the x dimension.
+            :param emitt_x: Emittance function value in the x dimension.
+
+        This will produce:
+            {
+                "beta_x": "Beta function value in the x dimension.",
+                "emitt_x": "Emittance function value in the x dimension."
+            }
+
+        :param function: The function whose docstring you want to parse.
         :return: A dict mapping each parameter name to its description.
         """
-        tooltips = {}
-        doc = inspect.getdoc(func) or ""
+        tooltip_results = {}
+        docstring = inspect.getdoc(function) or ""
         pattern = re.compile(r"^\s*:param\s+(\w+)\s*:\s*(.+)$", re.MULTILINE)
-        pattern_matches = list(pattern.finditer(doc))
+        pattern_matches = list(pattern.finditer(docstring))
 
         if not pattern_matches:
-            raise ValueError(f"Found no docstrings to parse in function {func.__name__}")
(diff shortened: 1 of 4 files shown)


src/python/impactx/dashboard/Input/distributionParameters/distributionMain.py
```diff
@@ -177,19 +177,27 @@ def card_content(self):
...
-                            type="number",
-                            step=("parameter.parameter_step",),
-                            __properties=["step"],
-                            density="compact",
-                            variant="underlined",
-                            hide_details="auto",
-                        )
+                        with vuetify.VTooltip(
+                            location="top",
+                            text=("all_tooltips[parameter.parameter_name]",),
```
EZoni, line 182: I'm surprised that this works and you don't need 
```py
text=(f"all_tooltips[{parameter.parameter_name}]",),
```
instead. What am I missing?

src/python/impactx/dashboard/Input/defaults_helper.py
```diff
@@ -42,4 +45,26 @@ def get_docstrings(
...
+        Extract all ':param name: description' entries from a function's docstring.
+
+        :param func: The function whose docstring you want to parse.
+        :return: A dict mapping each parameter name to its description.
+        """
+        tooltips = {}
+        doc = inspect.getdoc(func) or ""
+        pattern = re.compile(r"^\s*:param\s+(\w+)\s*:\s*(.+)$", re.MULTILINE)
+
+        for match in pattern.finditer(doc):
```
EZoni: Do we maybe need to `try` this and raise an exception to avoid trying to call the method `finditer` from a `pattern` that may not have that method, e.g., in case the `re.compile` call that returned `pattern` did not run successfully (for some reason that we may or may not be able to anticipate)?

src/python/impactx/dashboard/Input/defaults_helper.py
```diff
@@ -42,4 +45,26 @@ def get_docstrings(
                     docstring = inspect.getdoc(attribute) or ""
                     docstrings[name] = docstring
 
+        distribution_tooltips = InputDefaultsHelper.get_tooltips_from_param(twiss)
```
EZoni, line 48: Just double checking, this is intentionally parsing the Twiss docstring as the default docstring to be parsed?

cemitch99 (approved):
Thanks a lot.

src/python/impactx/dashboard/Input/defaults_helper.py
proy30: For now, yes since Twiss is our default set of parameters. A future PR needs to add logic to retrieve the tooltips for when non-twiss parameters are selected, and in that PR this line of code possibly has to change.

src/python/impactx/dashboard/Input/defaults_helper.py
proy30: This is a good point. Right now we just assume the function we pass has the proper formatted docstring. We should add  a check to verify if that is not the case.

EZoni (approved):

PR #924, opened 2025-03-28
Incoherent Synchrotron Radiation in bending dipoles

cemitch99:
This PR adds support for ISR within bending dipoles, using a model based on these two sources:

- J. M. Jowett, "Introductory Statistical Mechanics for Electron Storage Rings," SLAC-PUB-4033 (1986).
- F. Niel et al, "From quantum to classical modeling of radiation reaction:  A focus on stochasticity effects," Phys. Rev. E 97, 043209 (2018).

To do:
- [x] draft of basic math following the IMPACT-Z/Jowett implementation (for testing)
- [x] add logic to handle ISR kicks within the main tracking loop
- [x] add support for user input:  `algo.isr = true` or `false`
- [x] update RNG for support on GPUs
- [x] add Taylor series for quantum corrections to the ISR kick coefficients
- [x] add Python bindings
- [x] add benchmarking test(s)
- [x] update documentation

Follow-ups:  
Reproduce Figs. 9-10 of Niel et al (2018) to benchmark higher-order quantum effects.
Change internal naming of particles/wakefields and related files, Issue #933

PR #951, opened 2025-04-22
Dashboard: Add ISR to inputs

proy30:
Needed after the merge of ISR in https://github.com/BLAST-ImpactX/impactx/pull/924
image


cemitch99:
This looks great, thanks!  Does the ISR panel appear/disappear as ISR is toggled on/off, in the same was as the CSR panel?

proy30:
@cemitch99 Yes! The ISR inputs will only show when toggled on.

cemitch99 (approved):

PR #954, opened 2025-04-24
Dashboard: Record failed sims

proy30:
This PR introduces initial support for tracking failed simulations in the dashboard.

If a simulation ends with a Python `Traceback` error, the dashboard now updates the progress status to indicate a failure. This failure is also logged in the Simulation History view.

Previously, when a simulation failed, users were blocked from launching a new one unless they restarted the dashboard.

### DIFF
--- a/src/python/impactx/dashboard/Run/utils.py
+++ b/src/python/impactx/dashboard/Run/utils.py
@@ -52,6 +52,20 @@ class SimulationHelper:
         SimulationHistory.save_view_details_log()
 
     @staticmethod
+    def fail_simulation() -> None:
+        """
+        Updates the UI and simulation history to reflect a failed simulation.
+        """
+        state.sim_is_running = False
+        state.sim_progress_status = "Failed"
+        state.sim_status_color = "error"
+        state.sims[state.sim_index]["status"] = "Failed"
+        state.dirty("filtered_sims")
+        state.flush()
+        ctrl.terminal_print("Simulation failed due to the above error.")
+        SimulationHistory.save_view_details_log()
+
+    @staticmethod
     def reset():
         state.sim_is_running = True
         state.sim_progress = 0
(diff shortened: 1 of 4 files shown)


proy30:
View of seeing a failed sim in the `Run` tab
chrome_nG8kcsEHni
View of seeing a failed sim in the `sim history` panel
chrome_hCDwwM9URD

src/python/impactx/dashboard/Run/utils.py
```diff
@@ -51,6 +51,20 @@ def complete_simulation():
...
+    def fail_simulation() -> None:
+        """
+        Updates the UI and simulation history to reflect a failed simulation.
+        """
+        state.sim_is_running = False
+        state.sim_progress_status = "Failed"
+        state.sim_status_color = "error"
+        state.sims[state.sim_index]["status"] = "Failed"
+        state.dirty("filtered_sims")
+        state.flush()
```
EZoni, line 64: Just for clarity, what happens (or doesn't happen) in this context if you do not call `state.flush()` manually here (but still call `state.dirty`)?

proy30: Without `state.flush()` the UI won’t properly show the "Failed" message or change the progress bar color to error. So it is needed for force the UI to show lines `61` and `62`. 

This is how it is without the `state.flush()`:

image
And with the `state.flush()`:

image


`state.dirty("filtered_sims")` is specifically needed to update the status in the **Simulation History panel**. When a sim is running it shows "In Progress", but if it fails, it should switch to "Failed". Without this dirty call, the history panel won’t update. I assume the flush doesn't solve this because it doesn't know that `filtered_sims` changed without the `dirty()` call

cemitch99 (approved):

EZoni (approved):

PR #955, opened 2025-04-27
Dashboard: Add unit test

### LINKED ISSUE #667: [Dashboard] Missing Unit Testing
Can add with selenium library, as mentioned here: https://github.com/Kitware/trame/discussions/411

Create tests for:
- A CSR example
- A space charge example
- FODO example
- Loading files
- Dashboard stess test?

proy30:
This first CI test for the dashboard utilizes inputs from `examples/run_fodo.py`. The `seleniumbase` python library is utilized to simulate user interaction with the dashboard.

Passes the following workflows:
- [X] macOS
- [X] Ubuntu
- [X] Stubs

I currently have the dashboard tests run **separately** from the other testing. The initial reason was to isolate the job in the workflows for trial and error.

One note for the code changes is that `run installed python module` in `macOS` and `Ubuntu` is moved to run before both the dashboard and unit testing, rather than after (which was the previous state). This is because the dashboard tests rely on the python bindings for specific imports.

Closes #706 
Closes #667 

Because the `Windows` and `Tooling` workflows do not pass, and possibly is a more involved process to solve, the following issue is created https://github.com/BLAST-ImpactX/impactx/issues/968


.github/workflows/ubuntu.yml
EZoni, line 1: Here I see that the dashboard tests are added as a step in each of the jobs of this workflow. So we run the dashboard tests for the job "GCC w/ MPI w/ Python" and also for the job "GCC w/o MPI w/ Python".

What differences should one expect from interacting with the dashboard in these two different configurations? 

Somehow I would have expected the dashboard tests to be within a new, separate job in this workflow, rather than a new step in each existing job. May be something to discuss in person.

EZoni:
Is there a place in the documentation where we could add guidance on how to run the dashboard tests manually? Including as simple as `ctest --test-dir build --output-on-failure -R dashboard` or similar.

EZoni:
Are all the changes in src/python, as opposed to the changes in tests/python, necessary for the tests to work? In other words, do all those changes belong to this PR or could any of those be implemented in a separate PR, just for clarity?

proy30:
> Are all the changes in src/python, as opposed to the changes in tests/python, necessary for the tests to work? In other words, do all those changes belong to this PR or could any of those be implemented in a separate PR, just for clarity?

@EZoni Yes, the addition of the `id` is needed by selenium to interact with the specific elements. Everything else (the removal of the `@staticmethod` and fixing of the type hint) was caught by the Stubs CI.

> Is there a place in the documentation where we could add guidance on how to run the dashboard tests manually? Including as simple as ctest --test-dir build --output-on-failure -R dashboard or similar.

@EZoni  Yes, on a TODO to add to https://github.com/BLAST-ImpactX/impactx/issues/983

.github/workflows/ubuntu.yml
proy30: Currently, the dashboard does not operate differently in any build, including Ubuntu. The dashboard tests were added to the same jobs that ran our unit tests. As we discussed, it would make sense that future dashboard versions would be different (testing specific inputs for MPI on `GCC w/ MPI w/ Python` only) and could utilize the current implementation of CI. 

Would need input from @ax3l.

tests/python/dashboard/test_dashboard.py
```diff
@@ -0,0 +1,124 @@
...
+    wait_for_interaction_ready,
+    wait_for_server_ready,
+)
+
+
+@pytest.mark.skipif(
+    importlib.util.find_spec("seleniumbase") is None,
+    reason="seleniumbase is not available",
+)
+def test_dashbnoard():
```
EZoni:
```suggestion
def test_dashboard():

```

tests/python/dashboard/utils.py
```diff
@@ -0,0 +1,220 @@
...
+                for _ in range(2):
+                    next_line = process.stdout.readline()
+                    if next_line:
+                        print(next_line, end="")
+                print("Dashboard server has successfully launched.")
+                return
+        else:
+            time.sleep(1)
+
+    raise RuntimeError(f"Dashboard server did not start within {timeout} seconds.")
```
EZoni, line 90: Here I wonder if one could make use of Trame's built-in server control, e.g. `ctrl.on_server_ready` or something like https://trame.readthedocs.io/en/latest/core.server.html#trame_server.core.Server.ready or https://trame.readthedocs.io/en/latest/core.server.html#trame_server.core.Server.get_server_state.

Or, somewhat related, as you check `.trame__loader` above, on server ready `dir(state)` gives
```
['__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__enter__', '__eq__', '__exit__', '__format__', '__ge__', '__getattr__', '__getattribute__', '__getitem__', '__getstate__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__setitem__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_change_callbacks', '_children_state', '_hot_reload', '_modified_keys', '_parent_state', '_pending_update', '_push_state_fn', '_pushed_state', '_ready_flag', '_state_listeners', '_translator', 'change', 'clean', 'client_only', 'dirty', 'flush', 'has', 'initial', 'is_dirty', 'is_dirty_all', 'is_ready', 'modified_keys', 'ready', 'setdefault', 'to_dict', 'translator', 'update']
```
I see `is_ready` and `ready` - I wonder if one can access those through `sb` somehow.

Some version of the above may eventually lead to a simpler implementation than parsing the `stdout`.

EZoni (commented):
A couple more comments!

.github/workflows/ubuntu.yml
```diff
@@ -117,9 +122,20 @@ jobs:
...
+        cmake --build build --target pip_install
+
+        python3 examples/fodo/run_fodo.py
+
+    - name: run dashboard tests
+      run: |
+        cd tests/python/dashboard
+        pytest --confcutdir=.
+
     - name: run tests
```
proy30: @ax3l Moved `run installed python module` to run before testing.

.github/workflows/ubuntu.yml
```diff
@@ -53,9 +53,20 @@ jobs:
           -DMPIEXEC_POSTFLAGS="--use-hwthread-cpus"
         cmake --build build -j 4
 
+    - name: run installed python module
+      run: |
+        cmake --build build --target pip_install
```
ax3l: Double check:
- As of now, do pytest test work even without `pip_install` (yes?)?
- Can we make the dashboard tests work the same way?

ax3l:
@proy30 sorry, small conflict in `tests/python/requirements.txt`

ax3l:
It looks like the subs hang on the new test. Maybe you can exclude the dashboard tests from pytest in stubs to overcome this?

.github/workflows/ubuntu.yml
ax3l: Resolved in last meeting :+1: 

Summary: tests are integrated with pytest (great) and do not run long at the moment (<1min). We exclude them nonetheless already for a few runners where we also exclude "slow" unit tests, because we expect them to grow.

.github/workflows/ubuntu.yml
ax3l: (To be revisited in a follow-up.)

ax3l: I will add this to this PR now.

.github/workflows/ubuntu.yml
ax3l: I improved the CMake now, so we do not need this changed -- restored to previous state :)

tests/python/dashboard/utils.py
ax3l: ping @proy30 did you see this comment?

ax3l (approved):
This is great, thank you so much!
I finalized the PR and will merge it now (please see the commits I added for details).

Please see the open comment by @EZoni on the wait logic. This can be addressed in a follow-up PR if needed.

ax3l:
```
2025-06-22T04:45:02.5764735Z Error copying directory from "/home/runner/work/impactx/impactx/src/python/impactx" to "/home/runner/work/impactx/impactx/build/lib/site-packages/impactx".
```
Uh oh, are we running out of CI disk space? :thinking:

ax3l:
@EZoni we might need to port over our WarpX "clean up each test after it ran and succeeded" logic for ctest to ImpactX.

ax3l:
Got this, pushed the functionality.

Will enable for debug builds in CI for now, because they create the biggest binary artifacts. But can enable on all runners as well later on.

ax3l:
Ah, need to define dependent jobs as in WarpX, too.

examples/CMakeLists.txt
```diff
@@ -681,6 +699,10 @@ add_impactx_test(solenoid.restart
     OFF  # no plot script yet
 )
 set_property(TEST solenoid.restart.run APPEND PROPERTY DEPENDS "solenoid.run")
+if(ImpactX_TEST_CLEANUP)
+    # do not clean up dependency test before current test is completed
+    set_property(TEST solenoid.cleanup APPEND PROPERTY DEPENDS "soilenoid.restart.cleanup")
```
ax3l:
```suggestion
    set_property(TEST solenoid.cleanup APPEND PROPERTY DEPENDS "solenoid.restart.cleanup")
```

examples/CMakeLists.txt
```diff
@@ -691,6 +713,10 @@ add_impactx_test(solenoid.restart.py
 if(ImpactX_PYTHON)
     set_property(TEST solenoid.restart.py.run APPEND PROPERTY DEPENDS "solenoid.py.run")
 endif()
+if(ImpactX_TEST_CLEANUP)
+    # do not clean up dependency test before current test is completed
+    set_property(TEST solenoid.py.cleanup APPEND PROPERTY DEPENDS "soilenoid.restart.py.cleanup")
```
ax3l:
```suggestion
    set_property(TEST solenoid.py.cleanup APPEND PROPERTY DEPENDS "solenoid.restart.py.cleanup")
```

ax3l:
Did a few CI updates in separate PRs (#994 #995) -- now this should pass :)

ax3l:
Hm, still
```
2025-06-23T07:05:56.5156822Z Run Build Command(s): /usr/local/bin/ninja -v pip_install
...
FAILED: lib/site-packages/impactx/impactx_pybind.cpython-310-x86_64-linux-gnu.so 
...
2025-06-23T07:06:00.2608304Z Error copying directory from "/home/runner/work/impactx/impactx/src/python/impactx" to "/home/runner/work/impactx/impactx/build/lib/site-packages/impactx".
...
2025-06-23T07:06:06.5634664Z ninja: build stopped: subcommand failed.
```

That looks still like running out of space to me...

tests/python/dashboard/utils.py
proy30: @EZoni, Definitely think there is a simpler alternative like your suggestion. I have not experimented with Trames server control, so I would not know for sure now. A new TODO on my end to see how it works.

ax3l:
Ok, that was actually a permission issue from doing
```
sudo ... install
... pip_install
```
Must have changed permissions in `build/` in the first step. fixed.

PR #971, opened 2025-05-20
Dashboard: refactor str -> numeric input logic

proy30:
There are two functions in the dashboard code that essentially solve the same problem of converting a text field input (str) to a numeric value. This PR condenses that logic into one.

### DIFF
--- a/src/python/impactx/dashboard/Input/generalFunctions.py
+++ b/src/python/impactx/dashboard/Input/generalFunctions.py
@@ -52,20 +52,26 @@ class generalFunctions:
     # -----------------------------------------------------------------------------
 
     @staticmethod
-    def determine_input_type(value):
+    def convert_to_numeric(input: 'str') -> int | float:
         """
-        Determines the type of the input value.
-        :param value: The input value whose type needs to be determined.
-        :return: A tuple containing the value converted to its determined type and the type itself.
+        Converts string inputs to their appropriate numeric type.
+        Needed as text fields inputs on the dashboard are inherently strings.
+
+        It first tries to convert the value to int, then to float.
+        If the conversion fails, it raises a ValueError.
+
+        :param input: The input to convert to a numeric type.
+        :return: The input converted to a numeric type.
         """
 
         try:
-            return int(value), int
+            return int(input)
         except ValueError:
             try:
-                return float(value), float
+                return float(input)
             except ValueError:
-                return value, str
+                raise ValueError(f"Cannot convert '{input}' to a numeric type.")
+
 
     @staticmethod
     def validate_against(input_value, value_type, additional_conditions=None):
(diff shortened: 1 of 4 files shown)


src/python/impactx/dashboard/Input/generalFunctions.py
```diff
@@ -52,20 +52,26 @@ def get_default(parameter, type):
...
         try:
-            return int(value), int
+            return int(input)
         except ValueError:
             try:
-                return float(value), float
+                return float(input)
             except ValueError:
-                return value, str
+                return None
```
EZoni, line 77: What happens if the conversion indeed fails and `None` is returned? Once `None` is assigned to whatever left-hand side variable (e.g., `parameter_value` above), do you expect things to work fine, is there a check somewhere else that does something special if it encounters `None`, or do you think something eventually crashes downstream?

proy30: The `convert_to_numeric` function is currently used for inputs that are already restricted to `type="number"`. I return `None` instead of raising an error because the function runs on every keystroke. For example, when a user types something like `1e` (as part of `1e8`), it’s not yet a valid number, and raising an error at that point would be disruptive.

Returning `None` allows the application to quietly ignore the input until it becomes valid, without breaking or interrupting the user. **There’s validation and error messaging in place** already for the same inputs, so this approach avoids redundancy. I could have returned the original value instead, but in this case, it makes no practical difference.

proy30: Example:
For an input with `type="number"`:

- User types `1` -> valid, successfully converts to a numeric value.
- User types `1e` or `-` -> invalid (until fully typed), so the conversion returns `None`.

At this point, the dashboard's existing validation will trigger an error message like:

> "Must be an int or float"  
(*Exact message depends on the default type for the input*)

Returning `None` here avoids prematurely throwing an exception while the user is still typing an incomplete but potentially valid number.

EZoni: Perfect, thanks for clarifying!

src/python/impactx/dashboard/Input/generalFunctions.py
```diff
@@ -52,20 +52,26 @@ def get_default(parameter, type):
...
         """
-        Determines the type of the input value.
-        :param value: The input value whose type needs to be determined.
-        :return: A tuple containing the value converted to its determined type and the type itself.
+        Converts string inputs to their appropriate numeric type.
+        This method is needed since text fields inputs on the dashboard
+        are inherently strings.
+
+        It first tries to convert the value to int, then to float.
+        If the conversion fails, returns None.
```
EZoni, line 62: Could we just add this note, or something along these lines, to the docstring?
```suggestion
        It first tries to convert the value to int, then to float.
        If the conversion fails, returns None.
        Note that the function runs on every keystroke.
        For non-trivial input, e.g., '1e-2', the conversion 
        fails silently until the full number is typed.
```

EZoni (approved):

PR #982, opened 2025-06-05
Dashboard: add `sim.periods` to inputs

proy30:
default = `1`
type = `int`
additional conditions = `Must be positive`

TODO
- [X] add to ui
- [X] set defaults (`default value`, `default type`, `additional conditions`)
- [X] connect to simulation validation (do not allow to run simulation if `periods` input is invalid)
- [X] connect to simulation

Lattice Configuration now:
image

Notes for future reference:
* Because `state.periods` matches `sim.periods`  (the naming is the same) and isn’t nested, the dashboard auto-loads tooltips from the `ImpactX` class and can parse `sim.periods` directly from `.py` files when importing.

### DIFF
--- a/src/python/impactx/dashboard/Input/latticeConfiguration/latticeMain.py
+++ b/src/python/impactx/dashboard/Input/latticeConfiguration/latticeMain.py
@@ -274,6 +274,11 @@ class LatticeConfiguration(CardBase):
             )
             with vuetify.VCardText(**self.CARD_TEXT_OVERFLOW):
                 with vuetify.VRow(**self.ROW_STYLE):
+                    with vuetify.VCol(cols=2):
+                        InputComponents.text_field(
+                            label="Periods",
+                        )
+                    vuetify.VDivider(vertical=True)
                     with vuetify.VCol(cols=True):
                         InputComponents.combobox(
                             label="Select Accelerator Lattice",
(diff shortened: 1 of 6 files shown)


src/python/impactx/dashboard/Input/latticeConfiguration/latticeMain.py
```diff
@@ -274,6 +274,11 @@ def card_content(self):
             )
             with vuetify.VCardText(**self.CARD_TEXT_OVERFLOW):
                 with vuetify.VRow(**self.ROW_STYLE):
+                    with vuetify.VCol(cols=2):
+                        InputComponents.text_field(
+                            label="Periods",
```
proy30, line 279: For future reference:

* We don't need to add `v_model_name="periods"` because `v_model_name` by default is 
`label.lower().replace(" ", "_")`. If the label were something like "Lattice Periods", I would have added `v_model_name="periods"`. Ideal to have the naming of the v_model be the same name defined in ImpactX

src/python/impactx/dashboard/Input/validation.py
```diff
@@ -174,6 +174,8 @@ def update_simulation_validation_status():
 
         if state.selected_lattice_list == []:
             error_details.append("LatticeListIsEmpty")
+        if state.periods_error_message:
+            error_details.append(f"Periods: {state.periods_error_message}")
```
proy30, line 178: This part will be refactored in a later PR to not have to manually write this every time a new non-nested state input is added.

cemitch99 (approved):
Great, thanks.

PR #981, opened 2025-06-02
Dashboard: restructure/refactor toolbar code

proy30:
**Note:** Commit history is inaccurate due to utilizing a merge of `dev` branch rather than rebasing. 

As the number of components in the toolbar grows (`import`, `export`, `reset`, `history`, etc.), it makes more sense long-term to split the logic into separate files.

### Changes in this PR:
- Split `Toolbar/controls.py` into three files:
  - `Toolbar/input.py`
  - `Toolbar/run.py`
  - `Toolbar/analyze.py`

- Moved the Python import parser code into its own directory:
  - Originally in `Toolbar/importParser.py` and `Toolbar/importParserHelper.py`
  - Now renamed to `parser.py` and `helper.py` respectively, and moved to `Toolbar/file_imports/python/`

- Previously, `Toolbar/file_imports/python/parser.py` handled both parsing and populating the dashboard UI.
  - This is now separated:
    - `parser.py` – remains in `Toolbar/file_imports/python/`, now focused purely on parsing
    - `ui_populator.py` – moved to `Toolbar/file_imports/`, handles dashboard UI integration

### Refactoring
- `ui_populator.py` logic is broken down into helper functions to isolate the import logic for each section. Prior to this, all of the logic was stored inside one method.
   - `_populate_distribution_inputs_to_ui`
   - `_populate_lattice_config_to_ui`
  - `_populate_lattice_config_variables_to_ui`
  
### Additions
- Docstrings/some comments have been added across toolbar components

----------------------------
- [x] Rebase after #977

PR #985, opened 2025-06-10
[WIP] Dashboard: add input error tracking

proy30:
This PR provides the user guidance on which inputs are invalid. Additionally, the PR refactors the current logic to track such errors.

**Changes:**
- A new dynamic notification banner in the `Inputs` toolbar which will display the count of input errors the user has. Hovering over the banner will tell the user exactly which inputs are incorrect along with the error message.
- A new class, `SimulationValidation` in `validation.py`, is created to encapsulate all logic for determining whether the simulation can run based on current input errors.
- The primary method in the class, `update`, takes the `input section` header as an argument.
  - This allows the method to check only the relevant section for errors.
- Add a JS function to suppress `resizeWarning` error.
  - Does not affect the dashboard functionality.
- Remove unhelpful (and unneeded) comments in `generalFunctions.py`
- Fix variable naming in `shared.py`
   - A variable referring to csr inputs was incorrectly labeled under space charge

**Demo:**

https://github.com/user-attachments/assets/b921e68d-f1a6-45af-b15c-6b23fc592152

Merge after https://github.com/BLAST-ImpactX/impactx/pull/981


src/python/impactx/dashboard/Input/inputParameters/inputMain.py
```diff
@@ -42,7 +42,6 @@ def on_tracking_mode_change(**kwargs) -> None:
         current_sc_list = ui_props.get("space_charge_list", [])
         if state.space_charge not in current_sc_list:
             state.space_charge = current_sc_list[0]
-        DashboardValidation.update_simulation_validation_status()
```
proy30: It is fine to not have to update the simulation with `tracking_mode` because `tracking_mode` is a dropdown and will always have a valid input.

PR #987, opened 2025-06-10
pytest-benchmark for all elements

ax3l:
This set of tests is used for performance benchmarking of individual elements pushes (as micro-benchmarks). We use this file to rapidly evaluate performance changes when tuning beamline element performance on CPUs and GPUs.

This integrates seamless with our pytest tests, e.g., to run only the benchmark tests (after `pip_install`):
```console
python -m pytest tests/python/test_benchmark_elements.py
```
There are additional command line flags for saving outputs and comparing them in the [pytest-benchmark documentation](https://pytest-benchmark.readthedocs.io).

Depends on:
- [x] pyAMReX feature https://github.com/AMReX-Codes/pyamrex/pull/447
- [x] work around this AMReX bug https://github.com/AMReX-Codes/amrex/issues/4498

### DIFF
--- a/tests/python/test_benchmark_elements.py
+++ b/tests/python/test_benchmark_elements.py
@@ -1,0 +1,466 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022-2025 The ImpactX Community
+#
+# Authors: Axel Huebl
+# License: BSD-3-Clause-LBNL
+#
+# This set of tests is used for performance benchmarking of individual
+# elements pushes (as micro-benchmarks). We use this file to rapidly
+# evaluate performance changes when tuning beamline element performance
+# on CPUs and GPUs.
+#
+# -*- coding: utf-8 -*-
+
+from functools import partial
+
+import pytest
+
+from impactx import ImpactX, Map6x6, distribution, elements
+
+# benchmark config
+rounds = 5
+npart = 1_000_000  # increase this to >10M or even 100M to avoid L1/L2/L3 cache effects on some hardware.
+
+# element config
+nslice = 1
+mapsteps = 4
+
+
+@pytest.fixture(scope="function")
+def sim():
+    class SimContextManager:
+        def __enter__(self):
+            self.sim = ImpactX()
+
+            # set numerical parameters and IO control
+            self.sim.particle_shape = 2  # B-spline order
+            self.sim.space_charge = False
+            self.sim.diagnostics = False  # benchmarking
+            self.sim.slice_step_diagnostics = False
+
+            self.sim.init_grids()
+
+            beam = self.sim.particle_container()
+            beam.clear_particles()
+
+            # load a 2 GeV electron beam with an initial
+            # unnormalized rms emittance of 2 nm
+            kin_energy_MeV = 2.0e3  # reference energy
+            bunch_charge_C = 1.0e-9  # used with space charge
+
+            #   reference particle
+            ref = self.sim.particle_container().ref_particle()
+            ref.set_charge_qe(-1.0).set_mass_MeV(0.510998950).set_kin_energy_MeV(
+                kin_energy_MeV
+            )
+
+            #   particle bunch
+            distr = distribution.Waterbag(
+                lambdaX=3.9984884770e-5,
+                lambdaY=3.9984884770e-5,
+                lambdaT=1.0e-3,
+                lambdaPx=2.6623538760e-5,
+                lambdaPy=2.6623538760e-5,
+                lambdaPt=2.0e-3,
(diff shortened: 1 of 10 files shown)


tests/python/test_benchmark_elements.py
```diff
@@ -0,0 +1,455 @@
...
+
+
+def pc_setup(sim):
+    """Fresh PC for each call of benchmark"""
+
+    # instead of drawing from the distribution again, create a 2nd
+    # particle container and copy the same initial particles again.
+    pc = sim.particle_container()
+    pc.clear_particles()
+    pc.add_particles(sim.backup_beam)
```
ax3l: @atmyers should I use
```suggestion
    pc.add_particles(sim.backup_beam, local=True)
```
re: https://github.com/AMReX-Codes/amrex/issues/4498 ?

ax3l: Same issue if used in both `add_particles` calls with `local=True`...

tests/python/test_benchmark_elements.py
```diff
@@ -0,0 +1,466 @@
...
+                lambdaX=3.9984884770e-5,
+                lambdaY=3.9984884770e-5,
+                lambdaT=1.0e-3,
+                lambdaPx=2.6623538760e-5,
+                lambdaPy=2.6623538760e-5,
+                lambdaPt=2.0e-3,
+                muxpx=-0.846574929020762,
+                muypy=0.846574929020762,
+                mutpt=0.0,
+            )
```
ax3l: We could make this simpler:
Maybe we can use simple numbers for twiss parameters of the distribution and a 1 GeV beam or so.

tests/python/test_benchmark_elements.py
```diff
@@ -0,0 +1,466 @@
...
+    focal_length = 0.5  # focal length in m
+    dtaper = 11.488289081903567  # 1/(horizontal dispersion in m)
+    dk = 1.0 / focal_length
+    el = elements.TaperedPL(name="pl", k=dk, taper=dtaper, unit=0)
+    benchmark.pedantic(el.push, setup=partial(pc_setup, sim), rounds=rounds)
+
+
+def test_ThinDipole(benchmark, sim):
+    el = elements.ThinDipole(name="kick", theta=0.45, rc=1.0)
+    benchmark.pedantic(el.push, setup=partial(pc_setup, sim), rounds=rounds)
```
ax3l, line 468: We could add a post-benchmark validation that:
- particles where not marked as invalid (e.g., turned out)
- no inf/nan is present in the pushed particle beam attributes

tests/python/test_benchmark_elements.py
cemitch99:
```suggestion
            distr = distribution.Waterbag(
                **twiss(
                     beta_x=1.0,
                     beta_y=1.0,
                     beta_t=1.0,
                     emitt_x=1e-09,
                     emitt_y=1e-09,
                     emitt_t=1e-06,
                     alpha_x=0.0,
                     alpha_y=0.0,
                     alpha_t=0.0,
                )
            )
```

tests/python/test_benchmark_elements.py
```diff
@@ -0,0 +1,466 @@
...
+            self.sim.slice_step_diagnostics = False
+
+            self.sim.init_grids()
+
+            beam = self.sim.particle_container()
+            beam.clear_particles()
+
+            # load a 2 GeV electron beam with an initial
+            # unnormalized rms emittance of 2 nm
+            kin_energy_MeV = 2.0e3  # reference energy
```
cemitch99:
```suggestion
            kin_energy_MeV = 1.0e3  # reference energy
```

tests/python/test_benchmark_elements.py
```diff
@@ -0,0 +1,466 @@
...
+            self.sim.diagnostics = False  # benchmarking
+            self.sim.slice_step_diagnostics = False
+
+            self.sim.init_grids()
+
+            beam = self.sim.particle_container()
+            beam.clear_particles()
+
+            # load a 2 GeV electron beam with an initial
+            # unnormalized rms emittance of 2 nm
```
cemitch99:
```suggestion
            # load a 1 GeV electron beam with an initial
            # unnormalized rms emittance of 1 nm
```

cemitch99 (approved):
I suggested a simplification of the beam energy/distribution inputs.  I think it's ok to leave the element parameters unchanged (from the originating examples).  Otherwise, it looks good to me.

ax3l:
ouch, because we did not release ImpactX yet, this relies on post 25.06 changes #993 in pyAMReX/WarpX -.

PR #989, opened 2025-06-10
CI: Windows 2019->2022

ax3l:
Update images since the Windows server 2019 image is EOL. https://github.com/actions/runner-images/issues/12045

Docs: https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -16,7 +16,7 @@ jobs:
     - uses: actions/setup-python@v5
       name: Install Python
       with:
-        python-version: '3.x'
+        python-version: '3.12'
     - name: CCache Cache
       uses: actions/cache@v4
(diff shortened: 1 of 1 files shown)


ax3l:
Hm, not fully clear why the Python tests segfault with ClangCl compiler.
They do run in pyAMReX (w/o OpenMP). Now, they segfault after:
```
Initializing AMReX (25.05)...
OMP initialized with 2 OMP threads
Segfault
See Backtrace.0.0 file for details
```

AMReX does not implement backtraces for Windows (yet) and debugging with [tmate](https://github.com/marketplace/actions/debugging-with-tmate) was hard to run an individual Python script so I could attach a debugger (need to try again separately).

Would merge as is for now, skip Python runtime tests, and continue in a separate PR...

PR #992, opened 2025-06-17
CI: Fix ClangCl Python

ax3l:
Follow-up to #989

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -126,7 +126,7 @@ jobs:
 
   build_win_clang:
     name: Clang w/ OMP
-    runs-on: windows-2022
+    runs-on: windows-2025
     if: github.event.pull_request.draft == false
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


ax3l:
Ok, so a solution might be for now to build ClangCl in Debug.

What I observe is: Only the Python tests segfault the moment that we build in Release/RelWithDebInfo. I was not yet able to run a Windows debugger on it, and this might be environment specific to this runner because we do _not_ see similar issues with ClangCl on Conda-Forge (building Release).

What could happen in the specific environment here: it is possible that the Python build pulls in, in some way, a debug build and standard-lib, which is not compatible in Windows (debug-symbol and release stdlibs are ABI incompatible in Windows). So for now for testing, it is sufficient to just roll with it and build Debug to keep going. Running `ldd` to find the dynamic libs did not immediately show something weird...

We could ultimately run a debugger like WinDbg (see commits for details) with `tmate` interactively on this and find out more, but for now I would book this as environment issue.

EZoni (approved):
Thanks for digging into this, @ax3l!

ax3l:
For future updates: I think we can set `Python3_FIND_ABI` on the command line for these CI runners to better fine-select with Python ABI flavor is picked up.
https://cmake.org/cmake/help/latest/module/FindPython3.html#hints

PR #993, opened 2025-06-17
AMReX & pyAMReX: `development`

ax3l:
Pull in the updates to the ParticleContainer bindings from https://github.com/AMReX-Codes/pyamrex/pull/447


ax3l:
ouch, because we did not release ImpactX yet, those are post 25.06 changes in pyAMReX/WarpX -.-

PR #998, opened 2025-06-23
`sim.verbose` also AMReX

ax3l:
While technically two options, we usually need this to set `sim.verbose = 0`. We now also set the AMReX verbose level to the same number, which ensures that init/finalize messages are also not printed, e.g.,
```
Initializing AMReX (25.05)...
OMP initialized with 14 OMP threads
AMReX (25.05) initialized
AMReX (25.05) finalized
```


cemitch99 (approved):
Looks ready to merge when all tests pass.

PR #999, opened 2025-06-23
Python: `sim.omp_threads`

ax3l:
Expose `amrex.omp_threads` (ParmParse) via a `ImpactX::omp_threads` attribute in Python.

### DIFF
--- a/tests/python/test_impactx.py
+++ b/tests/python/test_impactx.py
@@ -81,6 +81,11 @@ def test_impactx_nofile():
     """
     sim = ImpactX()
 
+    # various ways to change OMP threads from the default "nosmt"
+    sim.omp_threads = 1
+    sim.omp_threads = "1"
+    sim.omp_threads = "nosmt"
+
     sim.particle_shape = 2
     sim.slice_step_diagnostics = True
     sim.init_grids()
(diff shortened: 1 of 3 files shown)


tests/python/test_impactx.py
```diff
@@ -81,6 +81,11 @@ def test_impactx_nofile():
     """
     sim = ImpactX()
 
+    # various ways to change OMP threads from the default "nosmt"
+    sim.omp_threads = 1
+    sim.omp_threads = "1"
+    sim.omp_threads = "nosmt"
```
cemitch99, line 87: The string "nosmt" seems opaque to me.  Is there a more descriptive string that would work better for the default here?

ax3l: Ah that is a bit opaque in the test. The docstring and manual shows the meaning:
https://amrex-codes.github.io/amrex/docs_html/InputsComputeBackends.html

(SMT means "simultaneous multithreading" or [hyper-threading and is the same as Intel SMT](https://en.wikipedia.org/wiki/Hyper-threading). I.e., if there are 10 physical cores on your CPU but the system says there are "20" because it has 2x hyper-threading per core, then we only use 10, because we do not benefit from HT/SMT every time we tested)

AMReX default's  to `system`, but BLAST codes change that default to `nosmt`. ImpactX does that here (#475 & [AMReX](https://github.com/AMReX-Codes/amrex/pulls?q=is%3Apr+nosmt+is%3Aclosed)):
https://github.com/BLAST-ImpactX/impactx/blob/25.05/src/initialization/InitParser.cpp#L28-L30

cemitch99 (approved):

PR #1006, opened 2025-06-23
Explicit Power 2 & 3

ax3l:
@AlexanderSinn reminded me, that pow is really slow, e.g., on GPU. After a few experiments with Godbolt, it seems that indeed modern compilers do not replace powers of 2 and 3 with a (fast) multiplication unless fast-math is used. We currently only use fast-math on GPU, not CPU.

Thus, simple instances of `std::pow(..., 2)` and `std::pow(..., 3)` are replaced now with explict multiplications.

In many cases, writing power ^2 and ^3 is also shorter without `std::pow`.

- [ ] finalize all occurances

Tests:
- `-O3` on CPU https://godbolt.org/z/E951Gs9eW (current AMReX default)
- `-O3` on GPU https://godbolt.org/z/35TroGaTY
- `-O3` fast-math on CPU https://godbolt.org/z/31sonob91
- `-O3` fast-math on GPU https://godbolt.org/z/Yozdrcoa3 (current AMReX default)

### DIFF
--- a/src/elements/Aperture.H
+++ b/src/elements/Aperture.H
@@ -155,13 +155,13 @@ namespace impactx::elements
                   switch (m_shape)
                   {
                       case Shape::rectangular :  // default shape
-                        if (std::pow(u,2)>1 || std::pow(v,2) > 1_prt) {
+                        if (u * u > 1_prt || v * v > 1_prt) {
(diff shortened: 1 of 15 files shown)


ax3l:
Replaced by #1007

PR #1007, opened 2025-06-23
Use amrex::Math::powi

AlexanderSinn:
See #1006


ax3l (approved):
Thank you! :rocket: :sparkles:

ax3l (approved):

PR #1009, opened 2025-06-24
Fix chicane example plotting script

### LINKED ISSUE #997: Chicane example plot is different for python vs. input file
Running the python version of the chicane example with `python chicane.py` or with the input file `impactx input_chicane.in` and then `python plot_chicane.py` gives a different plot. The `analysis_chicane.py` file results are the same for the two runs. The first plot is from running with

Image

Image

 @chicane.py@ and the second is from @input_chicane.in@.

cemitch99:
This PR:

1) adds additional monitors to the chicane example input
2) modifies the script `plot_chicane.py` to reproduce the figures shown in the documentation [here](https://impactx.readthedocs.io/en/latest/usage/examples/chicane/README.html).

Fixes #997

### DIFF
--- a/examples/chicane/input_chicane.in
+++ b/examples/chicane/input_chicane.in
@@ -21,8 +21,8 @@ beam.mutpt = 0.999999961419755
 ###############################################################################
 # Beamline: lattice elements and segments
 ###############################################################################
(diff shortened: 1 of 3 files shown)


ax3l (approved):
Thank you! :)

PR #1011, opened 2025-06-25
PipeAperture Mix-In w/o Divisions

ax3l:
The `PipeAperture` class is a mixin class used in all thick elements and always checked and potentially applied for on the critical path of the computation.

A small optimization we can apply is to pre-compute the inverse of the pipe aperture in x and y, so that we can do a simple multiplication to calculate aperture horizontal and vertical comparison instead of a (way more) costly division _per particle_.

First spotted by @AlexanderSinn :+1:

### DIFF
--- a/src/elements/mixin/pipeaperture.H
+++ b/src/elements/mixin/pipeaperture.H
@@ -4,7 +4,7 @@
  *
  * This file is part of ImpactX.
  *
- * Authors: Axel Huebl, Chad Mitchell
+ * Authors: Axel Huebl, Chad Mitchell, Alexander Sinn
  * License: BSD-3-Clause-LBNL
  */
 #ifndef IMPACTX_ELEMENTS_MIXIN_PIPE_APERTURE_H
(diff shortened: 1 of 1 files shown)


cemitch99 (approved):
Looks great.  I would just add a comment line explaining why the inverse aperture size is being used, to avoid confusion for someone reading the code (since it could seem like an unnecessary complication).

src/elements/mixin/pipeaperture.H
```diff
@@ -60,42 +62,43 @@ namespace impactx::elements::mixin
             using namespace amrex::literals; // for _rt and _prt
 
             // skip aperture application if aperture_x <= 0 or aperture_y <= 0
-            if (m_aperture_x > 0 && m_aperture_y > 0) {
+            if (m_inv_aperture_x > 0_prt && m_inv_aperture_y > 0_prt) {
 
                // scale horizontal and vertical coordinates
```
ax3l, line 67:
```suggestion
               // scale horizontal and vertical coordinates
               // performance optimization: we intentionally store the inverse of
               //   the aperture, so we can do a quick multiplication (instead of a
               //   costly division) here.
```

src/elements/mixin/pipeaperture.H
```diff
@@ -60,42 +62,43 @@ namespace impactx::elements::mixin
...
         amrex::ParticleReal aperture_y () const
         {
-            return m_aperture_y;
+            using namespace amrex::literals; // for _prt
+            return 1_prt / m_inv_aperture_y;
         }
 
-        amrex::ParticleReal m_aperture_x = 0; //! horizontal aperture size [m]
-        amrex::ParticleReal m_aperture_y = 0; //! vertical aperture size [m]
+        amrex::ParticleReal m_inv_aperture_x = 0; //! inverse of the horizontal aperture size [1/m]
```
ax3l, line 106:
```suggestion
        // performance optimization: we intentionally store the inverse of
        //   the aperture, so we can do a quick multiplication instead of a costly
        //   division in the comparison operation per particle
        amrex::ParticleReal m_inv_aperture_x = 0; //! inverse of the horizontal aperture size [1/m]
```

ax3l:
Good point, comments added! :)

ax3l:
The same optimization can be added to the `Aperture` element: #1020

PR #1013, opened 2025-06-26
Removed unneeded specification for BOOST path.

egstern:
Notated that the module versions go with CPE 24.07.


ax3l (approved):
Perfect, thank you! :)

PR #1014, opened 2025-06-26
Dashboard: Test python importing

proy30:
The dashboard’s `ui populator` for the python parser silently broke after #981 due to a wrong function call. Since no test existed for file imports, CI didn’t catch it. This PR fixes the bug and adds a test to add coverage to the parser.

Additionally, a `conftest.py` is added to handle dashboard setup and shared fixtures. Tests reuse a single dashboard instance and reset inputs to a clean default state before each run.

**Changes:**
- Adds a test that imports a custom ImpactX Python file (testdata/example.py) into the dashboard
    - This file is a simple example to begin with and will have to be expanded when the python parser receives an update to handle more complex lattice builds.
- Checks that all expected values are populated correctly in both the UI and backend state.
- Add `conftest.py`

### DIFF
--- a/tests/python/dashboard/conftest.py
+++ b/tests/python/dashboard/conftest.py
@@ -1,0 +1,42 @@
+import pytest
+from seleniumbase import SB
+
+from .utils import (
+    DashboardTester,
+    start_dashboard,
+    wait_for_interaction_ready,
+    wait_for_server_ready,
+)
+
+
+@pytest.fixture(scope="session")
+def dashboard():
+    """
+    Sets up a single ImpactX dashboard server and headless browser for all tests in the session.
+    Automatically shuts down the server after all tests complete.
+    """
+
+    app_process = None
+
+    try:
+        app_process = start_dashboard()
+        wait_for_server_ready(app_process)
+
+        with SB(headless=True) as sb:
+            sb.open("http://localhost:8080/index.html#/Input")
+            wait_for_interaction_ready(sb)
+            yield DashboardTester(sb)
+    finally:
+        if app_process:
+            app_process.terminate()
+
+
+@pytest.fixture(autouse=True)
+def reset_dashboard_inputs(dashboard):
+    """
+    Resets the dashboard to its default state before each test.
+
+    This ensures all tests start from a clean, consistent baseline.
+    """
+    dashboard.sb.click("#Input_route")
+    dashboard.sb.click("#reset_all_inputs_button")
--- a/tests/python/dashboard/utils.py
+++ b/tests/python/dashboard/utils.py
@@ -94,6 +94,23 @@ class DashboardTester:
     def __init__(self, sb):
         self.sb = sb
 
+        # Set up the examples directory path once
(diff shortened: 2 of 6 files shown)


proy30:
@ax3l , Is there a way we could initialize Amrex just once for the dashboard?

(Manually highlighted in yellow to show that it gets called at the start of each test)
image

ax3l:
Hi @proy30 ,

Yes that is definitely possible, but let me make sure I understand what you like to do.

Currently, we intentionally start/shutdown AMReX for each unit test, keeping them in the same process for speed but isolating them somewhat to avoid confusion.

Do you want to keep AMReX "alive" for follow-up runs in the same dashboard or do you want to change something in the tests?

proy30:
@ax3l I’m suggesting initializing AMReX just once for the dashboard tests because currently only one of those tests (`test_dashboard.py`) is running a simulation. I assume only that test really needs AMReX involved. The test getting added in this PR (`test_python_import.py`) is purely interacting with the dashboard's UI. 

There's no problem with how it currently runs — everything works as expected. I am just thinking that the initialization process can be time-consuming as the number of dashboard tests grows, and the dashboard tests themselves don't actually need AMReX in most cases.

It could also be helpful to have the option to pick and choose which dashboard tests initialize AMReX.

ax3l:
Thanks, understood!

The init process is not too costly usually, so for sake of simplicity I would keep the logic as is: isolating the AMReX environments between subsequent dashboard tests for now.

But we can definitely revisit this for the dashboard in general and tests. There is some potential to "leak" state from prior sims into subsequent ones if we reuse the AMReX env, so something we need to evaluate if it is worth the trouble and also something to profile/benchmark if it is actually taking significant time for our use case.

tests/python/dashboard/conftest.py
```diff
@@ -0,0 +1,42 @@
+import pytest
```
ax3l, line 1: Just to understand: does this build on top of the fixtures in `tests/python/conftest.py` or is this its own new test root with this?

tests/python/dashboard/utils.py
```diff
@@ -94,6 +94,23 @@ class DashboardTester:
     def __init__(self, sb):
         self.sb = sb
 
+        # Set up the examples directory path once
+        impactx_directory = Path(get_impactx_root_dir())
```
ax3l, line 98: Unrelated to this PR:
It would be useful to generalize `get_impactx_root_dir` to also work with an installed ImpactX (after `pip_install` and with no original "source" dir present).

For that, I need to make sure we actually install the examples & tests (add control in the build system).

ax3l (approved):

tests/python/dashboard/conftest.py
proy30: `tests/python/dashboard/conftest.py` builds on top of `tests/python/conftest.py`, but the imports are not inherited.

ax3l: Thanks!

PR #1015, opened 2025-06-27
Dashboard: Add force quit

proy30:
Not exactly a "needed" feature, but it mimics a manual Ctrl+C in the terminal for a quick shutdown.

https://github.com/user-attachments/assets/168798fa-152f-48ab-a459-f7ff85bf61ee


ax3l (approved):
Good point :)

PR #1016, opened 2025-06-27
Dashboard: Fix runpy warning

proy30:
Prior to this PR, launching the dashboard with `python -m dashboard` consistently triggered the following warning:

```py
/usr/lib/python3.10/runpy.py:126: RuntimeWarning: 'dashboard.__main__' found in sys.modules
after import of package 'dashboard', but prior to execution of 'dashboard.__main__';
this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
```
The warning is now gone with simplifying `__main__.py` and moving the core application logic to `app.py`.


ax3l (approved):
Thanks for the fix! :)

PR #1017, opened 2025-06-28
Dashboard: Add test for lattice variable handler

proy30:
Testing to ensure the lattice can correctly reference a variable from the variables configuration.

    1. Add lattice element and map its variables.
    2. Delete a variable and check that the corresponding lattice parameter is invalid.
    3. Update a variable's value and ensure the lattice element reflects the change.
    4. Confirm invalid variable names produce an error.
    5. Reset variables and validate.


ax3l (approved):
Thanks! Yay, code coverage is increasing :)

PR #1018, opened 2025-06-28
Dashboard: Refactor analysis tab

proy30:
This is the first refactor PR for the analysis page since its initial implementation. Much of the code has been simplified and the functionality remains the same. Some UI adjustments were made, but that was not the primary goal. A follow-up PR will address user interaction improvements.

* The table and plot are now displayed in tabs instead of side by side. This layout may later move to the toolbar in follow-up.
* The default visualization is now automatically set to "Over-S Plot" instead of None. The visualization selector has been changed from a dropdown to a tab group on the toolbar because of the limited options currently.
* The plot's y-axis selector now has its own section. It remains a dropdown for now, but a follow-up PR will improve its interactivity. The changes are too extensive to include in this PR.
image


ax3l (approved):
Nice!!

PR #1019, opened 2025-06-29
`Aperture`: Hexagonal Mask Patterns

### LINKED ISSUE #931: Hexagonal arrangement for periodically masked aperture
Hi,

When using the aperture collimation with periodic masking, the hole patterns are currently arranged in a squared pattern to each other. Would it be possible to add a flag that enables different arangements, in particular a hexagonal / triangular? In that case every n+1 row (or column) would be offset by repeat_x/2 (or repeat_y/2)

Thank you,
Sarah

ax3l:
Support hexagonal mask patterns by adding an option to shift every other (odd) vertical period by a horizontal half-period. Alignment can be used to move the whole mask as needed.

Enhances #739
Close #931

## Visually

Default repeating pattern so far:
default

With this PR, one can shift every other line:
shift

## Checklist

- [x] [Notebook draft: Mask.zip](https://github.com/user-attachments/files/20970578/Mask.zip)
- [x] C++ implementation
- [x] Python bindings
- [x] docs: inputs & python
- [x] Test case: similar to existing one, reproduce the aperture in Python

### DIFF
--- a/src/elements/Aperture.H
+++ b/src/elements/Aperture.H
@@ -67,6 +67,7 @@ namespace impactx::elements
          * @param aperture_y vertical half-aperture (m)
          * @param repeat_x horizontal period for repeated masking, optional (m)
          * @param repeat_y vertical period for repeated masking, optional (m)
+         * @param shift_odd_x for hexagonal/triangular mask patterns: horizontal shift of every 2nd (odd) vertical period by repeat_x / 2. Use alignment offsets dx,dy to move whole mask as needed.
          * @param dx horizontal translation error in m
          * @param dy vertical translation error in m
          * @param rotation_degree rotation error in the transverse plane [degrees]
@@ -77,6 +78,7 @@ namespace impactx::elements
             amrex::ParticleReal aperture_y,
             amrex::ParticleReal repeat_x,
             amrex::ParticleReal repeat_y,
+            bool shift_odd_x,
             Shape shape,
             Action action,
             amrex::ParticleReal dx = 0,
@@ -86,7 +88,7 @@ namespace impactx::elements
         )
         : Named(std::move(name)),
           Alignment(dx, dy, rotation_degree),
-          m_shape(shape), m_action(action), m_repeat_x(repeat_x), m_repeat_y(repeat_y)
+          m_shape(shape), m_action(action), m_repeat_x(repeat_x), m_repeat_y(repeat_y), m_shift_odd_x(shift_odd_x)
         {
             using namespace amrex::literals; // for _rt and _prt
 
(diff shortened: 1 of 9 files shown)


ax3l:
Observed today by @cemitch99 and @egstern: is the line along y=0 of aperture holes somehow smaller in height (y)??

ax3l:
I tried the notebook linked above with a few more sizes and spheres and while the middle aperture is often off by a pixel or two, I think it is likely a plotting artifact on the y=0 holes we see to be slightly smaller.
image
image
:salt:

ax3l:
@SchroederSa @cemitch99 ready for review/merge :)

cemitch99 (approved):
Looks great!

src/elements/Aperture.H
```diff
@@ -101,6 +103,10 @@ namespace impactx::elements
             } else {
                 throw std::runtime_error("Aperture: aperture_y must be > 0!");
             }
+
+            if (m_repeat_y == 0_prt && m_shift_odd_x) {
+                throw std::runtime_error("Aperture: shift_odd_x can only be used if repeat_y is set, too!");
```
SchroederSa, line 108: I was stumbling over this sentence a little, maybe "Aperture: shift_odd_x can only be used if a value is set for repeat_y!" ore something similar.

src/elements/Aperture.H
```diff
@@ -67,6 +67,7 @@ namespace impactx::elements
          * @param aperture_y vertical half-aperture (m)
          * @param repeat_x horizontal period for repeated masking, optional (m)
          * @param repeat_y vertical period for repeated masking, optional (m)
+         * @param shift_odd_x for hexagonal/triangular mask patterns: horizontal shift of every 2nd (odd) vertical period by repeat_x / 2. Use alignment offsets dx,dy to move whole mask as needed.
```
SchroederSa, line 70: Is there any reason for horizontal shifting (x) but using the vertical repeat value (y)? Imagine we have x_repeat != y_repeat; one could consider having a shift_odd_x and a shift_add_y (maybe relevant for EEX applications where more advanced aperture shapes are needed).
There is no imminent need for this; it would just be a little more consistent in my view.

SchroederSa (commented):
Thanks @ax3l !
Is looking good to me, I am just a bit confused by mixing the two plans: shifting in one but using the repeat value of the other. Is this on purpose?
Best
Sarah

src/elements/Aperture.H
ax3l: Yes, so the idea is to shift every 2nd row (repeat_y) by half of repeat_x in x direction. For that to work, we need a pattern that has both non-zero repeat_x and repeat_y. (Technically, our input also allows stripes.)

src/elements/Aperture.H
ax3l: So the shifting only happens horizontally, in x (by repeat_x/2). But, we only shift every other row. Otherwise, we would not get a staggering.

In order to find that every-other-row, we check the vertical repetition (repeat_y) and count:
- 0: the usual pattern 
- 1: shift in x by repeat_x/2
- 2: the usual pattern
- 3: shift in x by repeat_x/2
- ...

PR #1020, opened 2025-06-29
`Aperture`: Remove Code Duplication, Avoid Division

ax3l:
- [x] Refactor our duplicate code paths in `Aperture` for transmit/absorb logic.
- [x] Avoid division as in #1011


cemitch99 (approved):
Looks good.

PR #1023, opened 2025-07-01
Dashboard: Update selector for over-s plot

proy30:
https://github.com/user-attachments/assets/8a24a866-ef85-4cc6-a5a7-983605996f7c


ax3l (approved):
Great UI!

We probably need to think about potentially having a 2nd y axis for some quantities (e.g. emittance), but for now this works well :+1:

PR #1025, opened 2025-07-02
Python: Interactive/In Situ Beam History

ax3l:
Add support to in situ store the reduced beam history with a beam instead of writing it to file. This is useful for interactive use of the simulation and to avoid writing files.

Before that, we had access to this information only through text files and we could access already this information, but only for the very last step or by writing our own push/tracker loop in Python.

Useful for the notebook in https://github.com/BLAST-ImpactX/2025_pilot_digital_twin_LDRD/pull/37

### DIFF
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -390,6 +390,17 @@ Particles
       :return: beam properties with string keywords
       :rtype: dict
 
+   .. py::property:: enable_beam_history
+
+      In situ record and store the beam history for every simulation step (default: ``False``).
+
+   .. py:method:: beam_history()
+
+      Return the history of the beam as calculated by the reduced beam characteristics on every step.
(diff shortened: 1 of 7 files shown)


docs/source/usage/python.rst
```diff
@@ -390,6 +390,17 @@ Particles
       :return: beam properties with string keywords
       :rtype: dict
 
+   .. py::property:: enable_beam_history
+
+      In situ record and store the beam history for every simulation step (default: ``False``).
```
cemitch99, line 395: The phrase "beam history" is a bit ambiguous, and it could suggest that the history of every particle is being stored.  Maybe we should specify that what is meant are the reduced diagnostics, etc.

cemitch99 (approved):
Ok, looks good to me.

PR #1027, opened 2025-07-04
Rename to Beam Moments History

ax3l:
Rename the recently introduced method to store beam moments over s to a clearer name.

Follow-up to #1025

### DIFF
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -390,23 +390,27 @@ Particles
 
       :param impactx.RefPart refpart: a reference particle to copy all attributes from
 
-   .. py:method:: reduced_beam_characteristics()
+   .. py:method:: beam_moments()
 
-      Compute reduced beam characteristics like the position and momentum moments of the particle distribution, as well as emittance and Twiss parameters.
+      Calculate beam moments at current ``s`` like the position and momentum moments of the particle distribution, as well as emittance and Twiss parameters.
 
       :return: beam properties with string keywords
       :rtype: dict
 
-   .. py:property:: enable_beam_history
+   .. py:property:: store_beam_moments
 
-      In situ record and store the beam history for every simulation step (default: ``False``).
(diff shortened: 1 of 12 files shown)


docs/source/usage/python.rst
```diff
@@ -383,24 +383,28 @@ Particles
...
+   .. py:method:: beam_moments()
 
-      Compute reduced beam characteristics like the position and momentum moments of the particle distribution, as well as emittance and Twiss parameters.
+      Calculate beam moments at current ``s`` like the position and momentum moments of the particle distribution, as well as emittance and Twiss parameters.
 
       :return: beam properties with string keywords
       :rtype: dict
 
-   .. py::property:: enable_beam_history
+   .. py::property:: store_beam_moments
```
```suggestion
   .. py:property:: store_beam_moments
```

ax3l:
As a comment, I think we thought about calling the RBC beam moments before and did not because some fo the quantities we calculated are not strictly moments but also correlations, etc.

Beam statistics would be more general, but I find maybe also a little less descript than beam moments, ignoring that we also refer to twins parameters, etc. with it for sake of simplicity.

Let me know what you prefer. Here is the full list again:
https://impactx.readthedocs.io/en/latest/dataanalysis/dataanalysis.html#reduced-beam-characteristics

egstern:
To my way of thinking, a statistic refers to any quantity derived from the
data, which would include first and second moments, also min and max and
correlations which are ratios of second moments. The emittance and twiss
functions that are there are also derived from second moments, The total
charge is a scaled 0th moment. So I think *statistics* is still a good term
(remembering that I suggested it.)

Calling them more generally *moments* is OK with me too, and again most of
the other quantities you mentioned are derived from moments.

At this point I think we're playing word games. Either one describes the
quantities and is not misleading.

cemitch99 (approved):
This looks fine to me.  Using "moments" seems perfectly reasonable.

PR #1029, opened 2025-07-04
Reset Particle Container

### LINKED ISSUE #1026: Reset Simulation
For runs where we repeatedly want to do:
```py
sim.track_particles()
```

from the beginning (e.g., in an optimizer or training loop), we need to provide a robust way to reset the simulation beam.

Currently, I do per run this and re-init the beam again via `sim.add_particles(...)` afterwards.
```py
    pc = self.simulation.particle_container()
    pc.clear_particles()
            
    ref = self.simulation.particle_container().ref_particle()
    # TODO: reset reference particle properly
    ref.s = 0
    ref.x = 0
    ref.y = 0
    ref.z = 0
    ref.t = 0
    ref.px = 0
    ref.py = 0
    ref.pz = 0
    ref.pt = 0
    # ref.sedge = 0  # missing
```

ax3l:
Add a functionality to clear the particle container and/or reset the reference particle.

Close #1026

### DIFF
--- a/src/particles/ReferenceParticle.H
+++ b/src/particles/ReferenceParticle.H
@@ -44,6 +44,24 @@ namespace impactx
         amrex::ParticleReal sedge = 0.0;  ///< value of s at entrance of the current beamline element
         amrex::SmallMatrix<amrex::ParticleReal, 6, 6, amrex::Order::F, 1> map{}; ///< linearized map
 
+        /** Reset the reference particle
+         *
+         * @param keep_mass do not reset the reference particle mass
+         * @param keep_charge do not reset the reference particle charge
+         */
+        void
+        reset (bool keep_mass=false, bool keep_charge=false)
+        {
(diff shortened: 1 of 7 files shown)


src/particles/ReferenceParticle.H
```diff
@@ -44,6 +44,14 @@ namespace impactx
         amrex::ParticleReal sedge = 0.0;  ///< value of s at entrance of the current beamline element
         amrex::SmallMatrix<amrex::ParticleReal, 6, 6, amrex::Order::F, 1> map{}; ///< linearized map
 
+        /** Reset the reference particle */
```
ax3l: For practical applications, it could be nice if we did not reset the mass and charge.
Let me know what you think.

ax3l: Added options for that now.

cemitch99 (approved):
Great.

PR #1028, opened 2025-07-04
Dashboard: allow to load outputs from history

proy30:
Close https://github.com/BLAST-ImpactX/impactx/issues/941

https://github.com/user-attachments/assets/cf9ecea7-4f70-4d99-91dc-a8d2f8d30a7b

Additionally changes were made to `tests/python/dashboard/test_lattice_variable_handler.py` to improve robustness. The key change includes adding a helper function that will retry an assert for ~10 seconds before deeming a fail. This should allow the UI to smoothly update its state before having to be checked.

PR #1032, opened 2025-07-06
Dashboard: Centralize Server Instance

proy30:
Reduce redundant server initialization to just a single instance.

~~**Note:** Rebase after https://github.com/BLAST-ImpactX/impactx/pull/1028~~

### DIFF
--- a/src/python/impactx/dashboard/Analyze/over_s/plot.py
+++ b/src/python/impactx/dashboard/Analyze/over_s/plot.py
@@ -8,9 +8,7 @@ License: BSD-3-Clause-LBNL
 
 import plotly.graph_objects as go
 
-from ... import setup_server
-
-server, state, ctrl = setup_server()
+from ... import state
 
 
 def over_s_plot():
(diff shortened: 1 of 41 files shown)


proy30:
Tests should pass with the changes in #1028

ax3l (approved):
Awesome, thank you!

Can you remind me how to make a trame instance local to a connected client? I am planning to deploy this to NERSC Spin and want that every client has their own instance to work with.

proy30:
@ax3l Does deploying on NERSC Spin mean we’ll run one dashboard 24/7? For that setup, currently every browser client connects to the same Trame server and shares its global state—so they’d overwrite each other. 

One way around that is to have some type of “Launch” button that spins up a brand-new dashboard process on its own port for each user (for example, I could do something like this locally

```py
python3 -m dashboard --port 8080  
python3 -m dashboard --port 8081
```
), which would then two isolated dashboard instances

ax3l:
> Does deploying on NERSC Spin mean we’ll run one dashboard 24/7?

We would "elastically" configure like 20 instances to be spun up on demand when people connect :)
They will "appear" to be online 24/7, yes :)

PR #1038, opened 2025-07-09
Fix Quad Transport Map w/ Slicing and Zero Fields

ax3l:
The linear transport map of the quad, used in envelope tracking mode, did not scale its lengths properly when set to zero field strength and sliced for space charge effects. This fixes it.

Isolated from #1030

Author: @kyrsjo :pray:

### DIFF
--- a/src/elements/Quad.H
+++ b/src/elements/Quad.H
@@ -4,7 +4,7 @@
  *
  * This file is part of ImpactX.
  *
- * Authors: Chad Mitchell, Axel Huebl
+ * Authors: Chad Mitchell, Axel Huebl, Kyrre Ness Sjobak
  * License: BSD-3-Clause-LBNL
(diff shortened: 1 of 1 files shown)


cemitch99 (approved):
Yes, excellent.  As pointed out - this only affects the result in envelope mode when k=0.

kyrsjo:
Ah, thanks. Sorry, I got busy with other things for a few days. I started by writing tests here (not 100% finished), as you requested:
https://github.com/kyrsjo/impactx/tree/fix_transportmap_slicing

I was a little surprised by the tests living in the examples folder, and the test folder being something else...

There is also this related bug:
https://github.com/BLAST-ImpactX/impactx/pull/1030/files#r2198822227

Should that also go in a separate PR?

PR #1041, opened 2025-07-11
ConstF: Allow kx<0 and ky<0

kyrsjo:
This adds support for negative focusing strengths `kx` and `ky` for the element `ConstF`.

It also significantly updates the example `active_plasma_lens`, demonstrating both `ConstF` and `ChrPlasmaLens` elements and added in PR #1030. The CTest setup for this example was also simplified, avoiding needlessly copying files.

Finally, I signed my name to the top of both `ConstF.H` and `ChrPlasmaLens.H`, as was previously suggested.

Original top comment:
> This pulls out the changes to ConstF from PR #1030 . It allows the ConstF elements to handle negative (defocusing) k values.
> 
> - [x] Add testing.
> 
> Question: Why is the unit of kx and ky in this element 1/m, i.e. the square root of the more common definition of k?


cemitch99:
@kyrsjo I don't see an element exactly analogous to this one in the MAD-X documentation.  However, (at least for the transverse focusing), the closest analogy is the SOLENOID.  If the rotation in x-y is ignored, then the solenoid provides equal focusing in x/y with strength k=Bz/Brho, following the convention defined here.  In other words, the focusing field strength normalized by Brho.

src/elements/ConstF.H
```diff
@@ -100,19 +100,60 @@ namespace impactx::elements
...
 
             // trigo
-            auto const [sin_kxds, cos_kxds] = amrex::Math::sincos(m_kx * slice_ds);
-            m_cos_kxds = cos_kxds;
-            m_const_x = -m_kx * sin_kxds;
-            auto const [sin_kyds, cos_kyds] = amrex::Math::sincos(m_ky * slice_ds);
-            m_cos_kyds = cos_kyds;
-            m_const_y = -m_ky * sin_kyds;
+            // Note: The convention here is m_kx[1/m] = sign(k)*sqrt(abs(k))
+            // where k[1/m^2] is in the usual MADX convention
```
cemitch99: This statement is not clear, since there is not direct MADX analog.  For solenoids, the focusing strength is k [1/m].

src/elements/ConstF.H
```diff
@@ -239,29 +280,67 @@ namespace impactx::elements
             amrex::ParticleReal const betgam2 = powi<2>(refpart.pt) - 1.0_prt;
 
             // trigo
-            auto const [sin_kxds, cos_kxds] = amrex::Math::sincos(m_kx * slice_ds);
-            amrex::ParticleReal const_x = -m_kx * sin_kxds;
-            auto const [sin_kyds, cos_kyds] = amrex::Math::sincos(m_ky * slice_ds);
-            amrex::ParticleReal const_y = -m_ky * sin_kyds;
+            // Note: The convention here is m_kx[1/m] = sign(k)*sqrt(abs(k))
+            // where k[1/m^2] is in the usual MADX convention
```
cemitch99: See the comment above.

cemitch99:
This looks great; thanks a lot!  I made only a minor comment.  It looks ready to go once a simple test is added.  Let us know if you would like some help with this.

kyrsjo:
Hi, that's great!

I'm on summer holiday right now, will be back in August. I can look at it then, or if you want to add the test and merge, please feel free!

Thanks again for all the help - it makes it a lot more likely that I will contribute again, and now I know your setup and standards!

src/elements/ConstF.H
kyrsjo: I was mainly thinking that this is a strong analog to quadrupole mangets and plasma lenses, since the equations are exactly the same*, and there is no rotation implemented between the horizontal/vertical plane? So providing the conversion explicitly to the units used for quadrupoles makes sense IMO.

 `*` Except for the convention of how `k` works.

kyrsjo:
TODO: Fix https://github.com/BLAST-ImpactX/impactx/pull/1030/files/4866ef838a31ed23f185e52f1978a544c740fafe#r2361158766

Basically in `run_APL_tracking()`, delete the comment 
```
# [Doesn't give same result as ChrPlasmaLens and ChrQuad for some reason, even when sigpt_0 = 0]
```

kyrsjo:
In tests, also look at this for how I get the data out:
https://github.com/BLAST-ImpactX/impactx/pull/1030#discussion_r2361210443
Says:

> 
> Note: This works, but the values defined here are computed internally and stored as reduced_diagnostics. See this example for an alternative approach: https://impactx.readthedocs.io/en/latest/usage/examples/fodo_space_charge/README.html
> It's fine to keep this for now.

Relative to:
    return (beta_x, beta_y, alpha_x, alpha_y)
in analysis_APL_ChrPlasmaLens.py
function `get_twiss()`

kyrsjo:
TODO: Fix this https://impactx.readthedocs.io/en/latest/usage/examples/active_plasma_lens/README.html

kyrsjo:
TODO's above are FIXED.

kyrsjo:
@ax3l How do I add figures to the documentation like in https://impactx.readthedocs.io/en/latest/usage/examples/optimize_triplet/README.html#visualize ?

kyrsjo:
It looks like these tests are failing due to errors in the test system on GitHub, not in the actual tests.

kyrsjo:
Seems like the pre-commit wanted the exact opposite order of imports as me. Oh well, I just wanted them that way for readability (there is a logical order), but standards are standards I guess...

kyrsjo:
> @ax3l How do I add figures to the documentation like in https://impactx.readthedocs.io/en/latest/usage/examples/optimize_triplet/README.html#visualize ?

Seems like I can just commit it and reference it in the .rst, just like for code examples... See eg. https://github.com/BLAST-ImpactX/impactx/pull/1041/commits/8f59642d00d61ddc6c139385a2c512f8dbbae110

Why is this not done other places? Why use out-of-repo gists?

kyrsjo:
@cemitch99 I think this is now ready to merge?

cemitch99 (approved):
I verified manually all of the tests, and everything appears to work as expected.  Thanks for the thorough contribution!

kyrsjo:
Thanks for reviewing!

Also note that I added the images for the manual in a slightly different way than what you normally do (i.e. in the repo, which then gets pulled in during building of the REST stuff) - I think its cleaner since it avoids hosting the files in various places across the web and also versions them, and I hope its fine to do it like that. They are small png files.

kyrsjo:
LGTM ?

ax3l (approved):
Thank you, this looks great!

PR got a bit long, but in it goes now :)

Will ship with 25.12 :)

kyrsjo:
Thank you :D :tada: 

Yes, I did a lot of documentation/test work in this PR as well, which was related to the code work. But indeed, it made it long.

PR #1045, opened 2025-07-12
Dashboard: Set default name for BeamMonitor if name is empty

proy30:
A convenient case to cover in my usage of the dashboard. The way I’ve seen `elements.BeamMonitor` added in ImpactX examples follows this structure: `elements.BeamMonitor("monitor", backend="h5")`. The dashboard doesn’t catch "monitor" in that case because it’s passed positionally — not like `elements.BeamMonitor(name="monitor", backend="h5")`. Since the dashboard expects `name=`, it ends up empty and throws a validation error. This PR sets a default name of `"DefaultName"` if the value is missing to avoid that. To do this nicely, we also adjust the loop that adds a lattice element to make it more readable.

### DIFF
--- a/src/python/impactx/dashboard/Input/defaults.py
+++ b/src/python/impactx/dashboard/Input/defaults.py
@@ -38,6 +38,7 @@ TRACKING_MODE_PROPERTIES: dict[str, dict[str, Any]] = {
     },
 }
 
+BEAM_MONITOR_DEFAULT_NAME = "DefaultName"
 
 class DashboardDefaults:
     """
--- a/src/python/impactx/dashboard/Input/defaults.py
+++ b/src/python/impactx/dashboard/Input/defaults.py
@@ -40,6 +40,7 @@ TRACKING_MODE_PROPERTIES: dict[str, dict[str, Any]] = {
 
 BEAM_MONITOR_DEFAULT_NAME = "DefaultName"
 
+
 class DashboardDefaults:
     """
     Defaults for simulation parameters in the ImpactX dashboard.
--- a/src/python/impactx/dashboard/Input/lattice/ui.py
+++ b/src/python/impactx/dashboard/Input/lattice/ui.py
@@ -56,7 +56,9 @@ def add_lattice_element() -> dict:
 
     parameters = []
     for name, default_value, default_type in parameters_data:
-        value = "DefaultName" if selected_lattice == "BeamMonitor" and name == "name" else default_value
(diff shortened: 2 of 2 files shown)


src/python/impactx/dashboard/Input/defaults.py
```diff
@@ -38,6 +38,8 @@
     },
 }
 
+BEAM_MONITOR_DEFAULT_NAME = "DefaultName"
```
ax3l, line 41: A better default name would be `monitor`.

src/python/impactx/dashboard/Input/lattice/ui.py
```diff
@@ -44,37 +45,43 @@
...
     )
 
-    selected_lattice_element = {
-        "name": selected_lattice,
-        "parameters": [
+    parameters = []
+    for name, default_value, default_type in parameters_data:
+        value = default_value
+
+        if selected_lattice == "BeamMonitor" and name == "name" and not value:
```
ax3l, line 62: There is a challenge here assigning the same default name to multiple beam monitors as a fallback, when in reality they could have a different name:

If we encounter the same BeamMonitor in a lattice using the same `name` attribute, then we intentionally aggregate the measured beam data into the same file series (i.e. file) -- and if the `name` differs they go in separate data series (i.e. different files).

So assuming we can fall back to assigning a default name works for most examples we parse here, but is error prone.

ax3l (commented):
We can try this because it should work for many examples, but it is a bit of a bandaid / assumption / hack :)

src/python/impactx/dashboard/Input/lattice/ui.py
proy30: We could assign a unique ID to each BeamMonitor, but at this point, it may be better to fix the default name logic directly instead of applying a bandaid.

Aside from the lines that insert the fallback name for BeamMonitor, I'd like to keep the refactored add lattice element logic.. Would it make sense to remove the BeamMonitor-specific logic and retitle the PR accordingly?

ax3l: Yeah the unique id would be a guess, too - and in our examples it is most of the time not a unique one but we actually reuse the name, as you currently implement.

I think for this PR, it is ok to give one name to all BeamMonitors you encounter in parsing of examples, it is mostly correct for our current examples, but of course users might write something different in their code for import.

ax3l (approved):

PR #1046, opened 2025-07-13
Dashboard: Disable “Add” Button for Invalid Lattice Elements

proy30:
Before:
Selecting an invalid element popped up an error message.

After:
The “Add” button is grayed-out until a valid element is chosen.

Before:

https://github.com/user-attachments/assets/3f62da16-e153-4cee-8358-a7f503ed42e1

After:

https://github.com/user-attachments/assets/35e35fde-77f4-424c-bcc4-e05a2befd51a


ax3l (approved):
Nice! :)

PR #1053, opened 2025-07-17
Dashboard Tests: Add lattice statistics coverage to `test_dashboard.py`

proy30:
Just making sure it works while testing. I don't think an extensive test is needed (ie importing a large lattice configuration and then testing stats).

Also updated a print statement inside of `tests/python/dashboard/utils.py` to help with debugging.


ax3l (approved):
Nice, thank you for covering this part in tests!

PR #1059, opened 2025-07-22
Particle Shape: Optional w/o Collective

ax3l:
Make the particle shape input mostly optional, unless collective effects are enabled.

This works pretty well for most tests, but will need a follow-up to make it also possible to skip in Python inputs that use space charge with envelope tracking.


cemitch99 (approved):
Looks good to me.

PR #1066, opened 2025-07-29
Load PALS Lattice

ax3l:
- [x] structure
- [x] docs
- [x] depends on: https://github.com/campa-consortium/pals-python/issues/17
- [x] translation logic
- [x] example

The example requires installation of pals-schema 0.1.1.


src/python/impactx/extensions/KnownElementsList.py
```diff
@@ -6,14 +6,70 @@
...
+    )
+
+
+def from_pals(self, pals_line, nslice=1):
+    """Load and append a lattice from a Particle Accelerator Lattice Standard (PALS) Python Line.
+
+    https://github.com/campa-consortium/pals-python
+    """
+    # TODO: Loop over the pals_line and create a new ImpactX KnownElementsList from it.
+    #       Use self.extend(...) on the latter.
```
ax3l: Here goes the logic we need to translate from PALS, something like:
```py
ix_line = []
for pals_element in pals_line:
    if isinstance(pals_element, pals.Drift):
        ix.line.append(elements.Drift(...))
    elif ...:  # Quad, etc.

self.extend(ix_line)
```

examples/pals/README.rst
```diff
@@ -0,0 +1,47 @@
+.. _examples-fodo-pals:
+
+FODO Cell with a PALS Lattice
+=============================
+
+This is a duplicate of the FODO cell example ``examples-fodo``, except that the lattice file is specified in PALS-compliant format.
```
EZoni: Shall we add a link to the standard?
```suggestion
This is a duplicate of the FODO cell example ``examples-fodo``, except that the lattice file is specified in `PALS <https://pals-project.readthedocs.io>`__-compliant format.
```

examples/pals/run_fodo_pals_json.py
EZoni, line 1: Not sure how you manage this kind of situation in ImpactX, but I wanted to point out that the only difference between `run_fodo_pals_json.py` and `run_fodo_pals_yaml.py` is the file extension `.json` vs `.yaml` in the file loading call `sim.lattice.load_file(...)` on line 51 - so it could be possible to merge all the code but that file name.

examples/pals/run_fodo_pals_json.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022-2023 ImpactX contributors
+# Authors: Marco Garten, Axel Huebl, Chad Mitchell
+# License: BSD-3-Clause-LBNL
```
EZoni: Copyright and authors to be updated.

examples/pals/run_fodo_pals_yaml.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022-2023 ImpactX contributors
+# Authors: Marco Garten, Axel Huebl, Chad Mitchell
+# License: BSD-3-Clause-LBNL
```
EZoni: Copyright and authors to be updated.

src/python/impactx/extensions/KnownElementsList.py
```diff
@@ -6,14 +6,95 @@
...
+
+        # example: fodo.pals.yaml
+        if extension == ".json":
+            import json
+
+            # Read the JSON data from the test file
+            with open(filename, "r") as file:
+                json_data = json.loads(file.read())
+            # Parse the JSON data back into a Line object
+            self.from_pals(Line(**json_data), nslice)
```
EZoni: Maybe we could add an inline comment that clarifies that validation happens here when we load the data in a `Line`:
```suggestion
            # Parse the JSON data back into a Line object
            # (data validation happens here automatically)
            self.from_pals(Line(**json_data), nslice)
```

src/python/impactx/extensions/KnownElementsList.py
```diff
@@ -6,14 +6,95 @@
...
+            return
+
+        elif extension == ".yaml":
+            import yaml
+
+            # Read the YAML data from the test file
+            with open(filename, "r") as file:
+                yaml_data = yaml.safe_load(file)
+            # Parse the YAML data back into a Line object
+            self.from_pals(Line(**yaml_data), nslice)
```
EZoni: Maybe we could add an inline comment that clarifies that validation happens here when we load the data in a `Line`:
```suggestion
            # Parse the YAML data back into a Line object
            # (data validation happens here automatically)
            self.from_pals(Line(**yaml_data), nslice)
```

EZoni (commented):
Thanks for setting this up, @ax3l and @cemitch99! I had a first pass at the PR and left a few comments, to be processed whenever we mark this as ready for review.

examples/pals/run_fodo_pals_json.py
ax3l: Absolutely!

examples/pals/run_fodo_pals_json.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022-2023 ImpactX contributors
+# Authors: Marco Garten, Axel Huebl, Chad Mitchell
```
ax3l:
```suggestion
# Copyright 2022-2025 ImpactX contributors
# Authors: Axel Huebl, Chad Mitchell, Edoardo Zoni
```

examples/pals/run_fodo_pals_yaml.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022-2023 ImpactX contributors
+# Authors: Marco Garten, Axel Huebl, Chad Mitchell
```
ax3l:
```suggestion
# Copyright 2022-2025 ImpactX contributors
# Authors: Axel Huebl, Chad Mitchell, Edoardo Zoni
```

examples/pals/run_fodo_pals_json.py
ax3l: Because this is in `examples/` and not in `tests/` (i.e., it is user-facing), I would only load a yaml file.

Technicalities like checking if the json parser works do not need to be covered with a user-facing example (but with tests, here or in pals-python).

src/python/impactx/extensions/KnownElementsList.py
```diff
@@ -6,14 +6,97 @@
...
+    )
+
+
+def from_pals(self, pals_line, nslice=1):
+    """Load and append a lattice from a Particle Accelerator Lattice Standard (PALS) Python Line.
+
+    https://github.com/campa-consortium/pals-python
+    """
+    from pals_schema.DriftElement import DriftElement
+    from pals_schema.QuadrupoleElement import QuadrupoleElement
```
ax3l, line 64: @EZoni the PyPI package might be called `pals-schema`, and that is ok. But for Python imports I would still call the module simply `pals`.

There is no module we are aware of that we will collide with using that name and we will not develop another `pals_...` module, so the suffix is a bit verbose.

EZoni: Got it, just opened https://github.com/campa-consortium/pals-python/issues/23 to track this.

examples/requirements.txt
```diff
@@ -5,5 +5,6 @@ git+https://github.com/ax3l/PyNAFF.git@fix-complex-warning
 matplotlib
 numpy
 openpmd-api
+pals-schema
```
ax3l: I wonder if I should write
```suggestion
pals-schema~=0.1
```
because we will release and break often :D

ax3l:
@cemitch99 @EZoni let me know what you think, we could probably merge this as a first quick draft in that state. I nailed the PALS Python version to semantic all compatible `0.1` releases in `requirements.txt`, so we should be able to continue to develop `pals-python` and break as needed without breaking CI in ImpactX :crossed_fingers:

requirements.txt
```diff
@@ -1,2 +1,3 @@
 numpy>=1.15
+pals-schema~=0.1
```
ax3l: Oh, this would match anything `<1.0`.

If we want `>=0.1.1` and `<0.2` it must read:
```suggestion
pals-schema~=0.1.1
```

examples/pals/fodo.pals.json
EZoni, line 1: Since the JSON file is not used in a test at this point, wouldn't it be safer to remove it? Otherwise it may get easily outdated as PALS evolves. For example, I believe even a simple example like this one may change after, say, https://github.com/campa-consortium/pals-python/pull/16.

cemitch99 (approved):
Ready to go once tests pass.

ax3l:
@EZoni let us make sure to use semantic versioning for PALS Python releases, e.g., `0.2` when we break API so ImpactX CI does not break.

It will be no issue if we break-and-bump often and end up with `0.123.0` before the first `1.0` release.

examples/pals/fodo.pals.json
ax3l: Oops, I forgot to push that. Thanks!

EZoni:
> @EZoni let us make sure to use semantic versioning for PALS Python releases, e.g., 0.2 when we break API so ImpactX CI does not break.

For now in PALS Python we have been using what I thought was semantic versioning in the form `MAJOR.MINOR.PATCH`, e.g., `0.1.0`, `0.1.1`, etc. 

We started with `0.1.0` as they suggest in https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase and then published `0.1.1` for the beamline name fix.

Is this compatible with what you are saying? I'm not sure I understood correctly.

PR #1070, opened 2025-07-30
OpenMP Parallelize InitDistribution

ax3l:
The drawing of particles from the distribution function was not threaded on CPU. This changes that to improve performance.

Fix #1069

```
Initializing AMReX (25.07-32-gaf07c6f1d7b8)...
MPI initialized with 1 MPI processes
MPI initialized with thread support level 0
OMP initialized with 14 OMP threads
AMReX (25.07-32-gaf07c6f1d7b8) initialized

Grids Summary:
  Level 0   1 grids  5120 cells  100 % of domain

Beam kinetic energy (MeV): 250
Bunch charge (C): 1e-09
Particle type: electron
Number of particles: 30000000

...
```

## BEFORE

```
TinyProfiler total time across processes [min...avg...max]: 33.99 ... 33.99 ... 33.99

--------------------------------------------------------------------------------------------------------------
Name                                                           NCalls  Excl. Min  Excl. Avg  Excl. Max   Max %
--------------------------------------------------------------------------------------------------------------
ParticleContainer::RedistributeCPU()                               41      7.137      7.137      7.137  21.00%
ImpactX::add_particles                                              1      5.231      5.231      5.231  15.39%
impactx::spacecharge::GatherAndPush                                40      5.045      5.045      5.045  14.84%
...
```

## NOW

```
TinyProfiler total time across processes [min...avg...max]: 29 ... 29 ... 29

--------------------------------------------------------------------------------------------------------------
Name                                                           NCalls  Excl. Min  Excl. Avg  Excl. Max   Max %
--------------------------------------------------------------------------------------------------------------
ParticleContainer::RedistributeCPU()                               41      6.973      6.973      6.973  24.05%
impactx::spacecharge::GatherAndPush                                40       4.95       4.95       4.95  17.07%
ImpactXParticleContainer::DepositCharge                            40      2.793      2.793      2.793   9.63%
ablastr::particles::deposit_charge::ChargeDeposition               90      2.351      2.351      2.351   8.11%
impactx::transformation::CoordinateTransformation                  80      1.796      1.796      1.796   6.19%
impactx::transformation::CoordinateTransformation::to_fixed_t      86      1.717      1.717      1.717   5.92%
impactx::Push::Drift                                               40      1.686      1.686      1.686   5.81%
impactx::transformation::CoordinateTransformation::to_fixed_s      84      1.436      1.436      1.436   4.95%
ImpactXParticleContainer::MinAndMaxPositions                       41      1.151      1.151      1.151   3.97%
[...]

PR #1072, opened 2025-07-30
Reuse Particle Chunking Logic

### LINKED ISSUE #1071: Refactor Out Divide-and-Conquer Process/Thread Math
We should refactor out the splitting / offset of number of particle logic from `ImpactX::add_particles` and `ImpactXParticleContainer::AddNParticles` and make it a helper function. After #1070, we use it already >=3 times in ImpactX itself:
- `int const myproc = ...` for MPI distribution of particles
- `int const navg_thread = ...` for threading
- https://github.com/BLAST-ImpactX/impactx/blob/25.07/src/particles/ImpactXParticleContainer.cpp#L253-L270 for threading

This is needed so often that I think this logic actually could go into AMReX itself. This can be a helper function that returns:
- offset in the global array
- how many elements to handle

depending on a passed id (MPI proc or OMP thread) and size (MPI size or OMP max threads).

ax3l:
Write a common function to partition/chunk the particle array for MPI and OpenMP parallelization.

- [x] rebase after #1070 was merged
- [x] `int`/`amrex::Long`

Close #1071

This function might also be a good candidate for ABLASTR or even AMReX.

### DIFF
--- a/src/particles/SplitEqually.H
+++ b/src/particles/SplitEqually.H
@@ -10,13 +10,15 @@
 #ifndef IMPACTX_SPLIT_EQUALLY_H
 #define IMPACTX_SPLIT_EQUALLY_H
 
+#include <AMReX_INT.H>
+
 namespace impactx
 {
     /** Offset and size of a chunk in a contigous array (e.g., of particle properties) */
     struct ParticleChunk
     {
-        int offset = 0;  //! where to start in the array, in elements
-        int size = 0;    //! how large is my chunk, counted from offset, in elements
+        amrex::Long offset = 0;  //! where to start in the array, in elements
+        amrex::Long size = 0;    //! how large is my chunk, counted from offset, in elements
     };
 
     /** Split an array of contiguous elements into well-balanched chunks.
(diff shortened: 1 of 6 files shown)


src/particles/SplitEqually.H
```diff
@@ -0,0 +1,40 @@
...
+#ifndef IMPACTX_SPLIT_EQUALLY_H
+#define IMPACTX_SPLIT_EQUALLY_H
+
+namespace impactx
+{
+    /** Offset and size of a chunk in a contigous array (e.g., of particle properties) */
+    struct ParticleChunk
+    {
+        int offset = 0;  //! where to start in the array, in elements
+        int size = 0;    //! how large is my chunk, counted from offset, in elements
```
ax3l: While we are on it, we could make this `amrex::Long`, take the 3-5% performance hit in `ParallelFor(RNG)`, and support more than 2,147,483,647 particles per process (OpenMP) and globally (MPI init).

So far, the implementation in ImpactX uses `int` which on our 64bit platforms (linux/macos/windows) defaults to 4 Byte signed ints with the above range.

Since this logic here is also used for MPI distribution, it is pretty much possible to reach that limit for high-quality runs, even though it is not super common.

ax3l: If we really want to optimize, we template it and walk the line of only using `amrex::Long` for MPI logic.

ax3l: Made the init logic all `amrex::Long` now.

atmyers (approved):

PR #1077, opened 2025-08-01
BeamMonitor: Support >2B Particles Robustly

ax3l:
Robustly support more than 2 billion particles in `BeamMonitor`.

2B+ particles per AMReX particle tile or rank is a separate topic that would need changes in AMReX first, where `numParticles` et al. default to `int` today.

### DIFF
--- a/src/elements/diagnostics/BeamMonitor.H
+++ b/src/elements/diagnostics/BeamMonitor.H
@@ -35,7 +35,7 @@ namespace detail
 
         ImpactXParticleCounter (ParticleContainer & pc);
 
-        unsigned long GetTotalNumParticles () { return m_Total; }
+        unsigned long long GetTotalNumParticles () { return m_Total; }
 
         std::vector<unsigned long long> m_ParticleOffsetAtRank;
         std::vector<unsigned long long> m_ParticleSizeAtRank;
@@ -46,7 +46,7 @@ namespace detail
         * @param[out] offset particle offset over all, mpi-global amrex fabs
         * @param[out] sum number of all particles from all amrex fabs
         */
-        void GetParticleOffsetOfProcessor (const long &numParticles,
+        void GetParticleOffsetOfProcessor (amrex::Long const &numParticles,
                                            unsigned long long &offset,
                                            unsigned long long &sum) const;
 
--- a/src/elements/diagnostics/BeamMonitor.cpp
+++ b/src/elements/diagnostics/BeamMonitor.cpp
@@ -47,10 +47,10 @@ namespace detail {
 
         for (auto currentLevel = 0; currentLevel <= pc.finestLevel(); currentLevel++)
         {
-            long numParticles = 0; // numParticles in this processor
+            amrex::Long numParticles = 0;  // numParticles on this MPI rank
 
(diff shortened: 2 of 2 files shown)


ax3l:
(Emulated) perf on CPU is the same for particle pushes :-o
Fun and unexpected!

src/elements/diagnostics/AdditionalProperties.cpp
```diff
@@ -80,7 +80,7 @@ namespace impactx::elements::diagnostics
                 const int grid_id = pti.index();
                 const int tile_id = pti.LocalTileIndex();
                 pc.DefineAndReturnParticleTile(lev, grid_id, tile_id);
-                auto np = pti.numParticles();
+                amrex::Long const np = pti.numParticles();
```
ax3l: Returns `int` in AMReX, so no need to cast up: would not help if `int` bounces at 2B local particles of the tile. https://github.com/AMReX-Codes/amrex/blob/26.06/Src/Particle/AMReX_StructOfArrays.H#L241

src/elements/diagnostics/BeamMonitor.cpp
```diff
@@ -47,10 +47,10 @@ namespace detail {
 
         for (auto currentLevel = 0; currentLevel <= pc.finestLevel(); currentLevel++)
         {
-            long numParticles = 0; // numParticles in this processor
+            amrex::Long numParticles = 0; // numParticles in this processor
```
ax3l:
```suggestion
            amrex::Long numParticles = 0;  // numParticles on this MPI rank
```

atmyers (approved):

atmyers:
Looks good to me.

src/elements/diagnostics/BeamMonitor.H
```diff
@@ -35,7 +35,7 @@ namespace detail
 
         ImpactXParticleCounter (ParticleContainer & pc);
 
-        unsigned long GetTotalNumParticles () { return m_Total; }
+        unsigned long long GetTotalNumParticles () { return m_Total; }
```
EZoni, line 38: Just curious, why couldn't this be `amrex::Long` too?

ax3l: Good question. At the moment, this is used exactly once in `openPMD::Dataset`:
https://github.com/BLAST-ImpactX/impactx/blob/26.06/src/elements/diagnostics/BeamMonitor.cpp#L239-L254

which in turn defines exactly `std::uint64_t`
https://github.com/openPMD/openPMD-api/blob/0.17.1/include/openPMD/Dataset.hpp#L34-L76

Now `long long` is guaranteed to "will have width of at least 64 bits"
https://en.cppreference.com/cpp/language/types
which matches this perfectly on any platform (in practice, always 64bits).

Could this also return directly  `std::uint64_t`? Yes, and it would be cleaner. `unsigned long long` just now also fits nicely the `m_ParticleOffsetAtRank` and `m_ParticleSizeAtRank` members, so to keep the diff small this is what I picked.

Could this use `amrex::Long`? Probably, but it sometimes uses `long` instead of `long long`, so why take the risk.
https://github.com/AMReX-Codes/amrex/blob/26.06/Src/Base/AMReX_INT.H#L19-L30

PR #1080, opened 2025-08-05
fix documentation for ExactCFbend

egstern:
The documentation for the ExactCFbend used K_normal and K_skew (capitalized) for parameter names of the function  call but the actual code needs them to be all lower case.


cemitch99 (approved):
Thanks for catching this.  This looks ready to merge once all tests have passed.

Issue #1081, opened 2025-08-05
Conda 25.07: `impactx-dashboard` not working


ax3l:
If I install ImpactX with:
```bash
conda create -c conda-forge -n impactx-25.07 impactx
conda activate impactx-25.07
```

and run
```bash
impactx-dashboard 
```

I get the error:
```
WARNING: Card 'base_section' has no doc link in DashboardDefaults.DOCUMENTATION
Traceback (most recent call last):
  File "/home/axel/micromamba/envs/impactx-25.07/bin/impactx-dashboard", line 7, in <module>
    from impactx.dashboard.__main__ import main
ImportError: cannot import name 'main' from 'impactx.dashboard.__main__' (/home/axel/micromamba/envs/impactx-25.07/lib/python3.13/site-packages/impactx/dashboard/__main__.py)
```

is something broken in out `setup.py`?

ax3l:
If I start the GUI from Jupyter, I see these issues with the lattice (in ImpactX 25.07):

<img width="1467" height="663" alt="Image" src="https://github.com/user-attachments/assets/63c5586f-f1c9-4a7d-8d5a-c9839135178b" />

ax3l:
Maybe both are already fixed, to be checked with the 25.08 release again in a day or two.

proy30:
@ax3l  Thank you for pointing this out. Created a new PR #1082.

For the second issue, I'm not seeing it on my end locally or on Jupyter. However, to quickly launch the dashboard in Jupyter, I did `pip install impactx-noacc` to get all of the dependencies quickly.

proy30:
<img width="1895" height="945" alt="Image" src="https://github.com/user-attachments/assets/cc4f5ef9-334a-4d67-bd8e-13afae00adf7" />

ax3l:
Thank you for the first fix!!

I will check again with the conda-forge build after the 25.08 release for the 2nd issue :)
https://github.com/conda-forge/impactx-feedstock/tree/main/recipe

proy30:
Fixed the second issue in https://github.com/BLAST-ImpactX/impactx/pull/1087

PR #1087, opened 2025-08-09
Dashboard: add simple fix for type extraction

proy30:
Error first noted in https://github.com/BLAST-ImpactX/impactx/issues/1081 — the **lattice configuration** section of the dashboard wasn’t pulling the correct type values for each parameter after the change in https://github.com/BLAST-ImpactX/impactx/commit/be9e504b3977864d0485bd1090f700706166d5e2 to `src/python/impactx/elements/__init__.pyi`. Instead of returning `int`, `float`, or `str` for each element parameter, it was returning `SupportInt` or `SupportFloat`.

This PR adds a simple hardcoded fix to correct the type values on the dashboard.

### DIFF
--- a/src/python/impactx/dashboard/Input/defaults_helper.py
+++ b/src/python/impactx/dashboard/Input/defaults_helper.py
@@ -159,8 +159,16 @@ class InputDefaultsHelper:
                 else:
                     parameter_type = type_and_default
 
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Oh that makes sense with the generalization of the type hints. Thank you! :)

PR #1094, opened 2025-08-14
test_charge_deposition: SP Tolerance

ax3l:
Lower the tolerance slightly for `test_charge_deposition` so it passes in SP and DP.

Isolated from #899

PR #1096, opened 2025-08-14
Python: Precision in `impactx.Config`

ax3l:
Add particle precision properties in the Python
API, similar to what we do in pyAMReX. This is useful for users to quickly find out what features are supported.

Also add details on SIMD.


ax3l:
Merging, so I can use it in #1094

PR #1097, opened 2025-08-14
Python: Load Arbitrary Particle Weight

ax3l:
Enhance the Python `add_n_particles` method to support loading particles of non-uniform particle weights. This is typically needed when loading from t-based PIC codes (EM/ES), like WarpX, which often modify particle weights for injection, ionization and density profile purposes.

X-ref: https://github.com/orgs/BLAST-ImpactX/discussions/1031 (ping Xuan Zhang @Xuan0533, Stony Brook University :) )

Check-list:
- [x] implement
- [x] document
- [x] test

### DIFF
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -379,7 +379,7 @@ Particles
       :param pt: momentum in t
       :param qm: charge over mass in 1/eV
       :param bunch_charge: total charge within a bunch in C
-      :param w: weight of each particle: how many real particles to represent
+      :param w: weight of each particle: the macroparticle charge in units of the elementary charge `e` (i.e., how many real particles to represent)
 
    .. py:method:: ref_particle()
 
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -360,9 +360,12 @@ Particles
       :param keep_mass: do not reset the reference particle mass
       :param keep_charge: do not reset the reference particle charge
 
-   .. py:method:: add_n_particles(x, y, t, px, py, pt, qm, bchchg)
+   .. py:method:: add_n_particles(x, y, t, px, py, pt, qm, bunch_charge=None, w=None)
 
       Add new particles to the container for fixed s.
+
+      Either the total charge (bunch_charge) or the weight of each
+      particle (w) must be provided.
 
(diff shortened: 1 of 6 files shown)


src/particles/ImpactXParticleContainer.cpp
```diff
@@ -304,7 +313,7 @@ namespace impactx
                 pt_arr[old_np+i] = pt_ptr[my_offset+i];
 
                 qm_arr[old_np+i] = qm;
-                w_arr[old_np+i]  = bchchg/ablastr::constant::SI::q_e/np;
+                w_arr[old_np+i]  = has_w ? w_ptr[my_offset+i] : bunch_charge.value() / ablastr::constant::SI::q_e/np;
```
ax3l: bunch_charge.value() needs to be stored in a temporary var before the kernel

docs/source/usage/python.rst
```diff
@@ -375,7 +378,8 @@ Particles
       :param py: momentum in y
       :param pt: momentum in t
       :param qm: charge over mass in 1/eV
-      :param bchchg: total charge within a bunch in C
+      :param bunch_charge: total charge within a bunch in C
+      :param w: weight of each particle: how many real particles to represent
```
cemitch99: What if we call `w` something like `macroparticle_charge`:  the macroparticle charge in units of the elementary charge `e`.

ax3l: It also scales the mass of each macro particle though :)

We usually in other PIC codes call this weight, and this is also the name we use in the openPMD standard. I think this is pretty universally understood.

But I am happy to change the doc string if that causes confusion?

cemitch99: Understood.  Let's keep the name, then.  I might just suggest a slight rewording of the doc string.

```suggestion
      :param w: weight of each particle: the macroparticle charge in units of the elementary charge `e`
```

ax3l: Would this be ok?
```suggestion
      :param w: weight of each particle: the macroparticle charge in units of the elementary charge `e` (i.e., how many real particles to represent)
```

cemitch99: Looks good to me.

cemitch99 (approved):

PR #1065, opened 2025-07-28
pybind11: 3.0.0

EZoni:
Update to pybind11 v3.0.0, following:
- [x] https://github.com/AMReX-Codes/pyamrex/pull/455
- [x] https://github.com/BLAST-WarpX/warpx/pull/6043

To-do:
- [x] Update ABLASTR branch after https://github.com/BLAST-WarpX/warpx/pull/6043 is merged.

PR #1106, opened 2025-08-20
Fix `.pyi` files after pybind11 v3

ax3l:
Fix the `impactx/impactx_pybind/__init__.pyi` file that references non-imported classes after the update to `pybind11` v3.0

Follow-up to #1065

Related: https://github.com/astral-sh/ruff/issues/19890


src/python/impactx/distribution_input_helpers.pyi
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
-import numpy as np
+
 import numpy
+import numpy as np
```
ax3l: Hm, this is an unused import and should be removed automatically...
```suggestion
```

ax3l: Stub generation unfortunately marks the import as re-export automatically:
https://docs.astral.sh/ruff/rules/unused-import/#why-is-this-bad

PR #1113, opened 2025-08-26
Envelope Tracking: No Mean or Dispersion

ax3l:
We do not support ininitalizing envelopes with a mean offset or dispersion.

Before this PR, provided means or dispersion were silently ignored

To be addressed:  Issue #1114

### DIFF
--- a/src/initialization/InitDistribution.cpp
+++ b/src/initialization/InitDistribution.cpp
@@ -224,6 +224,8 @@ namespace impactx
         std::optional<amrex::ParticleReal> intensity
     )
     {
+        using namespace amrex::literals;  // for _rt and _prt
+
         // zero out the 6x6 matrix
         CovarianceMatrix cv{};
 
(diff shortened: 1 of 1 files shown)


src/initialization/InitDistribution.cpp
```diff
@@ -246,6 +248,21 @@ namespace impactx
...
+                    distribution.m_meant  != 0.0_prt ||
+                    distribution.m_meanpx != 0.0_prt ||
+                    distribution.m_meanpy != 0.0_prt ||
+                    distribution.m_meanpt != 0.0_prt ||
+                    distribution.m_dispx  != 0.0_prt ||
+                    distribution.m_disppx != 0.0_prt ||
+                    distribution.m_dispy  != 0.0_prt ||
+                    distribution.m_disppy != 0.0_prt
+                ) {
+                    throw std::runtime_error("Cannot create envelope for distribution with non-zero means or dispersion!");
```
ax3l: @cemitch99 If you like, we could open an issue for tracking purposes (does not need to be addressed in the near-term) and we can link the issue for users here.
```suggestion
                    throw std::runtime_error("Cannot (yet) create envelope for distribution with non-zero means or dispersion! Please see: https://github.com/BLAST-ImpactX/impactx/issues/...");
```

cemitch99 (approved):
Looks good.

src/initialization/InitDistribution.cpp
```diff
@@ -246,6 +248,21 @@ namespace impactx
...
+                    distribution.m_meant  != 0.0_prt ||
+                    distribution.m_meanpx != 0.0_prt ||
+                    distribution.m_meanpy != 0.0_prt ||
+                    distribution.m_meanpt != 0.0_prt ||
+                    distribution.m_dispx  != 0.0_prt ||
+                    distribution.m_disppx != 0.0_prt ||
+                    distribution.m_dispy  != 0.0_prt ||
+                    distribution.m_disppy != 0.0_prt
+                ) {
+                    throw std::runtime_error("Cannot (yet) create envelope for distribution with non-zero means or dispersion! Please see: https://github.com/BLAST-ImpactX/impactx/issues/...");
```
```suggestion
                    throw std::runtime_error("Cannot (yet) create envelope for distribution with non-zero means or dispersion! Please see: https://github.com/BLAST-ImpactX/impactx/issues/1114");
```

PR #1118, opened 2025-08-28
Dashboard: fix flaky CI

### LINKED ISSUE #1107: Dashboard Test Sometimes Hangs
I see this specific test in `test_dashboard` hang quite often for
https://github.com/BLAST-ImpactX/impactx/blob/25.08/tests/python/dashboard/test_dashboard.py#L76-L89

Failing with
```
TimeoutError: state['total_length'] never became '3.0m' after 60 seconds (last value: '2.75m').
```
[job-logs.txt](https://github.com/user-attachments/files/21902118/job-logs.txt)

I can usually solve it with a test restart in CI, but that is of course cumbersome and does not work in unsupervised situations, eg.., when we release software in a package manager like Conda/Spack.

This seems to happen more often for the `GCC w/ SIMD w/o MPI w/ Python` test.

I have the following ideas how we could improve this test:

- Since this is a floating point comparison, should we introduce a small epsilon instead of expecting `3.0` exactly? (usually that is the best practice, and we will need this for single precision, too.)
- Does the test need a bit more time to load the GUI up, e.g., before it starts to set the lattice elements? A `0.25m` gap missing [could be the `ds1` or the `ds5` element](https://github.com/BLAST-ImpactX/impactx/blob/25.08/tests/python/dashboard/test_dashboard.py#L48-L66), maybe we can print the lattice for this value to find out what is missing? Do we need selenium to wait for the GUI to update after each `dashboard.add_lattice_element` or `dashboard.set_input(...)`?

proy30:
A series of code additions to help improve stability of dashboard CI tests and fix https://github.com/BLAST-ImpactX/impactx/issues/1107

Feature addition:
* For each of the dashboard tests, a screenshot is stored as an artifact. This will help with debugging purposes if a dashboard test fails as we will be able to see visually how the dashboard looks.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -85,7 +85,8 @@ jobs:
       with:
         name: dashboard-screenshots-macos
         path: |
-          build/bin/pytest.ImpactX.dashboard/screenshots/**/*.png
+          build/**/pytest.ImpactX.dashboard/screenshots/**/*.png
+          build/**/screenshots/**/*.png
         if-no-files-found: ignore
 
     - name: run installed python module
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -79,6 +79,15 @@ jobs:
       run: |
         ctest --test-dir build --output-on-failure -E pytest.AMReX
 
+    - name: Upload dashboard screenshots
+      if: always()
+      uses: actions/upload-artifact@v4
+      with:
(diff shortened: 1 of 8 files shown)


proy30:
@ax3l 🤞 That this solved the flakiness issue. I had tests rerun twice and both times passed all tests.

ax3l (approved):
Thank you, that looks great! :rocket: :sparkles:

PR #1119, opened 2025-08-28
Dashboard: add lattice configuration scrollbar and refine documentation warning

proy30:
This update adds a scrollbar to the lattice configuration panel, which now appears automatically when the number of elements exceeds a defined height threshold.

It also refines the warning system for missing documentation. Previously, any section using the shared card component would trigger the warning, even if documentation wasn’t required. For example, the lattice statistics section was incorrectly flagged. With this change, the warning only applies to input sections that explicitly require documentation.


ax3l (approved):
Looks great! :)

PR #1122, opened 2025-08-29
Copy the Reference Particle

ax3l:
Use this if you want to do an operation that modifies the reference particle, but you want to avoid modifying the particle container that it is assigned to.

### Example 1: Test Case

See test case in the PR.
Copies a reference particle, advances it, makes sure it does not change the original one.

### Example 2: Plotting a lattice by doing a loop over all elements.

In that loop, one wants to do always plot the next element, then push a reference particle through it, then continue with the next element. This is important to get the global coordinates of the next element and to properly treat accelerating elements (e.g., because energy changes the bending radius `rc` of following elements).

Global floor plan visualization pseudo-code:
```py
def plot_element(element, ref_vis)
    element_params = get_element_params(element, ref_vis)

    if thin:
    elif straight:
    elif arch/wedge:


def visualize():
    # ...
    ref_vis = sim.particle_container().ref_particle().copy()

    for element in sim.lattice:
        plot_element(element, ref_vis)
        push(ref_vis, element)
```

### DIFF
--- a/src/elements/Programmable.cpp
+++ b/src/elements/Programmable.cpp
@@ -28,7 +28,7 @@ namespace impactx::elements
             push_all(pc, *this, step, period, m_threadsafe);
         }
         else {
-            BL_PROFILE("impactx::Push::Programmable");
+            BL_PROFILE("impactx::push::Programmable");
             m_push(&pc, step, period);
         }
     }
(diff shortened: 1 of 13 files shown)


cemitch99 (approved):
Great.

PR #1134, opened 2025-09-10
Generalize logic for setting default tile size for OMP threading.

### LINKED ISSUE #1132: Perlmutter CPU: No Good Tile Size
When running on Perlmutter (NERSC) by default, with 1 MPI rank and 128 OMP threads, we get this error:
```
amrex::Abort::0::ImpactParticleContainer::prepare() could not find good tile size for the number of OpenMP threads. Consider lowering the number of OpenMP threads via the environment variable OMP_NUM_THREADS. !!!
```

Regression of #862 

## Reproducer:

Any, e.g. [FODO example](https://impactx.readthedocs.io/en/latest/usage/examples/fodo/README.html#run): `impactx input_fodo.in`

atmyers:
Fix #1132. 

This increases the number of threads available by default to 512 by changing the strategy to consider all three dimensions.

Also, if a good tile_size can't be set for the default number of threads, this throws a warning and reduces the number of threads OMP will use.

### DIFF
--- a/src/particles/ImpactXParticleContainer.cpp
+++ b/src/particles/ImpactXParticleContainer.cpp
@@ -156,27 +156,40 @@ namespace impactx
         // Then we try to reduce the tile size in the (z, y, x) directions alternately
         // until there are enough tiles for the number of threads.
 
-        const auto& ba = ParticleBoxArray(lid);
-        tile_size = ba[gid].size();
-        auto n_logical = numTilesInBox(ba[gid], true, tile_size);
+        // if the user has set particles.tile_size, do not override this choice
+        // unlike particles.do_tiling, we do not add this to the table in init,
+        // so if it's there the user set it
+    bool user_set = false;
+    {
+        amrex::ParmParse pp_particles("particles");
+            amrex::Vector<int> tilesize(AMREX_SPACEDIM);
+        user_set = pp_particles.queryarr("tile_size", tilesize, 0, AMREX_SPACEDIM);
+    }
 
+    const auto& ba = ParticleBoxArray(lid);
+    auto n_logical = numTilesInBox(ba[gid], true, tile_size);
+
+    if (!user_set) {
+        tile_size = ba[gid].size();
         int ntry = 0;
         constexpr int max_tries = 10;
         while ((n_logical < nthreads) && (ntry++ < max_tries)) {
-            int idim = 2 - (ntry % 3);  // alternate between (2, 1, 0)
-            if (tile_size[idim] == 1) { continue; }
-            tile_size[idim] /= 2;
-            n_logical = numTilesInBox(ba[gid], true, tile_size);
+        int idim = 2 - (ntry % 3);  // alternate between (2, 1, 0)
+        if (tile_size[idim] == 1) { continue; }
+        tile_size[idim] /= 2;
+        n_logical = numTilesInBox(ba[gid], true, tile_size);
         }
+    }
 
         if (n_logical < nthreads) {
         ablastr::warn_manager::WMRecordWarning(
-        "ImpactxParticleContainer::prepare",
-        "Could not find good tile size for the number of OpenMP threads. "
-        "Lowering the number of OpenMP threads used compared to the environment variable OMP_NUM_THREADS. "
-        "This may result in poorer than expected performance. "
-        "You may want to try increasing the blocking factor and max grid size.",
-        ablastr::warn_manager::WarnPriority::medium
+            "ImpactxParticleContainer::prepare",
+            "Could not find a good tile size for the requested number of OpenMP threads. "
+            "The number of threads is " + std::to_string(nthreads) + ", while the number "
+            "of available tiles is " + std::to_string(n_logical) + ". Using a lower number "
+            "of threads instead. This may result in poorer than expected performance. "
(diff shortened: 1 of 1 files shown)


ax3l (commented):
That looks great!

Shall we additionally mabe reduce the number of threads and warn, instead of aborting on `if (n_logical < nthreads)`? That could maybe make the code easier to use for arbitrary configurations, e.g., when we run with a user-specified grid for space charge?

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -152,30 +152,35 @@ namespace impactx
...
+        ablastr::warn_manager::WMRecordWarning(
+                "Could not find good tile size for the number of OpenMP threads. "
+                "Lowering the number of OpenMP threads used compared to the environment variable OMP_NUM_THREADS. "
+        "This may result in poorer than expected performance. "
+        "You may want to try increasing the blocking factor and max grid size.",
+        ablastr::warn_manager::WarnPriority::medium
+        );
+
+#if defined(AMREX_USE_OMP)
+        omp_set_max_threads(n_logical);
```
atmyers:
```suggestion
        omp_set_num_threads(n_logical);
```

atmyers:
> That looks great!
> 
> Shall we additionally mabe reduce the number of threads and warn, instead of aborting on `if (n_logical < nthreads)`? That could maybe make the code easier to use for arbitrary configurations, e.g., when we run with a user-specified grid for space charge?

Sure, see new commits.

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -152,30 +152,36 @@ namespace impactx
...
+        "ImpactxParticleContainer::prepare",
+        "Could not find good tile size for the number of OpenMP threads. "
+        "Lowering the number of OpenMP threads used compared to the environment variable OMP_NUM_THREADS. "
+        "This may result in poorer than expected performance. "
+        "You may want to try increasing the blocking factor and max grid size.",
+        ablastr::warn_manager::WarnPriority::medium
+        );
+
+#if defined(AMREX_USE_OMP)
+        omp_set_num_threads(n_logical);
```
ax3l: Two small changes we discussed:
- please include the `nthreads` number before and after we lower it (`n_logical`) in the warning message to the user
- please fix the alignment

Wording could be something like
```suggestion
        ablastr::warn_manager::WMRecordWarning(
        "ImpactxParticleContainer::prepare",
        "Could not find good tile size for the number of OpenMP threads (add: nthreads). "
        "Consider lowering the number of OpenMP threads via the environment variable OMP_NUM_THREADS. "
        "Will now lower it to the number of tiles, n_logical."
        "This may result in poorer than expected performance. "
        "You may want to try increasing the blocking factor and max grid size to create more tiles for OpenMP.",
        ablastr::warn_manager::WarnPriority::medium
        );

#if defined(AMREX_USE_OMP)
        omp_set_num_threads(n_logical);
```
In the wording, note that OMP_NUM_THREADS is not the only way we set the number of OMP threads. We pick a default by inspecting the system and also have an API to set it. `OMP_NUM_THREADS` is common and honored, unless the API is used to set num threads.

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -152,30 +152,36 @@ namespace impactx
...
-        // appropriately here. The tiles start as (very large number, 8, 8)
-        // in (x, y, z). So we try to reduce the tile size in the y and z
-        // directions alternately until there are enough tiles for the number of threads.
+        // appropriately here. We start by setting the tile_size to the size of box 0.
+        // Then we try to reduce the tile size in the (z, y, x) directions alternately
+        // until there are enough tiles for the number of threads.
 
         const auto& ba = ParticleBoxArray(lid);
+        tile_size = ba[gid].size();
         auto n_logical = numTilesInBox(ba[gid], true, tile_size);
```
ax3l: As discussed, we want to check that there is no user-specified tile size in ParmParse before adjusting tiles.

ax3l:
@atmyers if you rebase against `development` the docs build error will go away #1136

ax3l (approved):
Thank you, Andrew! :rocket: :sparkles:

ax3l:
Checked on Perlmutter and the bug is gone :)

src/particles/ImpactXParticleContainer.cpp
```diff
@@ -152,30 +152,58 @@ namespace impactx
...
-        if (n_logical < nthreads) {
-            amrex::Abort("ImpactParticleContainer::prepare() "
-                "could not find good tile size for the number of OpenMP threads. "
-                "Consider lowering the number of OpenMP threads via the environment variable OMP_NUM_THREADS."
+        if (n_logical < nthreads)
+        {
+            ablastr::warn_manager::WMRecordWarning(
+                "ImpactXParticleContainer::prepare",
+                "Could not find a good tile size for the requested number of "
+                "OpenMP threads. The number of threads is " +
```
ax3l: For clarity, this first sentence (and the last one) could be different when `particles.tile_size` is user-defined.

ax3l: Improved :)

ax3l (approved):

PR #1145, opened 2025-09-16
Fix CSR error message in reference particle mode.

cemitch99:
This PR fixes a bug in the error message related to CSR when running in reference particle mode using `track_reference(ref)`.

It affects only this mode, and is not relevant when using `track_particles` or `track_envelope`.


ax3l (approved):

ax3l:
@cemitch99 Thanks for catching this!

PR #1162, opened 2025-09-29
Topic gauss3d

qianglbl:
updated the Gauss3dPush.cpp and added 2p5D SC model with transverse Gaussian distribution.


cemitch99:
Hi @qianglbl .  Thanks a lot for contributing this.  If I understand correctly, this PR is intended to correct the errors introduced into the Gauss3D solver (that were discussed last week), and then also adds the 2.5D version.  Is this correct? 

I think the conflicts above occur because you pushed to the already-existing `topic-gauss3d` branch (which was previously merged).  I would suggest changing the name of your local branch (to create a new branch), and opening the PR using that branch instead.  What do you think?

qianglbl:
Sure, I will create a separate branch. Thanks.

PR #1163, opened 2025-09-30
Topic gauss2p5d

qianglbl:
fixed Gauss3dPush and added 2.5D SC with transverse Gaussian distribution.


src/particles/spacecharge/Gauss2p5dPush.H
```diff
@@ -0,0 +1,40 @@
...
+#include <string>
+
+
+namespace impactx::particles::spacecharge
+{
+    /**------------------------
+     *
+     * This calculates the space charge fields from a 3D Gaussian distribution
+     * with respect to particle position from the following paper.
+     * "Two-and-a-half dimensional symplectic space-charge slover"
```
cemitch99, line 25: This citation looks incomplete.

cemitch99 (commented):
Thanks a lot!  The only failing tests occur on GPU, due to a host-device error that we can fix.  Do you have a test in mind for the Gauss2p5d case?  We can exchange some ideas this afternoon.

src/particles/spacecharge/Gauss3dPush.cpp
```diff
@@ -73,27 +77,35 @@ namespace impactx::particles::spacecharge
             // Simpson's rule integration
 
             // Ex
-            amrex::ParticleReal const fx = f1 / f2x;
+            //amrex::ParticleReal const fx = f1 / f2x;
+            fx = f1 / f2x;
```
ax3l, line 81: With the current implementation, the `fx` value would only be the last value of `i=nint`, so one could calculate it outside of the loop.

src/particles/spacecharge/Gauss3dPush.cpp
```diff
@@ -73,27 +77,35 @@ namespace impactx::particles::spacecharge
...
+            //amrex::ParticleReal const fz = f1 / f2z;
+            fz = f1 / f2z;
             if (i%2 == 0)
                 sum0ez += 2_prt * fz * h;
             else
                 sum0ez += 4_prt * fz * h;
         }
 
+        //end point correction
+        sum0ex -= fx*h;
```
ax3l, line 105: It confuses me that the `fx` used here is only using the last value of `i=nint` for the calculation of `fx`.

ax3l: Ah, got it. ok end-point of the integration.

src/particles/spacecharge/Gauss3dPush.cpp
ax3l: Ah, got it. ok end-point of the integration.

ax3l:
Continued in #1166

PR #1172, opened 2025-10-13
Function Hooks (Callbacks) for Tracking Loop

### LINKED ISSUE #1171: Turn-Dependent Element Changes
Not all accelerator lattices are statically over tracking time and we need to expose ways to change parameters of elements over, e.g., turns in a ring.

Application: The Booster RF voltage varies during the acceleration cycle with a corresponding change in the phase to follow the momentum ramp.

Possible approach: expose a callback function per turn.

cc @egstern @cemitch99

ax3l:
Expose a callback function per turn/element/slice that allows to modify the simulation, e.g., to manipulate the lattice.

[Here is a preview of the docs with an example.](https://impactx--1172.org.readthedocs.build/en/1172/usage/examples/turn_update/README.html#run)
And here are the [available hooks](https://impactx--1172.org.readthedocs.build/en/1172/usage/python.html#impactx.ImpactX.hook).

Close #1171

---

- [x] We could refactor this a bit more and, as in WarpX, do something like `sim.hooks["name"]` to have it easier adding more hooks later on.
- [x] The `sim.hooks` element could become its own type, deriving from the current type, and check instantaneous on assignment if the hook name is valid. That way we can throw clear errors if a hook does not exist.
- [x] Add a `sim.tracking_element` that points to the current/next element? Useful to write `if element == ...` in `before/after_element` hooks.

### DIFF
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -303,6 +303,44 @@ Collective Effects & Overall Simulation Parameters
 
       :param ref: the reference particle (object from :py:class:`impactx.RefPart`)
 
+   .. py:property:: hook
+
+      User-defined function hooks that are called, e.g, during tracking.
+      Supported hook locations names are:
+
+      * ``"before_period"``: before each period (e.g., turn or channel period)
+      * ``"after_period"``: after each period (e.g., turn or channel period)
+      * ``"before_element"``: before each element is entered
+      * ``"after_element"``: after each element is exited
+      * ``"before_slice"``: before each element slice
+
+      Example: Function hook that can be called before each turn (sim):
+
+      .. code-block:: python3
+
+         def hook_before_period(sim):
+             beam = sim.particle_container()
+             turn = sim.tracking_period
+             # Example: you could now manipulate elements in sim.lattice
+             #          for the next turn.
+
+         sim.hook["before_period"] = hook_before_period
+
+   .. py:property:: tracking_step
+
+      For tracking hooks/callbacks, a global step of the simulation.
+
+      A state of internal simulation steps, increments also for space charge slice steps in elements.
+      We start in "step 0" (initial state).
+
+   .. py:property:: tracking_period
+
+      For tracking hooks/callbacks, the period in the lattice (e.g., turn or channel period).
+
+   .. py:property:: tracking_element
+
+      For tracking hooks/callbacks, the current lattice element.
+
    .. py:method:: resize_mesh()
 
       Resize the mesh :py:attr:`~domain` based on the :py:attr:`~dynamic_size` and related parameters.
--- a/examples/turn_update/run_turn_update.py
+++ b/examples/turn_update/run_turn_update.py
@@ -1,0 +1,100 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022-2023 ImpactX contributors
+# Authors: Chad Mitchell, Axel Huebl
+# License: BSD-3-Clause-LBNL
+#
+# -*- coding: utf-8 -*-
+
+from impactx import ImpactX, distribution, elements
+
+sim = ImpactX()
+
+# set numerical parameters and IO control
+sim.space_charge = False
+sim.slice_step_diagnostics = True
+sim.tiny_profiler = False
+
+# domain decomposition & space charge mesh
+sim.init_grids()
+
+# load a 5 GeV electron beam with an initial
+# normalized transverse rms emittance of 1 um
+kin_energy_MeV = 2.0e3  # reference energy
+bunch_charge_C = 1.0e-9  # used with space charge
+npart = 10000  # number of macro particles
+
+#   reference particle
+ref = sim.particle_container().ref_particle()
+ref.set_charge_qe(-1.0).set_mass_MeV(0.510998950).set_kin_energy_MeV(kin_energy_MeV)
+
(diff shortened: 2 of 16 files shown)


src/Hooks.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "ImpactX.H"
+
+#include <string>
+
+
+namespace impactx
+{
+    void ImpactX::call_hook (std::string const & name)
+    {
+        if (m_hook.contains(name))
```
ax3l: This is C++20. We can use `find` or `count`:
https://en.cppreference.com/w/cpp/container/unordered_map/find.html
https://en.cppreference.com/w/cpp/container/unordered_map/count.html

egstern:
What context does the hook execute in? Can it retain state? For instance, could the hook keep track of the previous N mean s positions?

ax3l:
Hi @egstern,

The context is the calling Python process, so you can be fully stateful. The logic is: Python process owns the simulation, calls into C++ for some computation and C++ can now call back (as in the `Programmable` element we have) into the same Python process and its context, with access to everything in it.

You can let it hook into any Python function with the right signature (one argument: `sim` of type `ImpactX` for the simulation state). The simplest way is this to store additional state in a global variable in your Python script.

Alternatively and slightly more involved, one can call a (e.g., static) method of a Python class that has he states as static member variables. One can even use regular classes and methods, as we do in the [PyTorch staging example](https://impactx.readthedocs.io/en/latest/usage/examples/pytorch_surrogate_model/README.html). (I have to check again, but I think you can also already store additional state directly on the `sim` variable as long as you create the extra attributes to it before the hook is called, so there are many options to carry states around.)

But I recommend to just go with a global variable first, the rest is syntactic sugar we can do later.

```py
# ...

my_old_s = []

def hook_before_period(sim):
    global my_old_s

    # turn = sim.tracking_period
    beam = sim.particle_container()
    ref = beam.ref_particle()
    # rbc = beam.beam_moments()

    s = ref.s
    # t = ref.t

    my_old_s += [s]
    # ...


sim.hook["before_period"] = hook_before_period

# ...
```

docs/source/usage/python.rst
```diff
@@ -303,6 +303,44 @@ Collective Effects & Overall Simulation Parameters
...
+   .. py:property:: hook
+
+      User-defined function hooks that are called, e.g, during tracking.
+      Supported hook locations names are:
+
+      * ``"before_period"``: before each period (e.g., turn or channel period)
+      * ``"after_period"``: after each period (e.g., turn or channel period)
+      * ``"before_element"``: before each element is entered
+      * ``"after_element"``: after each element is exited
+      * ``"before_slice"``: before each element slice
```
ax3l, line 315: To discuss in review:
Maybe the naming `before/end` is not ideal, because we want to characterize that we "just entered" or are about to "just exist" a turn/element/slice.

Potentially better names:
- `start_...` / `end_...`
- `enter_...` / `exit_...`

Let me know what you think.

src/tracking/envelope.cpp
```diff
@@ -125,14 +130,22 @@ namespace impactx
         int num_periods = 1;
         amrex::ParmParse("lattice").queryAddWithParser("periods", num_periods);
 
-        for (int period=0; period < num_periods; ++period)
+        for (period=0; period < num_periods; ++period)
         {
+            // optional, user-defined function call
+            m_tracking_state.m_element = &m_lattice.front();
```
ax3l, line 136: Note: We intentionally assign the next element already as we enter the period hook, because it is logical to know the element coming next at this point.
We assign it again in the element loop below.

(Same for the other two tracking loops.)

examples/turn_update/run_turn_update.py
```diff
@@ -0,0 +1,100 @@
...
+    print(
+        f"  Current element name: {element.name} with ds={element.ds:.2f}", flush=True
+    )
+
+    if element.name != "monitor":
+        print(f"  Drift ds is: {element.ds:.2f}m", flush=True)
+
+
+sim.hook["before_period"] = hook_before_period
+sim.hook["before_element"] = hook_before_element
```
ax3l, line 94: To discuss during review:
The quick-and-dirty design here allows to register 1 function per hook (and then branch inside that function, as needed).

We could instead use a syntax like:
```suggestion
sim.hook.add("before_period", hook_before_period)
sim.hook.add("before_element", hook_before_element)
```

and then we could add multiple functions at a hook location (called in the order they were registered).

To discuss if this is needed.

docs/source/usage/python.rst
egstern: @ax3l I actually like the before/after naming. It is very definite.

cemitch99 (approved):
Nice feature.

examples/turn_update/run_turn_update.py
cemitch99: Either syntax looks fine to me.  The additional flexibility of the second option is an advantage, but no reason to delay this feature if the existing implementation has everything needed by @egstern .

ax3l: I am on the fence to add multiple callbacks on he same hook, because it introduces some opaqueness on which one is called first. (We also did not add that in WarpX.) I would leave it out for now, once can always add this in user-code in the callback of each hook.

PR #1173, opened 2025-10-13
Envelope: Silence Warning w/o Space-Charge

### LINKED ISSUE #1170: Envelope Tracking: Suppress repeating messages "Warning: Space charge is off by default."
When running a simulation without space charge, the message "Warning: Space charge is off by default." is printed for every single step in the simulation.

This is coming from this line in `tracking/envelope.cpp`:
https://github.com/BLAST-ImpactX/impactx/blob/35d4762d0463392acf45c15621ec620d592bef2e/src/tracking/envelope.cpp#L165

Would it be possible to e.g. wrap this in `if (verbose > 0){}`, in order to cut down on the noise? Or make sure it is only outputted once per simulation?

ax3l:
Consistent with the other two tracking modes, we only print options upfront and do not warn every simulation step if space charge is not used.

Fix #1170

### DIFF
--- a/src/tracking/envelope.cpp
+++ b/src/tracking/envelope.cpp
@@ -161,8 +161,6 @@ namespace impactx
                     {
                         // push Covariance Matrix in 3D space charge fields
                         envelope::spacecharge::space_charge3D_push(ref, cm, intensity, slice_ds);
-                    } else {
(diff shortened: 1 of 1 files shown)


ax3l:
Merging for Oct release :)

kyrsjo:
Thanks! :)

PR #1178, opened 2025-10-13
Fix `BeamMonitor`: Delay Open

### LINKED ISSUE #963: Revisit (MPI) Init/Shutdown Control
In optimizer workflows (our xopt example, ABEL), we run ImpactX many times in the same process after each other.

As of 25.05, there seem to be still foot guns left to break mpi init/finalize and even non MPI runs in repeated runs, that we need to make cleaner. Our xopt example seems to work well, but deviating from its structure can cause issues.

Cc John Wieland (Mail from May 12, 2025) and Carl Lindstroem (ABEL ImpactX PR comments I added in early 2025)

### LINKED ISSUE #1169: Random jupyter kernel crashes when using envelope tracking for matching
Hello,
@aellings and I have both ran into a problem that when using ImpactX from Jupyter, especially when running envelope tracking inside `scipy.optimize.minimize` to match optics to achieve certain Twiss parameters, it sometimes crashes. These jobs require repeated initialization of Impactx, followed by calls to `track_envelope()`, and reading of the time series output to determine the resulting parameters this round.

This crash manifests itself as either a sudden kernel crash, or that the notebook cell just stops doing anything (just looking as it is running from Jupyter) but CPU being at zero. It is not deterministic, and it does not seem to be related to resource exhaustion. Sometimes running the same cell twice (or three, or four...) times will trigger the crash. We have not been able to reproduce the problem from regular `python` scripts (even when copy pasting in the same code, and running the regularly-crashing code in a loop in order to mimic the re-running of cells). We have seen the problem on both Linux, mac, and WSL. First observation was earlier this summer (with whatever version was current then), last was yesteday (with my self-compiled version on Linux and Abélia's conda installed version on Mac).

I'm a bit unsure how to attack this since I cannot reproduce it without jupyter; normally my go-to would be Valgrind...

Have you experienced something similar? 

One possibly related issue we've ran into is that on Windows (not WSL), it looks like it is not properly closing some of the files in the `diags` folder, resulting in a crash when cleaning up this folder (because on Windows you cannot delete a file that is not fully closed). But when I've exhausted the number of open files on Linux before, that has been a very predictable thing...

ax3l:
`BeamMonitor` opens the openPMD series now on first output. This simplifies usage dramatically, now allowing to:

* create the lattice (monitor elements) before the `sim` is even initialized
* creating a monitor element and not using it in a lattice

Fix #963
Fix #1169

### DIFF
--- a/src/elements/diagnostics/BeamMonitor.H
+++ b/src/elements/diagnostics/BeamMonitor.H
@@ -176,8 +176,12 @@ namespace detail
         finalize ();
 
     private:
+        /** Open the openPMD series (on first access) */
+        void open ();
+
         std::string m_series_name; //! openPMD filename
         std::string m_OpenPMDFileType; //! openPMD backend: usually HDF5 (h5) or ADIOS2 (bp/bp4/bp5) or ADIOS2 SST (sst)
+        std::string m_encoding; //! openPMD iteration encoding: "v"ariable based, "f"ile based, "g"roup based (default)
         std::any m_series; //! openPMD::Series that holds potentially multiple outputs
         int m_step = 0; //! global step for output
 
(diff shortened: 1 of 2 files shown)


EZoni (approved):
This looks good to me, but I'm not familiar with the use case, so I'll approve but let someone else merge.

PR #1181, opened 2025-10-14
Python: Improve `KnownElementsList`

ax3l:
Add per-element by-reference access for manipulation.

Isolated from #1172


ax3l:
Merging to simplify #1172

PR #1184, opened 2025-10-15
Update PALS to 0.2.0

EZoni:
Update the PALS requirement to use the latest release: https://pypi.org/project/pals-schema/0.2.0/.

### DIFF
--- a/src/python/impactx/extensions/KnownElementsList.py
+++ b/src/python/impactx/extensions/KnownElementsList.py
@@ -29,7 +29,7 @@ def load_file(self, filename, nslice=1):
         return
 
     elif extension_inner == ".pals":
-        from pals_schema.BeamLine import BeamLine
+        from pals.BeamLine import BeamLine
 
         # examples: fodo.pals.yaml, fodo.pals.json
         with open(filename, "r") as file:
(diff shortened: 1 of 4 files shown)


EZoni:
I see pre-commit.ci complains about things like
```
F821 Undefined name `typing`
  --> src/python/impactx/extensions/KnownElementsList.pyi:44:43
   |
42 |     def __repr__(self): ...
43 |     def __str__(self): ...
44 |     def count_by_kind(self, kind_pattern: typing.Union[str, type]) -> int:
   |                                           ^^^^^^
45 |         """
46 |         Count elements of a specific kind in the filtered list.
   |
```
though I don't understand why this shows up only now in this PR.

I don't see any change related to this - `typing` was undefined already, no?

src/python/impactx/extensions/KnownElementsList.pyi
```diff
@@ -11,6 +11,8 @@ from __future__ import annotations
 
 import os as os
 import re as re
+from types import NoneType
+from typing import List, Tuple, Union
```
EZoni: No idea why this would be caught in this PR.

src/python/impactx/extensions/KnownElementsList.py
```diff
@@ -62,20 +62,20 @@ def from_pals(self, pals_beamline, nslice=1):
 
     https://github.com/campa-consortium/pals-python
     """
-    from pals_schema.DriftElement import DriftElement
-    from pals_schema.QuadrupoleElement import QuadrupoleElement
+    from pals.Drift import Drift
+    from pals.Quadrupole import Quadrupole
```
ax3l, line 65: Does that work?
```suggestion
    from pals import Drift, Quadrupole
```

src/python/impactx/extensions/KnownElementsList.pyi
ax3l: Ah, I saw an issue updating stubs earlier today from my last PR... did not have time to check yet

EZoni: I could fix these (and the others missing) in a separate PR, if that sounds good to you.

EZoni: Resolved in #1186.

src/python/impactx/extensions/KnownElementsList.py
EZoni: I think it doesn't work yet, I think we need to address https://github.com/campa-consortium/pals-python/issues/29 first.

ax3l: Ah, got it! Ok, thanks for tracking!

ax3l (approved):

PR #1185, opened 2025-10-16
Fix RFCavity edge case

cemitch99:
In the special "edge case" when the local element longitudinal coordinate z in an RFCavity lies at or beyond the endpoint defined by the element length, the integrated on-axis field (callled internally `ezint`) contains a special treatment.  This special edge case previously contained a bug, which is repaired here.

### DIFF
--- a/src/elements/RFCavity.H
+++ b/src/elements/RFCavity.H
@@ -408,7 +408,7 @@ namespace RFCavityData
             }
             else  // endpoint of the RF, outsize zlen
             {
-               efieldint = std::copysign(z, z)*zmid*0.5_prt*cos_data[0];;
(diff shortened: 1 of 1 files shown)


src/elements/RFCavity.H
```diff
@@ -408,7 +408,7 @@ namespace RFCavityData
             }
             else  // endpoint of the RF, outsize zlen
             {
-               efieldint = std::copysign(z, z)*zmid*0.5_prt*cos_data[0];;
+               efieldint = std::copysign(zmid, z)*0.5_prt*cos_data[0];;
```
cemitch99:
```suggestion
               efieldint = std::copysign(zmid, z)*0.5_prt*cos_data[0];
```

ax3l (approved):
ty!!

PR #1186, opened 2025-10-16
Fix Python import errors

EZoni:


ax3l (approved):
Thank you!

src/python/impactx/extensions/KnownElementsList.pyi
```diff
@@ -11,6 +11,8 @@ from __future__ import annotations
 
 import os as os
 import re as re
+import typing
+from types import NoneType
```
ax3l, line 15: Since the `.pyi` file are auto-generated, I wonder if we should instead update the `.py` file... @EZoni 

To check.

EZoni: Yes, it seems that it did not work: https://results.pre-commit.ci/run/github/387853787/1760602848.eWQFoT3TSSq1GHV_UebehA.

However, some of these import errors seem to be already fixed in the `.py` file, e.g.,
https://github.com/BLAST-ImpactX/impactx/blob/80b4cfd625b9af47f4846d0fa3a6cada459cf2aa/src/python/impactx/extensions/KnownElementsList.py#L11
and yet we get errors like 
```
F821 Undefined name `typing`
  --> src/python/impactx/extensions/KnownElementsList.pyi:44:43
   |
42 |     def __repr__(self): ...
43 |     def __str__(self): ...
44 |     def count_by_kind(self, kind_pattern: typing.Union[str, type]) -> int:
   |                                           ^^^^^^
45 |         """
46 |         Count elements of a specific kind in the filtered list.
   |
```
when the `.pyi` files are generated.

EZoni: An LLM is telling me the following:

> You cannot fix it by modifying the .py file. pybind11-stubgen doesn't read your .py source files at all. Instead, it:
> - Imports the compiled module (the C++ extension built by pybind11)
> - Introspects it at runtime using Python's inspect module
> - Generates .pyi stubs based on what it finds in the compiled module
>
> So any imports or code in your .py files are completely ignored by pybind11-stubgen.


If this is true the fix should be added ad hoc in the bash script that updates the stub files, e.g., after
https://github.com/BLAST-ImpactX/impactx/blob/80b4cfd625b9af47f4846d0fa3a6cada459cf2aa/.github/update_stub.sh#L19-L21

EZoni: Unless the problem is with the fact that we do things like `from typing import Union` and then use `typing.Union` instead of simply `Union` - not sure if the `typing` in `typing.Union` can be recognized if we don't have `import typing` but only the direct import `from typing import Union`.

EZoni: Well, this is only in the `.pyi` file actually, we do the right thing in the `.py` file, if I'm not mistaken.

EZoni: Maybe then we could try reverting to `import typing` and then `typing.Union`, etc., in the `.py` file and see if that makes things better. Maybe pybind11-stubgen struggles with `from typing import Union` but not with `import typing`...?

PR #1189, opened 2025-10-17
Element Names: All, `None`

ax3l:
It is easier for user-facing workflows if `.name` exists for all elements, even `Empty`.

In Python, instead of throwing an exception we return the `None` value for elements without a user-provided name, which simplifies user-logic. In this case `.has_name` is `False`.

Isolated from #1172


cemitch99 (approved):
Looks good to me.

PR #1040, opened 2025-07-11
MPL: `KnownElementsList.plot_survey()`

ax3l:
Quick plot of a lattice over `s`.

Implement https://github.com/BLAST-ImpactX/impactx/issues/887#issuecomment-3059149253

## Example
```py
import matplotlib.pyplot as plt

sim.lattice.plot_survey(ref=ref)
plt.show()
```
Figure_1
HTU Beamline: #1042 

## To Do

- [x] add a legend
- [x] add a test
- [x] update the bends to use the sign of `rc`
- [x] rename to `plot_survey`
- [x] solve discrepancies to HTU plot (e.g, quad +/- is inverted?): charge-to-mass ratio of ref particle influences focusing/defocusing
- [x] user-facing doc

## Follow-Up

- [ ] make color-palette configurable for users (store profiles in a common place, e.g., named "cern-lhc", "lbnl-als", "slac-lcls", "lbnl-htu", ...)
- [ ] improve plotting of combined function elements: plot on top of each other with correct colors & different heights & opacities
- [ ] sneakily remember the `ref` particle if the lattice is crated from `sim.lattice`

PR #1195, opened 2025-10-21
[WIP] Compensated Sum for Global Coordinates

ax3l:
Long sums, e.g., `s` of the reference particle, suffer dramatically from floating point precision errors as large and small terms are added in succession thousands to millions of times. Using a [compensated sum](https://en.wikipedia.org/wiki/Kahan_summation_algorithm#Further_enhancements) can keep these numerical errors in check.

This implements, with some vibe coding assistance, the 2nd-order method of Klein (2006) that is linked on Wikipedia. The cost is that `s` is now of the size of 3 instead of 1 double, e.g., when copying to kernels. We could also use a first order method, which only needs 2 doubles (e.g., Neumaier (1974)).

The latter might be worth considering if we need to use compensated sums for further properties, e.g., x,y,z,t(,px,py,pz,pt) of the reference particle...?

Originally inspired by #1183, but needed at many locations in ImpactX, e.g., existing survey plots #1040, inserting elements #877 #1095 etc. Exposed to Python and used there, too.

## To Do

- [ ] define a test, tracking errors in `s`, `x`, `pt`, `t`, etc. of the reference particle -- .e.g, LHC size and beam lifetime
- [ ] decide if 1st (Neumann) or 2nd order (Klein) is sensible to use

### DIFF
--- a/.github/workflows/tooling.yml
+++ b/.github/workflows/tooling.yml
@@ -39,7 +39,7 @@ jobs:
         OMPI_CXX: clang++
         MPICH_CC: clang
         MPICH_CXX: clang++
-        CXXFLAGS: "-Werror -Wno-error=pass-failed -fsanitize=address,undefined -shared-libsan -Wno-self-assign-overloaded"
+        CXXFLAGS: "-Werror -Wno-error=pass-failed -Wno-error=self-assign-overloaded -fsanitize=address,undefined -shared-libsan"
         LDFLAGS: "-fsanitize=address,undefined -shared-libsan -fno-omit-frame-pointer"
         # Note: Clang<=16 triggers on pybind11 code a false-positive of -Wself-assign-overloaded
         #   https://github.com/llvm/llvm-project/issues/42469
--- a/.github/workflows/tooling.yml
+++ b/.github/workflows/tooling.yml
@@ -39,8 +39,10 @@ jobs:
         OMPI_CXX: clang++
         MPICH_CC: clang
         MPICH_CXX: clang++
-        CXXFLAGS: "-Werror -Wno-error=pass-failed -fsanitize=address,undefined -shared-libsan"
+        CXXFLAGS: "-Werror -Wno-error=pass-failed -fsanitize=address,undefined -shared-libsan -Wno-self-assign-overloaded"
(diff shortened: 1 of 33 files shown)


src/elements/diagnostics/BeamMonitor.cpp
```diff
@@ -248,13 +248,13 @@ namespace detail {
         auto d_ui = io::Dataset(dtype_ui, {np});
 
         // openPMD 1.* needs "seconds" here, but we fake it as "s"
-        iteration.setTime(ref_part.s);
+        iteration.setTime(ref_part.s.value());
```
ax3l, line 251: Interesting. Should we just use
```suggestion
        iteration.setTime(ref_part.t.value());
```
?

(Should be a separate PR, just noticed here...)

src/particles/ReferenceParticle.H
```diff
@@ -19,6 +19,8 @@
 #include <AMReX_REAL.H>
 #include <AMReX_SmallMatrix.H>
 
+#include <CompensatedReal.H>
```
ax3l: Since this is an ImpactX file, it should be the first import and
```suggestion
#include "CompensatedReal.H"
```

src/particles/ReferenceParticle.H
```diff
@@ -66,7 +68,7 @@ namespace impactx
             auto old_mass = mass;
             auto old_charge = charge;
 
-            RefPart zero{};
+            RefPart zero = RefPart();
```
ax3l:
```suggestion
            RefPart zero;
```

.github/workflows/tooling.yml
```diff
@@ -39,8 +39,10 @@ jobs:
         OMPI_CXX: clang++
         MPICH_CC: clang
         MPICH_CXX: clang++
-        CXXFLAGS: "-Werror -Wno-error=pass-failed -fsanitize=address,undefined -shared-libsan"
+        CXXFLAGS: "-Werror -Wno-error=pass-failed -fsanitize=address,undefined -shared-libsan -Wno-self-assign-overloaded"
```
ax3l:
```suggestion
        CXXFLAGS: "-Werror -Wno-error=pass-failed -Wno-error=self-assign-overloaded -fsanitize=address,undefined -shared-libsan"
```

PR #1179, opened 2025-10-13
CMake `pip_install`: Package Examples

ax3l:
Ensure the `examples/` directory is part of the `pip` wheels we build, e.g., on `cmake --build build -j 6 --target pip_install`.

Fix for issue introduced in #876

PR #1200, opened 2025-10-26
CMake: Install Python Wrappers w/o Test

ax3l:
The Python wrapper files and classes should be always installed, not only when tests are performed. Also, ideally before the examples #1179 are copied in.


ax3l:
Testing in https://github.com/conda-forge/impactx-feedstock/pull/58 ...

ax3l:
Did not help with the Windows issue, but definitely a bug.
Ready for merge.

cemitch99 (approved):

PR #1204, opened 2025-10-31
Dashboard: Wait for Inputs, Fix Python 3.14

ax3l:
Try to fix CI by asking Cursor to fix it.

## Wait for Inputs

Try to fix the failing Ubuntu test `test_python_import.py::test_python_import` saying
```
 Element {#lambdaX} was not visible after 10 seconds!
```
That one also passes on my local pc 

## Fix Python 3.14 (macOS CI)

Dashboard never started up:
```
      File "impactx/src/python/impactx/dashboard/start.py", line 45, in start
    server.start()
    
      File "/Library/Frameworks/Python.framework/Versions/3.14/lib/python3.14/site-packages/trame_server/core.py", line 716, in start
        task = CoreServer.server_start(
    ...
    
      File "/Library/Frameworks/Python.framework/Versions/3.14/lib/python3.14/site-packages/wslink/protocol.py", line 33, in __init__
        self._completion = asyncio.get_event_loop().create_future()
                           ~~~~~~~~~~~~~~~~~~~~~~^^
      File "/Library/Frameworks/Python.framework/Versions/3.14/lib/python3.14/asyncio/events.py", line 715, in get_event_loop
        raise RuntimeError('There is no current event loop in thread %r.'
                           % threading.current_thread().name)
    RuntimeError: There is no current event loop in thread 'MainThread'.
```

PR #1208, opened 2025-11-03
Dashboard: Fix population of distribution values & Python 3.14 issue

proy30:
This PR fixes issues from https://github.com/BLAST-ImpactX/impactx/pull/1204.
It’s submitted as a separate PR since I don’t have direct push access to that branch.

Fixes:
` Element {#lambdaX} was not visible after 10 seconds!` 

Right before we set distribution_type, we stash the parsed values from the import file; once `@state.change("distribution_type")` runs via the flush, we apply those cached numbers to the UI so the fields stay populated.

Unsure if the `state.flush()` mechanism changed recently in Trame to cause this error.


cemitch99 (approved):

ax3l (approved):
Thank you! :tada:

PR #1166, opened 2025-10-06
2.5D Gaussian Solver

qianglbl:
added 3 input parameters for the 2.5D Gaussian solver, modified Gauss2p5dPush.cpp, added examples in fodo_space_charge.

Enhances the implementation from #1127

PR #1214, opened 2025-11-10
2.5 Space Charge in Particle Tracking

cemitch99:
This PR adds a 2.5D Space Charge model for particle tracking, using a 2D Poisson solve (in the transverse variables X,Y) with current weighting along the longitudinal direction within the bunch.  This addresses the final part of #401, and is a follow-up to #909 .

- [x] add basic logic and input structure
- [x] add longitudinal charge deposition and derivative (following #1166 )
- [x] modify push logic
- [x] add Python bindings
- [x] update documentation
- [x] add example benchmark against Gauss2p5D solver (Gaussian bunch)
- [x] add example benchmark against 2D solver (KV bunch uniform in t)
- [x] fix the SC potential evaluation call (needed for the longitudinal kick)
- [x] select benchmark analysis tolerance
- [x] address the failing GPU checks
- [x] clean-up/check naming conventions for documentation


qianglbl (approved):

cemitch99:
I'm reconsidering this approach now.  The number of places where we need to replace `True_2D` with `True_2D` or `True_2p5D` is very large.  I'm tempted to re-use `True_2D` and add another boolean to set to `true` when the user sets `algo.space_charge = 2p5D`.

src/particles/spacecharge/GatherAndPush.cpp
```diff
@@ -35,6 +38,23 @@ namespace impactx::particles::spacecharge
 
         amrex::ParticleReal const charge = pc.GetRefParticle().charge;
 
+        // Deposit 1D charge density in cases where it is required.
+        amrex::Real * beam_profile = nullptr;
+        amrex::Real * beam_profile_slope = nullptr;
```
ax3l: Allocate the profiles here. Note that `_slope` is `num_bins` or so and the charge is `num_bins + 1`.
```suggestion
        amrex::Gpu::DeviceVector<amrex::Real> beam_profile(num_bins + 1, 0.0);
        amrex::Gpu::DeviceVector<amrex::Real> beam_profile_slope(num_bins, 0.0);
```

src/particles/spacecharge/GatherAndPush.cpp
```diff
@@ -35,6 +38,23 @@ namespace impactx::particles::spacecharge
...
+        [[maybe_unused]] auto const [x_min, y_min, t_min, x_max, y_max, t_max] =
+            pc.MinAndMaxPositions();
+
+        amrex::Real bin_min = t_min;
+        amrex::Real bin_max = t_max;
+        amrex::Real const bin_size = (bin_max - bin_min) / (num_bins - 1);
+
+        if (space_charge == SpaceChargeAlgo::True_2p5D) {
+
+            Deposit1D ( pc, beam_profile, beam_profile_slope, bin_min, bin_max, num_bins);
```
ax3l:
```suggestion
            Deposit1D( pc, beam_profile, beam_profile_slope, bin_min, bin_max, num_bins);
```

src/particles/spacecharge/Deposit1D.cpp
```diff
@@ -0,0 +1,66 @@
...
+        [[maybe_unused]] amrex::Real * beam_profile,
+        [[maybe_unused]] amrex::Real * beam_profile_slope,
+        amrex::Real bin_min,
+        amrex::Real bin_max,
+        int num_bins
+    )
+    {
+        BL_PROFILE("impactx::spacecharge::Deposit1D");
+
+        using namespace amrex::literals;
```
ax3l, line 36: Let's always assert here that `beam_profile.size()` is +1 of `beam_profile_slope.size()`

src/particles/spacecharge/Deposit1D.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include <cmath>
+
+
+namespace impactx::particles::spacecharge
+{
+
+    void Deposit1D (
+        ImpactXParticleContainer & pc,
+        [[maybe_unused]] amrex::Real * beam_profile,
+        [[maybe_unused]] amrex::Real * beam_profile_slope,
```
ax3l: Let's pass a `amrex::Gpu::DeviceVector<amrex::Real>` by reference `&` here for each array.
Omit `num_bins` and take it from `beam_profile_slope.size()`

ax3l: Even better: pass no arrays and just return the profile array and `bin_size`.

Let's do the one-liner for slope calc afterwards, it is safe and not part of deposit.

src/particles/spacecharge/GatherAndPush.cpp
```diff
@@ -111,11 +129,56 @@ namespace impactx::particles::spacecharge
...
+                        amrex::ParticleReal & AMREX_RESTRICT pz = part_pz[i];
+
+                        // force gather
+                        amrex::GpuArray<amrex::Real, 3> const field_interp =
+                            ablastr::particles::doGatherVectorFieldNodal<2>(
+                                x, y, z,
+                                scf_arr_x, scf_arr_y, scf_arr_z,
+                                invdr,
+                                prob_lo_2D
+                            );
```
ax3l, line 178: After passing also `m_phi` by reference into `GatherAndPush (...)`:
```suggestion
                            ablastr::particles::doGatherScalarFieldNodal<2>(
                                x, y, z,
                                phi_arr,
                                invdr,
                                prob_lo_2D
                            );
```

src/particles/spacecharge/GatherAndPush.cpp
```diff
@@ -111,11 +131,56 @@ namespace impactx::particles::spacecharge
...
+                        amrex::ParticleReal z = 0.0_prt;  // flatten 3rd dimension
+                        amrex::ParticleReal & AMREX_RESTRICT px = part_px[i];
+                        amrex::ParticleReal & AMREX_RESTRICT py = part_py[i];
+                        amrex::ParticleReal & AMREX_RESTRICT pz = part_pz[i];
+
+                        // force gather
+                        amrex::GpuArray<amrex::Real, 3> const field_interp =
+                            ablastr::particles::doGatherScalarFieldNodal<2>(
+                                x, y, z,
+                                phi_arr,
```
cemitch99: Here `phi_arr` still needs to be declared with the correct type.

cemitch99:
Finally, good agreement between the Gauss2p5D solver and the above 2p5D PIC solve, at least illustrated for a single space charge kick.
[Px_kick_cf.pdf](https://github.com/user-attachments/files/24202625/Px_kick_cf.pdf)
[Px_vs_z_kick_cf.pdf](https://github.com/user-attachments/files/24202626/Px_vs_z_kick_cf.pdf)
[Py_kick_cf.pdf](https://github.com/user-attachments/files/24202627/Py_kick_cf.pdf)
(Note: this is a very low-resolution comparison using only 10K particles.)

src/particles/spacecharge/GatherAndPush.cpp
```diff
@@ -111,11 +136,70 @@ namespace impactx::particles::spacecharge
...
+                                x, y, z_flat,
+                                scf_arr_x, scf_arr_y, scf_arr_z,
+                                invdr,
+                                prob_lo_2D
+                            );
+
+                        // potential gather
+                        amrex::Real potential_interp = 0.0;
+                        if (apply_longitudinal_kick) {
+                           potential_interp =
```
cemitch99, line 183: This function call is still causing a problem (only needed when `apply_longitudinal_kick=true`).

cemitch99:
Remaining issues to address:

- the potential return in `GatherAndPush.cpp`, lines 174-180 needs some repair (needed for the longitudinal kick); currently, it returns 0 or does not execute?
- add Python bindings for the new user inputs (address failing Python tests)
- address the two failing GPU checks

cemitch99:
Regarding the evaluation of the potential:  The array `phi_arr` in `GatherAndPush.cpp` needs to be set, similar to the arrays `scf_arr_x`, etc. in lines 89-91.  Is it necessary to pass `m_phi` as an additional argument to `GatherAndPush` (e.g. in `HandleSpaceCharge.cpp` lines 100-105), or is this information also available in the `space_charge_field` MultiFab or elsewhere? (Recall a previous conversation with Axel suggesting that passing the additional argument was not needed.)

examples/fodo_space_charge/README.rst
```diff
@@ -206,3 +208,108 @@ We run the following script to analyze correctness:
    .. literalinclude:: analysis_fodo_2d_sc.py
       :language: python3
       :caption: You can copy this file from ``examples/fodo_space_charge/analysis_fodo_2d_sc.py``.
+
+
+.. _examples-fodo-2p5d-sc:
+
+FODO Cell with 2.5D Space Charge using FFT IGF Poisson Solver
+=================================================================
```
ax3l:
```suggestion
FODO Cell with 2.5D Space Charge using FFT IGF Poisson Solver
=============================================================
```

examples/fodo_space_charge/README.rst
```diff
@@ -206,3 +208,108 @@ We run the following script to analyze correctness:
...
+
+   .. literalinclude:: analysis_fodo_2p5D_sc.py
+      :language: python3
+      :caption: You can copy this file from ``examples/fodo_space_charge/analysis_fodo_2p5D_sc.py``.
+
+
+.. _examples-fodo-2p5d-kv-sc:
+
+FODO Cell with 2.5D Space Charge using FFT IGF Poisson Solver (Matched)
+========================================================================
```
ax3l:
```suggestion
FODO Cell with 2.5D Space Charge using FFT IGF Poisson Solver (Matched)
=======================================================================
```

docs/source/usage/python.rst
```diff
@@ -93,6 +94,19 @@ Collective Effects & Overall Simulation Parameters
...
+
+        These models are supported only in particle tracking mode (when ``algo.track = "particles"``).
+
+   .. py:property:: space_charge_num_longitudinal_bins
+
+      Number of bins for longitudinal charge density deposition (default: ``1``).
+
+   .. py:property:: space_charge_apply_longitudinal_kick
+
+      Enable or disable the longitudinal space charge kick (default: ``false``).
```
ax3l: The Python keyword is uppercase:
```suggestion
      Enable or disable the longitudinal space charge kick (default: ``False``).
```

docs/source/usage/python.rst
```diff
@@ -93,6 +94,19 @@ Collective Effects & Overall Simulation Parameters
...
+      * ``"2p5D"``: Space charge forces are computed in the plane ``(x,y)`` transverse to the reference particle velocity, while the transverse space charge kicks are weighted by the
+        longitudinal line density determined by charge deposition (2.5D model).  Longitudinal space charge kicks are determined by the derivative of the line charge density.
+
+        These models are supported only in particle tracking mode (when ``algo.track = "particles"``).
+
+   .. py:property:: space_charge_num_longitudinal_bins
+
+      Number of bins for longitudinal charge density deposition (default: ``1``).
+
+   .. py:property:: space_charge_apply_longitudinal_kick
```
ax3l, line 106: Would it be physically more meaningful to turn this on by default and only let users that explicitly want to ignore the long. effects turn it off?

docs/source/usage/parameters.rst
```diff
@@ -957,6 +957,21 @@ See there ``nslice`` option on lattice elements for slicing.
...
+
+    This model is supported only in particle tracking mode (when ``algo.track = "particles"``).
+
+    This model supports the following sub-options:
+
+    * ``algo.space_charge.num_longitudinal_bins`` (``int``, default: ``1``)
+
+      The number of bins for longitudinal line density deposition.
+
+    * ``algo.space_charge.apply_longitudinal_kick`` (``bool``, default: ``false``)
```
ax3l: Would it be physically more meaningful to turn this on by default and only let users that explicitly want to ignore the long. effects turn it off?

docs/source/usage/python.rst
```diff
@@ -93,6 +94,19 @@ Collective Effects & Overall Simulation Parameters
...
       Number of bins for longitudinal charge density deposition (default: ``129``).
 
+      * ``"2p5D"``: Space charge forces are computed in the plane ``(x,y)`` transverse to the reference particle velocity, while the transverse space charge kicks are weighted by the
+        longitudinal line density determined by charge deposition (2.5D model).  Longitudinal space charge kicks are determined by the derivative of the line charge density.
+
+        These models are supported only in particle tracking mode (when ``algo.track = "particles"``).
+
+   .. py:property:: space_charge_num_longitudinal_bins
+
+      Number of bins for longitudinal charge density deposition (default: ``1``).
```
ax3l: Is this a meaningful default or likely results in wrong results if unchanged?

I am surprised that `2` is sufficient in the example. Wouldn't we usually default to like 100 bins, as we do for wake effects (CSR)?

docs/source/usage/parameters.rst
```diff
@@ -957,6 +957,21 @@ See there ``nslice`` option on lattice elements for slicing.
...
       Number of bins for longitudinal line density deposition.
 
+  * ``"2p5D"``: Space charge forces are computed in the plane ``(x,y)`` transverse to the reference particle velocity, while the transverse space charge kicks are weighted by the
+longitudinal line density determined by charge deposition (2.5D model).  Longitudinal space charge kicks are determined by the derivative of the line charge density.
+
+    This model is supported only in particle tracking mode (when ``algo.track = "particles"``).
+
+    This model supports the following sub-options:
+
+    * ``algo.space_charge.num_longitudinal_bins`` (``int``, default: ``1``)
```
ax3l: Pls see comment on the default below.

src/particles/spacecharge/GatherAndPush.cpp
ax3l: @cemitch99 is this function still returning zeros? Is `phi_arr` non-zero?

docs/source/usage/parameters.rst
cemitch99: Yes.  I agree it's a good idea to change the default here.  (I set this default before the longitudinal kick was working properly, in case this required a separate follow-up PR.)

docs/source/usage/python.rst
cemitch99: Yes, we can change this to 100.  

There are two examples.  In the example that uses a KV beam, the longitudinal density is uniform and so a small number of bins is ok (all bins receive the same space charge kick).  The purpose is to compare against a pure 2D solve.  For the other example, the longitudinal density is Gaussian and a sufficient number of bins is needed to resolve this.

docs/source/usage/python.rst
cemitch99: Agreed.

src/particles/spacecharge/Deposit1D.cpp
ax3l: Resolved

ax3l (approved):
Looks great! Thank you :sparkles:

PR #1215, opened 2025-11-11
Pytest: SP Tolerances

ax3l:
Isolated from #899

### DIFF
--- a/tests/python/test_impactx.py
+++ b/tests/python/test_impactx.py
@@ -168,7 +168,7 @@ def test_impactx_nofile():
 
     # simulate full lattice but keep beam global position
     sim.track_particles()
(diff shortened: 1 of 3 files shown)


tests/python/test_impactx.py
```diff
@@ -160,7 +168,7 @@ def test_impactx_nofile():
 
     # simulate full lattice but keep beam global position
     sim.track_particles()
-    assert np.allclose([ref.s], [7.0])
+    # assert ref.s == pytest.approx(7.0)
```
ax3l:
```suggestion
    assert ref.s == pytest.approx(7.0)
```

cemitch99 (approved):

PR #1217, opened 2025-11-17
plot_survey: Fix height calculation for (Exact)Sbend elements

KroneBonn:
Fix conditional check for 'ref' and add calculation for bending angle.

Use ref when it's not None and calculate the exact bending angle for Sbend through the ratio of segment length and bending radius.


ax3l (approved):
Thank you, good catch!

cemitch99 (approved):

Issue #1231, opened 2025-11-26
Warnings about missing files when building documentation


kyrsjo:
When I run `make html` in the docs folder, it builds correctly, but I get the warning (in red):

```
reading sources... [100%] usage/tests/python/transformation
/home/kyrsjo/code/impactx/docs/source/usage/examples/edge_effects/README.rst:43: WARNING: Include file '/home/kyrsjo/code/impactx/examples/edge_effects/analysis_exact_quad.py' not found or reading it failed [docutils]
/home/kyrsjo/code/impactx/docs/source/usage/howto/add_element.rst:97: WARNING: end-at pattern not found: register_envelope_push(py_Drift); [docutils]
```

This should be fixed, however it should probably also be detected by the CI for a pull request?

ax3l:
Thanks for the report! Yes they are typos and need fixes.

We do not fail in warnings in CI yet because Sphinx + all the plugins we use can be quite flaky 😅 
Not sure how to selectively fail Sphinx on missing files, do you know?

kyrsjo:
Great! Unfortunately, no. Just fail on any warning, which is probably too much.

I guess MK1 eyeball works, but one could maybe replace it with a small scripts that reads the output... But it's probably not worth the effort. Maybe open a feature request against Sphinx...

PR #1235, opened 2025-12-01
Fix: DipEdge Analysis Docs

ax3l:
The README included the wrong file, so the analysis script was not shown.

Related to #1231


cemitch99 (approved):
Thanks for catching this!

PR #1237, opened 2025-12-02
Updates to DipEdge documentation.

### LINKED ISSUE #1220: Default parameters for DipEdge element
Is it worth pointing out that the default for the K2 parameter in the DipEdge element corresponding to MAD-X FINT is now 1.0, while the default for FINT in MAD-X is 0? It doesn't matter if the gap g is 0, but if g is nonzero you will see differences. It's something that people simulating MAD-X lattices in ImpactX should be aware of.

cemitch99:
This PR makes a few changes to the documentation of the DipEdge element.

In particular, this is meant to address / close #1220

### DIFF
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -338,6 +338,8 @@ If ``model = "linear"``, then the linearized map is used.  This model is identic
 
 when expanded to first order in ``g/rc`` (gap / radius of curvature).
 
+By comparison, note that the MAD-X DIPEDGE element uses as input the half-gap ``HGAP = g/2``, and sets the default value ``FINT = 0`` (while the corresponding default value of ``K2`` is set to 1).
+
 This requires these additional parameters:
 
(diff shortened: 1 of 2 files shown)


docs/source/usage/parameters.rst
```diff
@@ -346,7 +348,7 @@ This requires these additional parameters:
 * ``<element_name>.R`` (``float``, in meters) scale length for the field integrals (default: ``1 m``)
 * ``<element_name>.K0`` (``float``, dimensionless) normalized field integral for fringe field
```
ax3l: @cemitch99 should the default value here also be `pi/6`?

cemitch99: @ax3l Oh, several lines above it is mentioned that the defaults come from the reference cited, but I didn't list them explicitly here.  I can add those values.

ax3l (approved):
Thank you, looks great!

PR #1239, opened 2025-12-03
Source: `active_once = True`

ax3l:
Add an option `active_once` to the `Source` element. This controls if the source should only inject particles when tracking through lattice period 0. This is useful when defining rings or periodic channels with a source that shall only inject particles once.

Breaking change: `active_once = True` is also the new default, since it is the most common use case.

Thanks to @egstern for raising this! :pray:


egstern (commented):
Looks good.

egstern (approved):

PR #1242, opened 2025-12-05
Enhance plot_survey: scale element heights based on total length

### LINKED ISSUE #1225: plot_survey: height of elements do not scale well with long beamlines.
https://github.com/BLAST-ImpactX/impactx/blob/88a4652a7289447e3e8c52ecead86486451234a5/src/python/impactx/plot/Survey.py#L85

KroneBonn:
Fix #1225: Scaling the elements height according to the length of the lattice (s position of last element). Move anchor point for monitor accordingly.

### DIFF
--- a/src/python/impactx/plot/Survey.py
+++ b/src/python/impactx/plot/Survey.py
@@ -52,9 +52,6 @@ def plot_survey(
     # backport:
     element_s = np.insert(np.cumsum(element_lengths), 0, 0)
 
-    height_scale = element_s[-1] / 6
-    # use total length in s as scale for y-axis
-
     ax.hlines(0, 0, element_s[-1], color="black", linestyle="--")
 
     # plot config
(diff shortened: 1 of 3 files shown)


src/python/impactx/plot/Survey.py
```diff
@@ -52,6 +52,9 @@ def plot_survey(
     # backport:
     element_s = np.insert(np.cumsum(element_lengths), 0, 0)
 
+    height_scale = element_s[-1] / 6
+    # use total length in s as scale for y-axis
```
ax3l: That works - or we could make this simpler -- multiply the aspect ratio with `height_scale` and leave the rest simple.

What do you think @KroneBonn ?

src/python/impactx/plot/Survey.py
```diff
@@ -128,7 +131,7 @@ def plot_survey(
 
     ax.set_xlabel(r"$s$ [m]")
 
-    ax.set_ylim(-1, 1)
+    ax.set_ylim(-height_scale, height_scale)
     ax.set_yticks([])
 
     ax.set_aspect(1 / 1.618)  # golden ratio
```
ax3l: Another approach: maybe replacing this with
```suggestion
    ax.set_box_aspect(1 / 6)  # just some nice ratio on canvas
```
Could help. Originally, I mean to set the ratio of the [axes box on canvas](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.set_box_aspect.html), not the aspect based on the plotted data.

This needs matplotlib 3.3+ or 3.1+ or so (for our `requirements.txt`)

KroneBonn: I think this is a nice approach. I think this looks nice. Looks like matplotlib 3.3+

KroneBonn: Just thought about it: If going that route one could also easily make the aspect-ratio be accessible for the user and just default to 1/6
Would just need another optional argument

ax3l: Sounds great! Do you like to adjust your PR accordingly by pushing another commit to it?

>  Just thought about it: If going that route one could also easily make the aspect-ratio be accessible for the user and just default to 1/6.
> Would just need another optional argument

Yes, I like that. But what I realize usually is that there are many things one might want to change in a final plot, including labels, etc. It is hard to draw a line what to pass into plot functions that way and one gets more and more arguments in the signature.

For that reason, I usually return the matplotlib `ax` object to the user in our plot functions, so one can call any other function and manipulation on it as needed :)
```py
ax = plot_survey(...)
ax.set_box_aspect(another_value)
```

KroneBonn: That's actually way nicer. Will update my PR tomorrow!

ax3l:
@KroneBonn let me know when you are ready. Will wait with the 25.12 release for this PR to be included :)

KroneBonn:
@ax3l Sorry today was a bit hectic. PR is now updated, I was not sure which requirements.txt to update, maybe you can help out. Thanks for waiting.

ax3l (approved):
Thank you! :rocket: :sparkles:

kyrsjo:
Thanks :)

PR #1251, opened 2025-12-17
Fix: `track_particles` with `1` particle

ax3l:
We can in principle allow to run with `1` particle in `track_particles()`, even though we do have the `track_reference(ref)` mode now: https://impactx.readthedocs.io/en/25.12/usage/how_to_run.html#how-to-select-a-tracking-mode

As long as no space charge and no CSR/wakefield effects are requested, this is still a valid configuration.

This PR removes an old assert, for which we have a more specific assert in the space charge routines, if enabled. Additionally, it adds a new check into the `HandleWakefield` routines (for CSR), which also rely on a non-flat beam to model anything sensible.

Other routines we have, notably ISR, are modeled on individual particles and need no asserts.

First seen in https://github.com/BLAST-ImpactX/impactx/issues/44#issuecomment-3667400046

### DIFF
--- a/src/particles/wakefields/HandleWakefield.H
+++ b/src/particles/wakefields/HandleWakefield.H
@@ -17,9 +17,7 @@
 #include "WakePush.H"
 
 #include <cmath>
-#ifndef ImpactX_USE_FFT
 #include <stdexcept>
-#endif
 #include <vector>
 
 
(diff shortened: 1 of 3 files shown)


src/particles/wakefields/HandleWakefield.H
```diff
@@ -64,6 +62,25 @@ namespace impactx::particles::wakefields
             [[maybe_unused]] auto const [x_min, y_min, t_min, x_max, y_max, t_max] =
                 particle_container.MinAndMaxPositions();
 
+            // Ensure we have a beam with a finite extend in t
```
cemitch99:
```suggestion
            // Ensure we have a beam with a finite extent in t
```

src/particles/wakefields/HandleWakefield.H
```diff
@@ -64,6 +62,25 @@ namespace impactx::particles::wakefields
...
+                std::string error_msg = "Coherent Synchrotron Radiation (CSR) "
+                                        "modeling was enabled, but ";
+                if (npart == 1) {
+                    error_msg.append("there is only one particle in the beam. ");
+                } else {
+                    error_msg.append("the beam is flat in t. ");
+                }
+                error_msg.append(
+                    "CSR modeling is a collective effect and requires "
+                    "a particle beam with a finite extend in t. "
```
cemitch99:
```suggestion
                    "a particle beam with a finite extent in t. "
```

cemitch99 (approved):

PR #1256, opened 2026-01-02
Calculate Linear Transfer Map

### LINKED ISSUE #1254: Implement return of linear transfer maps
Multiply (already implemented) element linear transfer matrices (at user request) to return the net linear transfer map between initial and final simulated lattice locations, e.g. to compute the linear one-turn map for a ring.

This is a follow-up to Issue #961

ax3l:
Add a `transfer_map(ref, order="linear", fallback_identity_map=False)` method to the element list, supporting the calculation of a linear transfer map of all elements in the list.

Close #1254

# To Do / To Discuss

- [x] Python: implementation, test, docs
- [x] Implementation: Do we need to advance the reference particle with every element?
- [x] Do we need an output/diagnostics (to a file) for inputs files? Should we skip it and add it later/when users ask for it?
- [x] Will we ever be able to support on lattice a transfer map calculation that is non-linear via a `lattice...` method or should we rather add an interface to `sim. ...` (assuming higher-order transfer maps require particle tracking) - discussed: the primary use cases are only on the reference particle trajectory (linear and higher order). We [could add this to tracking](https://github.com/ax3l/impactx/tree/topic-lin-transport-map-tracking) for the few collective effects that change the reference particle, but this is uncommon.

### DIFF
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -774,9 +774,9 @@ This module provides elements and methods for the accelerator lattice.
       Calculate the transfer map of the elements in the list.
 
       :param ref: A reference particle.
-      :param order: So far, only the calculation of linear transfer maps are supported in this function.
-      :param fallback_identity_map: For elements with an undefined transfer map in lattice, assume the identity matrix.
-      :return: The transfer map map of all elements in the list.
+      :param order: So far, only the calculation of linear transfer maps is supported.
+      :param fallback_identity_map: For elements with an undefined transfer map in the lattice, assume the identity matrix.
+      :return: The transfer map of all elements in the list.
       :rtype: Map6x6
 
    .. py:method:: plot_survey(ref=None, ax=None, legend=True, legend_ncols=5)
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -769,6 +769,16 @@ This module provides elements and methods for the accelerator lattice.
       :return: True if at least one element of the specified kind exists
       :rtype: bool
 
+   .. py::method:: transfer_map(ref, order="linear", fallback_identity_map=False)
+
+      Calculate the transfer map of the elements in the list.
+
+      :param ref: A reference particle.
+      :param order: So far, only the calculation of linear transfer maps are supported in this function.
+      :param fallback_identity_map: For elements with an undefined transfer map in lattice, assume the identity matrix.
+      :return: The transfer map map of all elements in the list.
+      :rtype: Map6x6
+
    .. py:method:: plot_survey(ref=None, ax=None, legend=True, legend_ncols=5)
 
       Plot over s of all elements in the KnownElementsList.
@@ -777,7 +787,6 @@ This module provides elements and methods for the accelerator lattice.
 
       Either populates the matplotlib axes in ax or creates a new axes containing the plot.
 
-      :param self: The KnownElementsList class in ImpactX
       :param ref: A reference particle, checked for the charge sign to plot focusing/defocusing strength directions properly.
       :param ax: A plotting area in matplotlib (called axes there).
       :param legend: Plot a legend if true.
--- a/src/python/elements.cpp
+++ b/src/python/elements.cpp
@@ -9,6 +9,7 @@
 #include <elements/All.H>
 #include <elements/mixin/lineartransport.H>
 #include <elements/transformation/Insert.H>
+#include <particles/CovarianceMatrix.H>
 
 #include <AMReX_Enum.H>
 #include <AMReX_REAL.H>
@@ -20,7 +21,6 @@
 #include <utility>
 #include <variant>
 #include <vector>
-#include <particles/transformation/CoordinateTransformation.H>
 
 namespace py = pybind11;
 using namespace impactx;
@@ -313,7 +313,7 @@ void init_elements(py::module& m)
         )
     ;
 
-    /*
+    /* TODO
     py::class_<elements::mixin::LinearTransport>(mx, "LinearTransport")
         // type of map
         .def_property_readonly_static("Map6x6",
@@ -2615,6 +2615,49 @@ void init_elements(py::module& m)
             auto it = std::next(v.begin(), index);
             return *it;  // return by reference
         }, py::return_value_policy::reference_internal)
+
+        .def(
+            "transfer_map",
+            [](
+                KnownElementsList &v,
+                RefPart const & ref,
+                std::string order,
+                bool fallback_identity_map
+            )
+            {
+                if (order != "linear") {
+                    throw std::runtime_error("So far, only the calculation of linear transfer maps are supported in this function.");
+                }
+                Map6x6 result = Map6x6::Identity();
+                for (auto & el_v : v) {
+                    std::visit([&result, &ref, &fallback_identity_map](auto const & el) {
+                        using Element = std::decay_t<decltype(el)>;
+                        std::string not_impl_msg = "Undefined transfer map in lattice for element ";
+                        if (el.has_name()) not_impl_msg += el.name() + " ";
+                        not_impl_msg += std::string("of type ") + Element::type;
+
+                        if constexpr (std::is_base_of_v<elements::mixin::LinearTransport<Element>, Element>) {
+                            try {
+                                result = result * el.transport_map(ref);
+                            } catch (std::exception const & e) {
+                                if (!fallback_identity_map) {
+                                    throw std::runtime_error(not_impl_msg);
+                                }
+                            }
+                        } else {
+                            if (!fallback_identity_map) {
+                                throw std::runtime_error(not_impl_msg);
+                            }
+                        }
+                    }, el_v);
+                }
+                return result;
+            },
+            py::arg("ref"),
+            py::arg("order") = "linear",
+            py::arg("fallback_identity_map") = false,
+            "Calculate the transfer map of the elements in the list."
+            )
     ;
 
 
--- a/src/python/elements.cpp
+++ b/src/python/elements.cpp
@@ -2620,7 +2620,7 @@ void init_elements(py::module& m)
             "transfer_map",
             [](
                 KnownElementsList &v,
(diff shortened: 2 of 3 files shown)


src/python/elements.cpp
```diff
@@ -2615,6 +2615,49 @@ void init_elements(py::module& m)
...
+                for (auto & el_v : v) {
+                    std::visit([&result, &ref, &fallback_identity_map](auto const & el) {
+                        using Element = std::decay_t<decltype(el)>;
+                        std::string not_impl_msg = "Undefined transfer map in lattice for element ";
+                        if (el.has_name()) not_impl_msg += el.name() + " ";
+                        not_impl_msg += std::string("of type ") + Element::type;
+
+                        if constexpr (std::is_base_of_v<elements::mixin::LinearTransport<Element>, Element>) {
+                            try {
+                                result = result * el.transport_map(ref);
```
ax3l: @cemitch99 I think we also need to advance the reference particle with every element, is that correct?
Simply multiplying the matrix elements with the initial reference particle is probably not sufficient, is that right?

cemitch99: That is correct.  For example, we do this during envelope tracking.  The reference particle is pushed along with the covariance matrix, and the push for the covariance matrix involves multiplication by the linear transfer matrix at s.

ax3l: That makes sense.

As a user-interface, I wonder if (also in light of calculating higher-order transfer maps via particle tracking) we could design this as an interface to `sim ...` (populating/calculating the map during `track_...`).

As a first draft, I added it on the lattice element list, but it is that way very specific to reference orbit tracking and maybe not sensibly extensible with the common usage.

What do you think?

ax3l: Hm, I implemented this in tracking https://github.com/ax3l/impactx/tree/topic-lin-transport-map-tracking

But because we still lack so many `transport_map` implementations, this could be a bit cumbersome to add.

Maybe I continue the route here for now.

docs/source/usage/python.rst
```diff
@@ -769,6 +769,16 @@ This module provides elements and methods for the accelerator lattice.
       :return: True if at least one element of the specified kind exists
       :rtype: bool
 
+   .. py::method:: transfer_map(ref, order="linear", fallback_identity_map=False)
```
ax3l, line 772: For review: we could also make `order` a number, if that is more sensible (linear: 1, etc.).

docs/source/usage/python.rst
```diff
@@ -769,6 +769,16 @@ This module provides elements and methods for the accelerator lattice.
...
+
+      Calculate the transfer map of the elements in the list.
+
+      :param ref: A reference particle.
+      :param order: So far, only the calculation of linear transfer maps is supported.
+      :param fallback_identity_map: For elements with an undefined transfer map in the lattice, assume the identity matrix.
+      :return: The transfer map of all elements in the list.
```
ax3l, line 779: For review: if we are later on calculating higher orders, should we return a list of transfer maps (one per order)?

src/python/elements.cpp
```diff
@@ -2615,6 +2615,62 @@ void init_elements(py::module& m)
...
+
+        .def(
+            "transfer_map",
+            [](
+                KnownElementsList &v,
+                RefPart ref, // note: intentional copy
+                std::string order,
+                bool fallback_identity_map
+            )
+            {
```
ax3l, line 2627: I think we could also implement this part fully in Python and it would probably be cleaner.
Only the exception handling in the inner loop is likely (quite) slower.

Can also be done later, when we find that we like to implement higher order logic rather in Python than C++.

docs/source/usage/python.rst
cemitch99: For higher orders, there are several ways to represent the map, which may be organized into one (or more) arrays.  It's useful to organize these by order-by-order.  However, regarding _terminology_ all this information is usually referred to as one map (singular).

docs/source/usage/python.rst
cemitch99: I think this is fine as-is for now.  The best key word really depends on what options we would like to support in the future.  For example, we could have multiple methods of computing the map (e.g., using Taylor maps or automatic differentiation or Lie maps), and "method" would be a good key word.

src/python/elements.cpp
```diff
@@ -2615,6 +2615,62 @@ void init_elements(py::module& m)
...
+                        } else {
+                            if (!fallback_identity_map) {
+                                throw std::runtime_error(not_impl_msg);
+                            }
+                        }
+                    }, el_v);
+
+                    // advance linear transfer map
+                    linear_transfer_map = linear_transfer_map * element_transport_map;
+                    // TODO: shorthand needs https://github.com/AMReX-Codes/amrex/pull/4880 from AMReX 26.02+
```
cemitch99, line 2664: Does this comment mean that we need to wait on this feature in AMReX before merging this PR, or is that just for line 2665 below?

tests/python/test_lattice_optics.py
```diff
@@ -0,0 +1,63 @@
...
+    R_expected = np.array(
+        [
+            [3.74670546e-01, 2.54005827e00, 0, 0, 0, -2.34864805e-01],
+            [-4.76219076e-01, -5.59489407e-01, 0, 0, 0, 3.20646253e-02],
+            [0, 0, 1.64872127e00, 6.59488508e00, 0, 0],
+            [0, 0, 5.21095305e-01, 2.69091188e00, 0, 0],
+            [9.98334297e-02, 4.99583537e-02, 0, 0, 1.00000000e00, -1.66466013e-03],
+            [0, 0, 0, 0, 0, 1.00000000e00],
+        ]
+    )
```
cemitch99, line 44: Are these currently the true expected values or just dummy values?

tests/python/test_lattice_optics.py
```diff
@@ -0,0 +1,63 @@
...
+        lattice.transfer_map(ref, order="invalid")
+
+    # Create a lattice with an element that does not define a linear transfer map
+    lattice.append(elements.TaperedPL(k=0, taper=0, unit=0))
+
+    # Ensure that the calculation asserts
+    with pytest.raises(RuntimeError):
+        lattice.transfer_map(ref)
+
+    # Now the user explicitly assumes that undefined maps are identify maps
```
cemitch99:
```suggestion
    # Now the user explicitly assumes that undefined maps are identity maps
```

src/python/elements.cpp
ax3l: No, I used the version `A = A * B` that is already supported.

Later on, we can use the short-hand `A *= B;` in line 2665.

tests/python/test_lattice_optics.py
ax3l: I quickly put in the current computational result as expected. We should verify manually.

src/python/elements.cpp
```diff
@@ -2615,6 +2615,62 @@ void init_elements(py::module& m)
...
+                            }
+                        } else {
+                            if (!fallback_identity_map) {
+                                throw std::runtime_error(not_impl_msg);
+                            }
+                        }
+                    }, el_v);
+
+                    // advance linear transfer map
+                    linear_transfer_map = linear_transfer_map * element_transport_map;
```
cemitch99: I think the order of multiplication needs to be reversed here.

cemitch99:
```suggestion
                    linear_transfer_map = element_transport_map * linear_transfer_map;
```

ax3l: dang, yes :sweat:

tests/python/test_lattice_optics.py
ax3l: Oh, def wrong values because of the multiplication order mix up :D

docs/source/usage/python.rst
ax3l: I could also remove the `order` keyword at all for now, since we do not use it. We can always introduce more keywords when we need them.

cemitch99 (approved):
I fixed the existing test and ran some additional sanity checks, and everything looks good.

ax3l:
Thank you! :)

src/python/elements.cpp
```diff
@@ -2660,7 +2660,7 @@ void init_elements(py::module& m)
                     }, el_v);
 
                     // advance linear transfer map
-                    linear_transfer_map = linear_transfer_map * element_transport_map;
+                    linear_transfer_map = element_transport_map * linear_transfer_map;
                     // TODO: shorthand needs https://github.com/AMReX-Codes/amrex/pull/4880 from AMReX 26.02+
                     // result *= element_transport_map;
```
ax3l, line 2665: Comment cleanup needed: `result *= element_transport_map;` is not identical to the line above.

src/python/elements.cpp
```diff
@@ -2615,6 +2615,62 @@ void init_elements(py::module& m)
...
+                            }
+                        } else {
+                            if (!fallback_identity_map) {
+                                throw std::runtime_error(not_impl_msg);
+                            }
+                        }
+                    }, el_v);
+
+                    // advance linear transfer map
+                    linear_transfer_map = element_transport_map * linear_transfer_map;
```
ax3l, line 2663: I forgot that the map is sliced by `nslice` here, fix is in #1338: need to apply the map `nslice` times

src/python/elements.cpp
```diff
@@ -2615,6 +2615,62 @@ void init_elements(py::module& m)
...
+            {
+                if (order != "linear") {
+                    throw std::runtime_error("So far, only the calculation of linear transfer maps are supported in this function.");
+                }
+                Map6x6 linear_transfer_map = Map6x6::Identity();
+                for (auto & el_v : v)
+                {
+                    // advance reference particle
+                    std::visit([&ref](auto && el) {
+                        el(ref);
```
ax3l, line 2636: I forgot that ref updates also need to be done per slice, fix in #1411.

I also forgot to set `ref.sedge` on element entry. Fix in #1411

PR #1259, opened 2026-01-09
Spin push for ChrQuad and all drift elements

cemitch99:
This PR adds the spin push for four elements, including the three drift elements (trivial).

- [x] add spin push operators
- [x] add benchmark example

Multi-element benchmark tests including this feature will appear as follow-up.
This will include the example of spin in a long FODO channel.
See Issue https://github.com/BLAST-ImpactX/impactx/issues/1287


qianglbl (approved):

examples/spin_tracking/run_fodo_channel_spin.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python3
```
ax3l, line 1: @cemitch99 did you mean to remove this file and `examples/spin_tracking/input_fodo_channel_spin.in`?

At the moment, they are undocumented and not part of `examples/CMakeLists.txt`, so also not run.

I think you probably meant to write a future tests that tests multiple elements at once, so we could remove the two here.

ax3l: Discussed offline: this is the test that will be expanded (to multiple elements) later on. Staying in this state until then.

PR #1264, opened 2026-01-13
Spin push in soft-edge elements

cemitch99:
This PR adds the spin push for the (linear) soft-edge elements (RFCavity, SoftQuad, SoftSol).

- [x] add spin-orbit coupling matrix to rf cavity data structure (similar to transport matrix)
- [x] add push updates to spin-orbit coupling matrix during reference particle push
- [x] implement spin push to particles using spin-orbit coupling matrix
- [x] add push updates to spin-orbit coupling matrix for SoftQuad and SoftSol elements
- [x] add Python bindings
- [x] double-check math and implementation
- [x] add benchmark example


tests/python/test_solenoid_softedge_solvable_spin.py
```diff
@@ -0,0 +1,163 @@
...
+
+    print()
+    print("Spin-orbit coupling matrix, predicted:")
+    for i in range(1, 4):
+        for j in range(1, 7):
+            print(i, j, Apred[i, j])
+
+    # clean shutdown
+    sim.finalize()
+
```
ax3l: This might finalize too early for data access, but I cannot put the finger on it exactly:
```suggestion
```

ax3l: Ah right, `sim.finalize()` frees the particle container, who owns the ref particle, who we reference to its maps the below.

src/particles/ReferenceParticle.H
```diff
@@ -46,6 +46,8 @@ namespace impactx
 
         amrex::ParticleReal sedge = 0.0;  ///< value of s at entrance of the current beamline element
         amrex::SmallMatrix<amrex::ParticleReal, 6, 6, amrex::Order::F, 1> map{}; ///< linearized map
+        amrex::SmallMatrix<amrex::ParticleReal, 3, 6, amrex::Order::F, 1> spin_coupling{}; ///< linearized spin-orbit coupling matrix
+        amrex::SmallMatrix<amrex::ParticleReal, 3, 1, amrex::Order::F, 1> spin_rotation_vector{}; ///< reference spin rotation vector
```
ax3l, line 50: Good for now. As discussed, I need to rethink the interfaces in #1423

cemitch99:
Hi @ax3l Just checking whether any follow-up is requested from my end on this one.

ax3l:
This looks great - just a small cleanup in `tests/python/test_solenoid_softedge_solvable_spin.py` per the auto-check annotations. I can push that.

ax3l (approved):
🚀 ✨

src/elements/SoftQuad.H
```diff
@@ -488,6 +544,19 @@ namespace impactx::elements
             refpart.map(4,3) = R(4,3) + tau*alpha*R(3,3);
             refpart.map(4,4) = R(4,4) + tau*alpha*R(3,4);
 
+            // BELOW: if spin is needed only:
+            amrex::SmallMatrix<amrex::ParticleReal, 3, 6, amrex::Order::F, 1> const A = refpart.spin_coupling;
```
ax3l, line 556: Codex spotted:

1. src/elements/SoftQuad.H:548 and src/elements/RFCavity.H:497: spin_coupling is accumulated from the previous reference-particle state instead of being reset for the current element/slice.
SoftSolenoid clears both spin_rotation_vector and spin_coupling before integration, but SoftQuadrupole and RFCavity only reset refpart.map. That means a lattice with repeated soft quads or RF cavities reuses prior spin maps and applies them again to particles. I confirmed with two identical SoftQuadrupole reference pushes: the second element’s spin_coupling was exactly 2*A1.

src/elements/RFCavity.H
```diff
@@ -438,10 +494,34 @@ namespace impactx::elements
 
             // push the linear map equations
             amrex::SmallMatrix<amrex::ParticleReal, 6, 6, amrex::Order::F, 1> const R = refpart.map;
+            amrex::SmallMatrix<amrex::ParticleReal, 3, 6, amrex::Order::F, 1> const A = refpart.spin_coupling;
```
ax3l, line 513: Codex spotted:

1. src/elements/RFCavity.H:497 (and src/elements/SoftQuad.H:548): spin_coupling is accumulated from the previous reference-particle state instead of being reset for the current element/slice.
     SoftSolenoid clears both spin_rotation_vector and spin_coupling before integration, but SoftQuadrupole and RFCavity only reset refpart.map. That means a lattice with repeated soft quads or RFcavities reuses prior spin maps and applies them again to particles. I confirmed with two identical SoftQuadrupole reference pushes: the second element’s spin_coupling was exactly 2*A1.

src/elements/RFCavity.H
ax3l, line 313: Codex spotted:

  2. src/elements/RFCavity.H:303: the RF cavity converts refpart.map from dynamic to static units after acceleration, but the new refpart.spin_coupling is left untouched. Since `spin_and_phasespace_push()` multiplies `spin_coupling` by the same particle phase-space coordinates used with the converted map, the momentum columns of the spin coupling matrix look like they remain in the wrong basis when the cavity changes energy.

cemitch99: Now addressed below.

src/elements/RFCavity.H
```diff
@@ -323,10 +321,82 @@ namespace impactx::elements
...
             }
 
+            // convert spin-orbit coupling map from dynamic to static units
+            for (int i=1; i<4; i++) {
+               for (int j=1; j<7; j++) {
+                  if( j % 2 == 0)
+                     scale_in = bgi;
+                  else
+                     scale_in = 1.0_prt;
+                  refpart.spin_coupling(i, j) = refpart.spin_coupling(i, j) * scale_in;
```
ax3l: We can probably simplify this to:
```suggestion
                  if (j % 2 == 0)
                     refpart.spin_coupling(i, j) *= bgi;
```
(Or even make it a matrix-matrix multiplication for readability, but initializing that 2nd matrix will probably look as verbose.)

src/elements/RFCavity.H
```diff
@@ -323,10 +321,78 @@ namespace impactx::elements
                }
             }
 
+            // convert spin-orbit coupling map from dynamic to static units
+            for (int i=1; i<4; i++) {
+               for (int j=1; j<7; j++) {
+                  if (j % 2 == 0)
+                     refpart.spin_coupling(i, j) *= bgi;
+               }
+            }
```
ax3l, line 336: Even simpler:
```suggestion
            for (int i=1; i<4; i++) {
                for (int j=2; j<7; j+=2) {
                     refpart.spin_coupling(i, j) *= bgi;
            }
```

.github/workflows/ubuntu.yml
```diff
@@ -164,7 +164,7 @@ jobs:
           -DImpactX_PYTHON=ON          \
           -DImpactX_SIMD=ON            \
           -DImpactX_TEST_CLEANUP=ON
-        cmake --build build -j 4
+        cmake --build build -j 3
```
ax3l:
```suggestion
        cmake --build build -j 2
```

src/elements/RFCavity.H
```diff
@@ -438,10 +503,34 @@ namespace impactx::elements
...
             amrex::ParticleReal const betgam = refpart.beta_gamma();
 
             refpart.map(5,5) = R(5,5) + tau*R(6,5)/powi<3>(betgam);
             refpart.map(5,6) = R(5,6) + tau*R(6,6)/powi<3>(betgam);
+
+            // BELOW: if spin is needed only:
+            // Define parameters and intermediate constants
+            using ablastr::constant::math::pi;
+            using ablastr::constant::SI::c;
+            amrex::ParticleReal const k = (2_prt*pi/c)*m_freq;
```
ax3l, line 524: @cemitch99 should this be only executed when spin tracking is on? Seems like we also run this code path with regular phase space pushes now.

src/elements/SoftQuad.H
```diff
@@ -488,6 +545,19 @@ namespace impactx::elements
...
+            amrex::SmallMatrix<amrex::ParticleReal, 3, 6, amrex::Order::F, 1> dA = {};
+            amrex::ParticleReal const gamma = refpart.gamma();
```
ax3l, line 558: Should this be only executed when spin tracking is on? Seems like we also run this code path with regular phase space pushes now.

src/elements/SoftSol.H
```diff
@@ -502,6 +569,31 @@ namespace impactx::elements
             refpart.map(4,3) = R(4,3) - tau*alpha2*R(3,3);
             refpart.map(4,4) = R(4,4) - tau*alpha2*R(3,4);
 
+            // BELOW: if spin is needed only:
+            amrex::SmallMatrix<amrex::ParticleReal, 3, 6, amrex::Order::F, 1> const A = refpart.spin_coupling;
+            amrex::SmallMatrix<amrex::ParticleReal, 3, 1, amrex::Order::F, 1> const v = refpart.spin_rotation_vector;
+            amrex::SmallMatrix<amrex::ParticleReal, 3, 6, amrex::Order::F, 1> dA = {};
+            amrex::ParticleReal const gamma = refpart.gamma();
```
ax3l, line 583: Should this be only executed when spin tracking is on? Seems like we also run this code path with regular phase space pushes now.

src/elements/SoftSol.H
```diff
@@ -562,6 +654,15 @@ namespace impactx::elements
             refpart.map(4,3) = R(4,3)*cs - R(2,3)*sn;
             refpart.map(4,4) = R(4,4)*cs - R(2,4)*sn;
 
+            // BELOW: if spin is needed only:
+            amrex::SmallMatrix<amrex::ParticleReal, 3, 1, amrex::Order::F, 1> const v = refpart.spin_rotation_vector;
+            amrex::ParticleReal dv_z = 0_prt;
+            amrex::ParticleReal G = refpart.gyromagnetic_anomaly;
+
```
ax3l, line 668: Should this be only executed when spin tracking is on? Seems like we also run this code path with regular phase space pushes now.

src/elements/RFCavity.H
ax3l: (We can check the performance impact once the SIMD test CI passes again. This is only on the ref particle, so potentially no problem unless we optimize for ref particle tracking performance.)

cemitch99: Yes--for all the soft-edge elements, the calculation of the spin map here is performed simultaneously with the calculation of the linear map and reference particle push (once per slice, independent of the particle).  As a result, all these calculated constants are needed.  I don't know how best to branch/restructure things otherwise.  This is open for discussion.

ax3l:
I gotta merge in `development` on #1441 passes, and revert the CI changes here.

src/elements/RFCavity.H
ax3l: Clarified wih the benchmark above: https://github.com/BLAST-ImpactX/impactx/pull/1264#issuecomment-4146270884

The performance impact for the current benchmark (mapsteps=4, n_part=10k) is negligible right now. In practice for RFCavity and soft edge elements, we probably want to bump up the map steps more to like ~100. This then increases in these cases the cost of the ref element push but not the particle push itself, which can change the ratio a bit.

Not important to address now or here, since the influence is so small, as measured.

PR #1265, opened 2026-01-14
Spin push for an ideal solenoid

cemitch99:
This PR adds the update of the spin vector in an ideal solenoid.

- [x] add spin push operator in Sol.H

Multi-element benchmark tests including this feature will appear as follow-up.
See Issue https://github.com/BLAST-ImpactX/impactx/issues/1287


qianglbl (approved):

ax3l (approved):
LGTM.

We discussed that we can calculate all the constants that only depend on inputs and the reference particle in `compute_constants`, but we can move that also later on for all elements. Just let us not forget. Opened #1288

PR #1271, opened 2026-01-21
routine to instantiate an impactx lattice based on a synergia lattice

egstern:
This is the start of an example that demonstrates converting a lattice parsed in Synergia from a MAD-X into an ImpactX lattice.

- [x] encapsulate as optional feature
- [x] document API & example, as integrated test
- [x] cover in CI
- [x] final review of user-facing APIs

User-facing docs:
- API: https://impactx--1271.org.readthedocs.build/en/1271/usage/python.html#alternative-mad-x-parser-synmadx
- Example: https://impactx--1271.org.readthedocs.build/en/1271/usage/examples/simple_booster/README.html#variations


cemitch99:
Great!  Thanks for sharing!

cemitch99 (commented):
It looks like one feature still needed for this PR is documentation in a README.

egstern:
@ax3l The first failure is because the name of the author in an included Boost header is European and has a non-ASCII character in their name which just appears in a comment. As a European, how do you feel about flagging this as an error?

examples/simple_booster_synmadx/run_parse_booster_lattice.py
```diff
@@ -0,0 +1,259 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022-2023 ImpactX contributors
+# Authors: Eric G. Stern
+# License: BSD-3-Clause-LBNL
+#
+# -*- coding: utf-8 -*-
+
+
+import mpi4py.MPI as MPI
```
ax3l: This is ok, and ImpactX will use the communicator from MPI4PY that way, but it is not needed.
Will remove to keep dependencies minimal.

```suggestion
```

examples/simple_booster_synmadx/run_parse_booster_lattice.py
```diff
@@ -0,0 +1,259 @@
...
+turns = 1
+
+DEBUG = True
+
+total_booster_charge = 6.7e12
+filled_buckets = 81
+bunch_charge_C = eV * total_booster_charge / filled_buckets
+
+myrank = MPI.COMM_WORLD.rank
+
```
ax3l:
```suggestion
```

examples/simple_booster_synmadx/run_parse_booster_lattice.py
```diff
@@ -0,0 +1,259 @@
...
+    # above transition, phase needs to be (pi - phase) for longitudinal stability
+    if above_transition:
+        phase_set = np.pi - phase
+    else:
+        phase_set = phase
+
+    # Offset phase by how far the bunch has shifted
+    phase_set = phase_set + bunch_phase_offset
+
+    if DEBUG and myrank == 0:
```
ax3l:
```suggestion
    if DEBUG and amr.ParallelDescriptor.IOProcessor():
```

examples/simple_booster_synmadx/run_parse_booster_lattice.py
```diff
@@ -0,0 +1,259 @@
...
+            ", phase = ",
+            phase_set,
+            end="",
+        )
+    # count RF cavities
+    cavities = 0
+    for elem in lattice.get_elements():
+        if elem.get_type() == synmadx.element_type.rfcavity:
+            cavities = cavities + 1
+    if DEBUG and myrank == 0:
```
ax3l:
```suggestion
    if DEBUG and amr.ParallelDescriptor.IOProcessor():
```

examples/simple_booster_synmadx/run_parse_booster_lattice.py
```diff
@@ -0,0 +1,259 @@
...
+        if elem.get_type() == synmadx.element_type.rfcavity:
+            elem.set_double_attribute("volt", 1000 * voltage / cavities)
+            elem.set_double_attribute("lag", phase_set / (2 * np.pi))
+            elem.set_double_attribute("harmon", harmno)
+            elem.set_double_attribute(
+                "freq", rf_freq * 1.0e-6
+            )  # MAD-X convention frequency in MHz
+
+    for elem in lattice.get_elements():
+        if DEBUG and myrank == 0 and elem.get_type() == synmadx.element_type.rfcavity:
```
ax3l:
```suggestion
        if DEBUG and amr.ParallelDescriptor.IOProcessor() and elem.get_type() == synmadx.element_type.rfcavity:
```

examples/simple_booster_synmadx/run_parse_booster_lattice.py
```diff
@@ -0,0 +1,259 @@
...
+
+# ========================================================================
+
+
+def main():
+
+    # Read the lattice
+    lattice = get_lattice()
+
+    if myrank == 0:
```
ax3l:
```suggestion
    if amr.ParallelDescriptor.IOProcessor():
```

examples/simple_booster_synmadx/run_parse_booster_lattice.py
```diff
@@ -0,0 +1,259 @@
...
+
+    # Assume the lattice sets the reference particle
+    refpart = lattice.get_reference_particle()
+
+    energy = refpart.get_total_energy()
+    momentum = refpart.get_momentum()
+    gamma = refpart.get_gamma()
+    beta = refpart.get_beta()
+
+    if myrank == 0:
```
ax3l:
```suggestion
    if amr.ParallelDescriptor.IOProcessor():
```

examples/simple_booster_synmadx/run_parse_booster_lattice.py
```diff
@@ -0,0 +1,259 @@
...
+    sim.add_particles(bunch_charge_C, distr, npart)
+
+    # insert the converted MAD-X->Synergia lattice
+    init_monitor = False
+    final_monitor = True
+    ix_lattice = syn2_to_impactx(lattice, init_monitor, final_monitor)
+
+    # save the lattice
+
+    if myrank == 0:
```
ax3l:
```suggestion
    if amr.ParallelDescriptor.IOProcessor():
```

examples/simple_booster_synmadx/run_parse_booster_lattice.py
```diff
@@ -0,0 +1,259 @@
...
+# License: BSD-3-Clause-LBNL
+#
+# -*- coding: utf-8 -*-
+
+
+import mpi4py.MPI as MPI
+import numpy as np
+from scipy import constants
+
+from impactx import ImpactX, distribution, synmadx, twiss
```
ax3l:
```suggestion
from impactx import ImpactX, amr, distribution, synmadx, twiss
```

src/synmadx/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+set_target_properties(synmadx PROPERTIES POSITION_INDEPENDENT_CODE ON)
+
+target_include_directories(synmadx PUBLIC
+    ${CMAKE_CURRENT_SOURCE_DIR}
+)
+
+# Boost types (e.g., boost::variant in mx_expr.h) appear in public headers
+target_link_libraries(synmadx PUBLIC Boost::headers)
+
+target_compile_features(synmadx PUBLIC cxx_std_17)
```
ax3l: Since the rest of ImpactX is now on C++20, we can keep this consistent here:
```suggestion
target_compile_features(synmadx PUBLIC cxx_std_20)
```

CMakeLists.txt
```diff
@@ -542,6 +552,9 @@ if(ImpactX_PYTHON)
         VERBATIM
     )
     add_dependencies(pyImpactX_python_sources pyImpactX)
+    add_dependencies(${ImpactX_CUSTOM_TARGET_PREFIX}pip_wheel
+        pyImpactX_python_sources
+    )
```
ax3l: This looks odd... The sources should not depend on the wheel command...
Did the LLM mean to wrote

```suggestion
    if(ImpactX_SYNMADX)
        add_dependencies(pyImpactX_python_sources pysynmadx)
    endif()
```
?

CMakeLists.txt
```diff
@@ -478,7 +488,7 @@ if(ImpactX_PYTHON)
         WORKING_DIRECTORY
             ${ImpactX_BINARY_DIR}
         DEPENDS
-            pyImpactX
+            pyImpactX $<$<BOOL:${ImpactX_SYNMADX}>:pysynmadx>
```
ax3l: Confusing formatting:
```suggestion
            pyImpactX
            $<$<BOOL:${ImpactX_SYNMADX}>:pysynmadx>
```

src/synmadx/tests/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+# Fetch Catch2 v3 for unit tests
+include(FetchContent)
+FetchContent_Declare(
+    Catch2
+    GIT_REPOSITORY https://github.com/catchorg/Catch2.git
+    GIT_TAG        v3.5.2
+)
+FetchContent_MakeAvailable(Catch2)
```
ax3l: Similar to `ImpactX_openpmd_internal` et al., there should be an option to turn the fetch-content off and use `find_package` instead. Otherwise we cannot ship `ImpactX_SYNMADX` in package managers, run smoke/acceptance tests, without escaping dependency management in them.

src/synmadx/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+# install shared library
+install(TARGETS synmadx
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+)
+
+# Python bindings
+if(ImpactX_PYTHON)
+    pybind11_add_module(pysynmadx MODULE ${ImpactX_SOURCE_DIR}/src/python/synmadx.cpp)
+    target_link_libraries(pysynmadx PRIVATE synmadx)
```
ax3l, line 44: Codex:

>   1. [P1] The generated wheel omits libsynmadx.so.
     src/synmadx/CMakeLists.txt:42
>
>     synmadx_pybind dynamically requires libsynmadx.so, but the wheel contains only the extension. Its RPATH points back to the local build directory, so imports only work while that directory exists
>     and fail after relocation or on another machine. Package the shared library with a relocatable RPATH, or link it statically into the extension.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+    if ds_dipedge:
+        ixelem.append(ds_dipedge)
+
+    return ixelem
+
+
+# linear flag if true uses linearized models
+def cnv_quadrupole(elem, order):
+    ds = elem.get_length()
+    k1 = elem.get_double_attribute("k1", 0.0)
```
ax3l, line 331: Codex:

>   2. [P1] Skew quadrupole strength K1S is silently discarded.
     src/python/impactx/synmadx/syn2_to_impactx.py:331
>
>     cnv_quadrupole reads only k1. A valid quadrupole with K1=0, K1S=2 converts to an ExactQuad with zero strength and zero rotation, effectively tracking it as a drift. This produces silently
>     incorrect beam physics.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+        elif etype == ET.octupole:
+            ocelem = cnv_octupole(elem, order)
+            if isinstance(ocelem, list):
+                impactx_lattice.extend(ocelem)
+            else:
+                impactx_lattice.append(ocelem)
+        elif etype == ET.hkicker or etype == ET.vkicker or etype == ET.kicker:
+            # both H and V kickers handled by same routine since
+            # ImpactX has only one type of kicker element
+            impactx_lattice.append(cnv_kicker(elem, order))
```
ax3l, line 627: Codex:

>   3. [P2] Thick kickers create a nested list that cannot be added to the lattice.
     src/python/impactx/synmadx/syn2_to_impactx.py:627
>
>     For L>0, cnv_kicker returns [drift, kicker, drift], but this line uses append. Calling sim.lattice.extend() on the result raises a pybind cast error. It should flatten the returned list, as the
>     bend and RF conversion paths already do.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+
+    if order == Order.linear:
+        # There is no linear Octupole. Use a linear drift
+        ocelem = impactx.elements.Drift(L, name=nm)
+    else:
+        ocelem = impactx.elements.ExactMultipole(
+            ds=L,
+            k_normal=knorm,
+            k_skew=kskew,
+            order=4,
```
ax3l, line 555: Codex:

>   4. [P2] Exact octupole conversion always raises TypeError.
     src/python/impactx/synmadx/syn2_to_impactx.py:555
>
>     ExactMultipole has an int_order parameter, not order. Since exact tracking is the default, every MAD-X octupole currently fails during conversion.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+    knlvect = elem.get_vector_attribute("knl", [])
+    kslvect = elem.get_vector_attribute("ksl", [])
+    nlen = len(knlvect)
+    slen = len(kslvect)
+    maxlen = max(nlen, slen)
+    tilt = elem.get_double_attribute("tilt", 0.0)
+    rot = 180 * tilt / np.pi
+    if order == Order.linear:
+        # truncate at order 1
+        maxlen = max(maxlen, 2)  # 2 means dipole+quadrupole
```
ax3l, line 373: Codex:

>   5. [P2] Linear multipole conversion retains nonlinear orders and emits zero kicks.
     src/python/impactx/synmadx/syn2_to_impactx.py:373
>
>     max(maxlen, 2) expands rather than truncates the array, and the pass at line 382 does not skip zero components. For KNL={0,2,3}, Order.linear produces a zero dipole, quadrupole, and sextupole
>     instead of only the nonzero linear component.

src/synmadx/synergia/lattice/mx_tree.cc
```diff
@@ -0,0 +1,833 @@
...
+            subline.element_name((op > 0) ? (i) : (ne - 1 - i)));
+      }
+    }
+    // TODO: for now, we accept the sequence name as a simple line member
+    // Needs more works!
+    else if (type == ENTRY_SEQUENCE) {
+      line.insert_element(name);
+    }
+    // something we dont support
+    else {
```
ax3l, line 241: Codex:

> **[P1] Unknown or forward-referenced line members are silently omitted.**
>
> `line=(d, misspelled, d)` parses successfully and produces only the two drifts. Unknown members should be deferred for forward resolution or rejected, never dropped.

src/synmadx/synergia/lattice/madx_reader.cc
```diff
@@ -0,0 +1,364 @@
...
+    else
+      r = 0.0;
+
+    for (int i = 0; i < sequence.element_count(); ++i) {
+
+      double at = sequence.element_at(i);
+      double from = sequence.element_from(i);
+      std::string label = sequence.element_label(i);
+
+      if (sequence.element_type(i) == ENTRY_SEQUENCE) {
```
ax3l, line 99: Codex:

> **[P1] Nested sequences are inserted at the wrong parent position.**
>
> The recursive call appends the subsequence immediately without inserting the drift up to its parent `AT`. A 10 m parent with a 2 m subsequence at `AT=4` becomes a 6 m lattice.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+            else:
+                impactx_lattice.append(mpelem)
+        elif etype == ET.monitor:
+            ix_elem = cnv_monitor(elem)
+            impactx_lattice.append(ix_elem)
+        elif etype == ET.marker:
+            ix_elem = cnv_marker(elem)
+            impactx_lattice.append(ix_elem)
+        else:
+            print("warning: unsupported element: ", etype)
```
ax3l, line 652: Codex:

> **[P1] Unsupported parsed elements produce an incomplete lattice instead of failing.**
>
> For example, a 2 m solenoid converts to an empty list after printing a warning. Instruments, collimators, `HMONITOR`, and `VMONITOR` have similar paths. Until implemented, non-no-op elements should raise so callers cannot unknowingly track an incomplete lattice.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+# MP114BS: MULTIPOLE, TYPE=RGF021_S_BODY, KSL={-0.16E-4,  -0.808E-3*DIR,
+# -0.064894, -15.028637*DIR};
+
+
+#     Eric
+
+
+def cnv_sextupole(elem, order):
+    L = elem.get_length()
+    k2 = elem.get_double_attribute("k2", 0.0)
```
ax3l, line 509: Codex:

> **[P1] `K2S` and `K3S` remain silently discarded.**
>
> The sextupole and octupole converters derive skew strength only from `TILT`; explicit MAD-X `K2S`/`K3S` attributes are never read. A sextupole with `K2=0, K2S=2` converts to an all-zero multipole.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+        rotation=rot,
+    )
+    return ix_elem
+
+
+# linear flag if true uses linearized ImpactX models
+def cnv_sbend(elem, order):
+    bendangle = elem.get_bend_angle()
+    length = elem.get_length()
+    radius_of_curvature = length / bendangle
```
ax3l, line 104: Codex:

> **[P2] Valid zero-curvature elements divide by zero or create NaNs.**
>
> `SBEND, ANGLE=0` raises `ZeroDivisionError`, `RBEND, ANGLE=0` creates an `ExactSbend` with `ds=NaN`, and `DIPEDGE, H=0` also divides by zero. These should become appropriate straight elements or markers.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+        etype = edict["type"]
+        output_elem = f"impactx.elements.{etype}("
+        firstparm = True
+
+        # for ExactSbend the member phi which is the angle has been converted into radians but
+        # I need to convert it back to degrees which is what the contructor needs.
+        if etype == "ExactSbend":
+            edict["phi"] = edict["phi"] * 180 / np.pi
+
+        if etype == "DipEdge" or etype == "ShortRF" or etype == "BeamMonitor":
```
ax3l, line 683: Codex:

> **[P2] `unroll_impactx_lattice` does not generate recreatable source.**
>
> A `Kicker` emits unsupported `ds` and an unquoted `unit=dimensionless`; a `Multipole` emits unsupported `ds`; and every `BeamMonitor` is deliberately omitted. This contradicts the documented ‘recreates the lattice’ behavior.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+            k_skew=kskew,
+            order=4,
+            nslice=nslice_by_elem_type["octupole"],
+            name=nm,
+        )
+
+    return ocelem
+
+
+def cnv_monitor(elem):
```
ax3l, line 563: Codex:

> **[P2] Finite-length monitors lose their length.**
>
> `MONITOR, L=2` becomes only a zero-length `BeamMonitor`, shortening the lattice by 2 m. It needs a drift plus monitor; the dispatch should also include horizontal and vertical monitor variants.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+            model=de_model,
+            name=nm + "_usedge",
+            rotation=rot,
+        )
+    if e2 != 0.0:
+        ds_dipedge = impactx.elements.DipEdge(
+            e2,
+            radius_of_curvature,
+            2 * hgap,
+            K2=fint,
```
ax3l, line 142: Codex:

> **[P2] `FINTX` is ignored for bend exit edges.**
>
> Both edges receive `FINT`. With `FINT=0.2, FINTX=0.9`, the converted exit edge still has `K2=0.2`, producing incorrect exit fringe focusing.

src/synmadx/synergia/lattice/mx_tree.cc
```diff
@@ -0,0 +1,833 @@
...
+    if (attr.type() == MX_ATTR_STRING) {
+      t.insert_attribute(attr.name(), any_cast<string>(attr.value()));
+    } else if (attr.type() == MX_ATTR_PREDEFINED) {
+      t.insert_attribute(attr.name(), any_cast<mx_keyword>(attr.value()).name);
+    } else if (attr.type() == MX_ATTR_NUMBER) {
+#if 0
+        mx_expr e = simplify( any_cast<mx_expr>(attr.value()), mx );
+        t.insert_attribute( attr.name(), e );
+#endif
+      t.insert_attribute(attr.name(), any_cast<mx_expr>(attr.value()));
```
ax3l, line 48: Codex:

> **[P1] `=` and `:=` have identical deferred-evaluation semantics.** Both ordinary and lazy numeric attributes are stored as expressions. Thus `a=1; k=2*a; a=3;` yields `k=6`, whereas MAD-X requires eager `=` to preserve `k=2`.

src/synmadx/synergia/lattice/lattice_element.cc
```diff
@@ -0,0 +1,818 @@
...
+  // dynamic lattice
+  //
+  // default the references to 0.0 when evaluating the lazy value.
+  //
+  // this will set undefined variables to 0.0 instead of throwing
+  // an excpetion for "undefined reference".
+  //
+  if (lattice_ptr && lattice_ptr->is_dynamic_lattice()) {
+    auto const& mx = lattice_ptr->get_lattice_tree().mx;
+    return mx_eval(lr->second, mx, 0.0);
```
ax3l, line 397: Codex:

> **[P1] Undefined expression references silently become zero.** Evaluating with a `0.0` fallback means a typo such as `k1=missing_strength` produces a zero-strength quadrupole instead of failing.

src/synmadx/synergia/lattice/madx.cc
```diff
@@ -0,0 +1,1093 @@
...
+      bool found = false;
+
+      for (seq_ele_v_t::const_iterator cit = seq_.begin(); cit != seq_.end(); ++cit)
+      {
+        if (cit == it) continue;
+
+        if (cit->label == it->from_str)
+        {
+          double at = boost::apply_visitor(mx_calculator(parent), cit->at);
+          it->from = at;
```
ax3l, line 580: Codex:

> **[P1] Chained sequence `FROM` references use the referenced element's raw `AT`.** This omits that element's own resolved `FROM` offset. For `b,at=2,from=a; c,at=1,from=b`, `c` is placed relative to raw `b.at`, not `b`'s actual position.

src/synmadx/synergia/lattice/madx_reader.cc
```diff
@@ -0,0 +1,364 @@
...
+    return sequence.length();
+  }
+
+  void
+  extract_reference_particle(Lattice& lattice, MadX const& mx)
+  {
+    try {
+      synergia::MadX_command command(mx.command("beam"));
+      std::vector<std::string> attributes(command.attribute_names());
+      double mass = 0, charge = 0, energy = 0, pc = 0, gamma = 0;
```
ax3l, line 190: Codex:

> **[P1] MAD-X `BEAM,BV=-1` is discarded.** Only mass, charge, and momentum information is transferred into the lattice. Consequently bend angles/strengths, kicker kicks, and RF voltage are not polarity-reversed.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+
+# linear flag if true uses linearized ImpactX models
+def cnv_sbend(elem, order):
+    bendangle = elem.get_bend_angle()
+    length = elem.get_length()
+    radius_of_curvature = length / bendangle
+    nm = elem.get_name()
+    k1 = elem.get_double_attribute("k1", 0.0)
+    k1s = elem.get_double_attribute("k1s", 0.0)
+    k2 = elem.get_double_attribute("k2", 0.0)
```
ax3l, line 108: Codex:

> **[P1] SBEND/RBEND bodies still drop supported multipole components.** This reads only `K1`, `K1S`, and `K2`; an exact bend containing only `K2S`, `K3`, or `K3S` becomes a plain bend with those fields silently removed.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+
+    return ix_elem
+
+
+def cnv_multipole(elem, order):
+    # ImpactX multipole elements only have a single order each so
+    # we have to peel each order and possibly create multiple elements to get the
+    # full description.
+    nm = elem.get_name()
+    knlvect = elem.get_vector_attribute("knl", [])
```
ax3l, line 364: Codex:

> **[P1] Finite-length MAD-X MULTIPOLEs become zero-length kicks.** This conversion never reads `elem.get_length()`. A multipole with `L=2` therefore produces a converted lattice with total `ds=0`.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+
+def cnv_rfcavity(elem, refpart, order):
+    # I don't have a linear model of the RF cavity itself, but if there are
+    # drifts surrounding it, they should be linear if the linear model is
+    # requested
+    mp = refpart.get_mass()
+    # The cavity voltage is in relationship to the reference particle mass
+    # The MAD-X convention is for the voltage to be in MV
+    rfvolt = elem.get_double_attribute("volt", 0.0) / 1000.0  # convert to GV
+    freq = elem.get_double_attribute("freq", 0.0) * 1.0e6  # get the freq in Hz
```
ax3l, line 399: Codex:

> **[P1] RFCAVITY ignores `HARMON` and `LAGTAP`.** Using only explicit `FREQ` and `LAG` turns a normal `HARMON`-only cavity into a `ShortRF` with `freq=0`; `LAGTAP` is also absent from the phase.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+    ds_dipedge = None
+    cf = (k1 != 0.0) or (k2 != 0.0) or (k1s != 0.0)
+
+    # for RBENDS, the dipedge angles are relative to bendangle/2 and
+    # have opposite sense depending on the sign of the bend.
+    if bendangle > 0.0:
+        e1 = e1 + bendangle / 2
+        e2 = e2 + bendangle / 2
+    else:
+        e1 = -e1 - bendangle / 2
```
ax3l, line 237: Codex:

> **[P1] Negative RBEND edge angles have the wrong sign.** With `ANGLE=-0.2,E1=E2=0`, this emits `psi=+0.1`; the rectangular geometry requires `E+ANGLE/2=-0.1`.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+
+
+def cnv_rbend(elem, order):
+    # the RBEND is converted to an SBEND with dipedge elements
+    # before an after to get parallel faces
+
+    bendangle = elem.get_bend_angle()
+    straight_length = elem.get_length()
+    # MAD-X RBEND langths is the straight length.
+    radius_of_curvature = straight_length / (2 * np.sin(0.5 * bendangle))
```
ax3l, line 214: Codex:

> **[P1] RBEND always treats `L` as a chord.** This performs chord-to-arc conversion unconditionally, ignoring `RBARC=false`, where MAD-X already defines `L` as arc length.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+
+
+class Order(Enum):
+    linear = 0
+    chr = 1
+    exact = 2
+
+
+# set the following switch overrides the model for dipedges
+force_linear_dipedges_for_bends = True
```
ax3l, line 31: Codex:

> **[P1] `Order.exact` silently uses linear bend edges.** This global override contradicts the documented `order` API. If the downgrade is intentional, it needs an explicit option and warning.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+            # some lattice files use vkick attribute instead of kick
+            ykick = elem.get_double_attribute("vkick")
+
+    elif elem.get_type() == ET.kicker:
+        if elem.has_double_attribute("hkick"):
+            xkick = elem.get_double_attribute("hkick")
+        if elem.has_double_attribute("vkick"):
+            ykick = elem.get_double_attribute("vkick")
+
+    kickelem = impactx.elements.Kicker(xkick=xkick, ykick=ykick, name=ename)
```
ax3l, line 452: Codex:

> **[P1] Kicker rotation and compensated kicks are discarded.** The ImpactX kicker is constructed without `rotation`, and `CHKICK`/`CVKICK` are never included. A `TILT=0.5` kicker currently converts with `rotation=0`.

src/python/impactx/synmadx/syn2_to_impactx.py
```diff
@@ -0,0 +1,1491 @@
...
+    rot = 180 * tilt / np.pi
+    model = " "
+    if order == Order.linear:
+        model = "linear"
+    elif order == Order.exact or order == Order.chr:
+        model = "nonlinear"
+    ##
+    ## ImpactX has the location= argument which can be either 'entry' or 'exit'
+    ## but MAD-X doesn't. How do we resolve this?
+    ix_elem = impactx.elements.DipEdge(
```
ax3l, line 88: Codex:

> **[P2] Standalone DIPEDGE ignores `ENTRANCE`.** This always accepts ImpactX's default `"entry"` location. `ENTRANCE=false` therefore produces an entry edge, particularly wrong when the nonlinear model is selected.

src/synmadx/synergia/lattice/mx_tree.cc
```diff
@@ -0,0 +1,833 @@
...
+      for (int z = 0; z < repeat; ++z) {
+        for (size_t i = 0; i < ne; ++i)
+          line.insert_element(
+            subline.element_name((op > 0) ? (i) : (ne - 1 - i)));
+      }
+    }
+    // TODO: for now, we accept the sequence name as a simple line member
+    // Needs more works!
+    else if (type == ENTRY_SEQUENCE) {
+      line.insert_element(name);
```
ax3l, line 238: Codex:

> **[P2] A SEQUENCE used inside a LINE is accepted and then crashes.** Inserting the sequence name as though it were an element makes `get_lattice()` later raise `cannot find command with label ...`. It should expand the sequence or reject the construct during parsing.

src/synmadx/synergia/lattice/madx.cc
```diff
@@ -0,0 +1,1093 @@
...
+  }
+}
+
+void
+  MadX::insert_line(string_t const & name, MadX_line const & line)
+{
+  string_t key(name);
+  std::transform( key.begin(), key.end(), key.begin(), ::tolower );
+
+  lines_.insert( std::make_pair(key, line) );
```
ax3l, line 972: Codex:

> **[P2] Repeated LINE and SEQUENCE definitions are silently ignored.** These definitions use `map::insert`, unlike element definitions which explicitly overwrite. The parser should either apply the new definition or report a duplicate.

src/synmadx/synergia/lattice/madx_reader.cc
```diff
@@ -0,0 +1,364 @@
...
+std::vector<std::string>
+MadX_reader::get_sequence_names() const
+{
+  return madx.sequence_labels();
+}
+
+std::vector<std::string>
+MadX_reader::get_all_names() const
+{
+  std::vector<std::string> retval(madx.line_labels());
```
ax3l, line 266: Codex:

> **[P3] `get_all_names()` does not return element names.** It returns only line and sequence labels, despite the Python binding documenting "Get all element names." For `d:drift; ring:line=(d)`, it returns only `["ring"]`.

src/synmadx/CMakeLists.txt
ax3l: changed to a static `libsynmadx.a` to avoid these.

ax3l (approved):
This is ok now from an integration standpoint :)

We can fix physics issues in #1584 / follow-up PRs and ideally transition to our new all-Python parser. For now, the physics for the booster works in both parsers, which is what counts as the use case.

src/synmadx/synergia/lattice/mx_parse.cc
```diff
@@ -0,0 +1,782 @@
+
```
ax3l, line 1: Codex:

> **[P1] Resolve the imported Synergia code’s licensing.**
>
> The ported sources under `src/synmadx/synergia` contain no copyright or license notices, while upstream Synergia is GPL-2.0-or-later and requires derivative works to retain notices and licensing terms. This needs explicit relicensing permission or proper GPL attribution/compliance confirmed by the project before merging into a BSD-distributed tree. See the upstream [Synergia license](https://github.com/fnalacceleratormodeling/synergia2/blob/devel3/LICENSE).

src/python/synmadx.cpp
```diff
@@ -0,0 +1,779 @@
...
+             "Get MAD-X representation",
+             "sanitize"_a = false)
+
+        .def("__repr__",
+             &Lattice_element::as_string)
+        ;
+
+    // ---- Lattice_element_slice ----
+    py::class_<Lattice_element_slice>(m, "Lattice_element_slice")
+        .def(py::init<Lattice_element const&>(),
```
ax3l, line 528: Codex:

> **[P1] `Lattice_element_slice` bindings permit use-after-free.**
>
> These constructors accept an element by reference, but the slice stores only a raw pointer and the bindings provide no `py::keep_alive`. Constructing a slice from a temporary element leaves a dangling pointer; a minimal reproduction hung/crashed when accessing it. The constructors need lifetime management—or the slice must own its element—and `get_lattice_element()` should return with `reference_internal`.

src/synmadx/synergia/lattice/mx_tree.cc
```diff
@@ -0,0 +1,833 @@
...
+
+void
+mx_if::interpret(MadX& mx)
+{
+  if (!if_.valid())
+    throw runtime_error("mx_if::interpret() Invalid if command");
+
+  if (if_.evaluate_logic(mx)) {
+    if_.interpret_block(mx);
+  } else if (!elseif_.empty()) {
```
ax3l, line 661: Codex:

> **[P1] `IF`/`ELSEIF`/`ELSE` execution is incorrect.**
>
> The presence of any `ELSEIF` prevents `ELSE` from ever executing, even when every condition is false. There is also no `break`, so multiple true `ELSEIF` branches execute instead of only the first.

src/synmadx/synergia/lattice/mx_tree.cc
```diff
@@ -0,0 +1,833 @@
...
+void
+mx_while::assign(mx_logic const& logic, mx_tree const& block)
+{
+  while_ = mx_if_block(logic, block);
+}
+
+void
+mx_while::interpret(MadX& mx)
+{
+  return;
```
ax3l, line 710: Codex:

> **[P1] Parsed `WHILE` loops are silently ignored.**
>
> `mx_while::interpret()` is an unconditional return. For example, `i=0; while(i<1){i=1;}` leaves `i` equal to zero. Either execute the loop with appropriate safeguards or reject `WHILE` explicitly.

src/synmadx/synergia/lattice/mx_parse.cc
```diff
@@ -0,0 +1,782 @@
...
+        primary = real[_val = _1] |
+                  ('(' >> expr >> ')')[_val = phx::construct<nop_t>(_1, true)] |
+                  (uop >> primary)[_val = phx::construct<uop_t>(_1, _2)] |
+                  (no_case[constant] >> !char_(".a-zA-Z_0-9"))[_val = _1] |
+                  (no_case[ufunc] >> '(' >> expr >>
+                   ')')[_val = phx::construct<uop_t>(_1, _2)] |
+                  (no_case[bfunc] >> '(' >> expr >> ',' >> expr >>
+                   ')')[_val = phx::construct<bop_t>(_1, _2, _3)] |
+                  (lit("table") >> '(' >> name >> ',' >> name >>
+                   ')')[_val = 1.0] // eat table()
```
ax3l, line 278: Codex:

> **[P1] Every `TABLE()` lookup silently evaluates to `1.0`.**
>
> The grammar accepts `TABLE(name,column)` but hard-codes its value to `1.0`. That can silently produce incorrect strengths and placements. Unsupported table access should fail clearly until implemented.

src/synmadx/synergia/lattice/mx_tree.cc
```diff
@@ -0,0 +1,833 @@
...
+    double mass = 0, charge = 0, energy = 0, pc = 0, gamma = 0;
+    bool have_charge = false;
+    for (attrs_t::const_iterator it = attrs_.begin(); it != attrs_.end();
+         ++it) {
+      if (it->name() == "particle") {
+        string particle = any_cast<mx_keyword>(it->value()).name;
+        if (particle == "proton") {
+          mass = pconstants::mp;
+          charge = pconstants::proton_charge;
+        } else if (particle == "antiproton") {
```
ax3l, line 509: Codex:

> **[P2] Neither accepted spelling of antiproton works.**
>
> The parser accepts `anti-proton`, while this code checks for `antiproton`. The former reports an unknown particle and the latter fails during parsing/value conversion. Normalize both spellings to one internal value.

src/synmadx/synergia/lattice/madx_reader.cc
```diff
@@ -0,0 +1,364 @@
...
+  return get_lattice(line_name);
+}
+
+Lattice
+MadX_reader::get_lattice(std::string const& line_name, synergia::MadX const& mx)
+{
+  Lattice lattice(line_name);
+
+  std::vector<std::string> lines(mx.line_labels());
+  if (std::find(lines.begin(), lines.end(), line_name) != lines.end()) {
```
ax3l, line 309: Codex:

> **[P2] Lattice lookup incorrectly becomes case-sensitive.**
>
> This compares the caller’s unnormalized name against normalized lowercase names. Consequently, a line declared as `Ring` is listed as `ring`, but `get_lattice("RING")` fails even though MAD-X identifiers are case-insensitive.

examples/CMakeLists.txt
```diff
@@ -1687,6 +1687,20 @@ add_impactx_test(simple-booster.py
...
+# Fermilab Booster (synmadx MAD-X Parser) #####################################
+#
+if(ImpactX_SYNMADX)
+    # copy MAD-X lattice file
+    file(COPY ${ImpactX_SOURCE_DIR}/examples/simple_booster_synmadx/sbbooster-cooked.madx
+            DESTINATION ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/simple-booster-synmadx.py)
+    add_impactx_test(simple-booster-synmadx.py
+        examples/simple_booster_synmadx/run_parse_booster_lattice.py
+        OFF  # ImpactX MPI-parallel
+        OFF  # no analysis script yet
```
ax3l, line 1699: Codex:

> **[P2] The primary conversion example has no correctness check.**
>
> This test explicitly has no analysis script. CI only verifies that conversion/tracking finishes, so numerically wrong element strengths, positions, or optics can pass. It needs numerical assertions against known lattice or Twiss results, plus focused parser tests for the cases above.

src/synmadx/synergia/lattice/madx_reader.cc
```diff
@@ -0,0 +1,364 @@
...
+            element.set_vector_attribute(*it,
+                                         cmd.attribute_as_number_seq(*it, 0.0));
+            break;
+          default:
+            throw std::runtime_error("unable to process attribute " + *it +
+                                     " of element " + name);
+        }
+      }
+
+      double drift_length =
```
ax3l, line 145: Codex:

> **[P1] Dynamic lattice updates corrupt sequence geometry.**
>
> This evaluates element positions and generated drift lengths only once. Changing a dynamic element length from 0.2 m to 0.4 m made a declared 1 m sequence become 1.2 m because both surrounding drifts remained 0.4 m. Geometry-dependent expressions must trigger sequence reconstruction or be rejected as dynamic parameters.

src/synmadx/synergia/lattice/lattice_tree.cc
```diff
@@ -0,0 +1,44 @@
...
+Lattice_tree::set_variable(std::string const& name, double val)
+{
+  mx.insert_variable(name, mx_expr(val));
+}
+
+void
+Lattice_tree::set_variable(std::string const& name, std::string const& val)
+{
+  synergia::mx_expr expr;
+  synergia::parse_expression(val, expr);
```
ax3l, line 16: Codex:

> **[P1] Dynamic setters silently accept malformed expressions.**
>
> This ignores the return value from `parse_expression()`. Setting a variable to `"1+"` succeeded and changed an element’s value from 2 to 1 because the parser retained the valid prefix. Both string setters must throw when parsing does not consume a valid complete expression.

src/synmadx/synergia/lattice/madx_reader.cc
```diff
@@ -0,0 +1,364 @@
...
+        at + from - current_pos - element.get_length() * (1.0 - r);
+      if (drift_length > min_drift_length) {
```
ax3l, line 147: Codex:

> **[P1] Overlapping sequence elements silently change the lattice length.**
>
> This simply omits a negative inter-element drift. A declared 1 m sequence containing two overlapping 0.8 m quadrupoles became a serial 1.6 m lattice. Since ImpactX cannot represent overlapping elements this way, conversion must reject the sequence rather than silently changing its geometry.

src/synmadx/synergia/lattice/mx_tree.cc
```diff
@@ -0,0 +1,833 @@
...
+    execute(mx);
+
+    // mx.insert_command(new_cmd);
+  }
+
+  return;
+}
+
+void
+mx_command::execute(MadX& mx)
```
ax3l, line 395: Codex:

> **[P1] Most accepted executable commands are silent no-ops.**
>
> This implements only `CALL`, `SEQUENCE`, `ENDSEQUENCE`, and `BEAM`; commands such as `SEQEDIT`, `REMOVE`, `INSTALL`, and `RETURN` fall through silently. A reproduction using `SEQEDIT; REMOVE, ELEMENT=d; ENDEDIT;` left `d` in the lattice. Unsupported commands must produce an explicit error until implemented.

src/synmadx/synergia/lattice/mx_tree.cc
```diff
@@ -0,0 +1,833 @@
...
+{
+  if (keyword_ == "call") {
+    // pull in the sub-file
+    for (attrs_t::const_iterator it = attrs_.begin(); it != attrs_.end();
+         ++it) {
+      if (it->name() == "file") {
+        string fname = std::any_cast<string>(it->value());
+        mx_tree subroutine;
+        parse_int_madx_file(fname, subroutine);
```
ax3l, line 404: Codex:

> **[P1] Relative `CALL` paths resolve against the process working directory.**
>
> This opens the raw `FILE` value without retaining the including file’s directory. Thus `/path/main.madx` containing `CALL, FILE="sub/elements.madx"` fails unless the caller happens to run from `/path`. Nested includes need an include-directory stack; the repository’s maintained parser already tests this behavior in `tests/python/test_madx_call.py`.

src/synmadx/synergia/lattice/mx_expr.cc
```diff
@@ -0,0 +1,254 @@
...
+    case op_tag::sub:
+      return apply_visitor(*this, b.lhs) + "-" + apply_visitor(*this, b.rhs);
+
+    case op_tag::mul:
+      return apply_visitor(*this, b.lhs) + "*" + apply_visitor(*this, b.rhs);
+
+    case op_tag::div:
+      return apply_visitor(*this, b.lhs) + "/" + apply_visitor(*this, b.rhs);
+
+    case op_tag::pow:
```
ax3l, line 201: Codex:

> **[P2] `^` expressions cannot be printed or exported.**
>
> The grammar tags the operator as `pow_op`, but this handles only `pow`. A dynamic lattice with `l=x^2` evaluated correctly to 4, but `repr(lattice)` raised `mx_expr_writer(): invalid bop_t operator`; MAD-X export fails through the same writer.

src/synmadx/synergia/lattice/lattice_element.cc
```diff
@@ -0,0 +1,818 @@
...
+      continue;  // skip this attribute if it is not madx kosher
+    }
+    ss << ", " << attr.first << "=" << std::setprecision(16) << mx_expr_str(attr.second);
+  }
+
+  for (auto const& attr : string_attributes) {
+    if (sanitize && non_madx_string_attributes.count(attr.first)) {
+      continue; // skip this attriute if it is not madx kosher
+    }
+    ss << ", " << attr.first << "=" << attr.second;
```
ax3l, line 801: Codex:

> **[P2] MAD-X export does not quote string attributes.**
>
> This emits values such as `file=diag.h5` instead of `file="diag.h5"`. Round-tripping that monitor changed `file` from a string into a numeric expression evaluating to zero. String values need quoting and escaping.

src/synmadx/synergia/lattice/madx_reader.cc
```diff
@@ -0,0 +1,364 @@
...
+                  MadX const& mx,
+                  std::string const& line_name,
+                  bool dynamic = false)
+  {
+    MadX_sequence const& sequence = mx.sequence(line_name);
+
+    double r = 0.5;
+    double total_length = sequence.length();
+    double current_pos = 0.0;
+    const double min_drift_length = 1.0e-6;
```
ax3l, line 81: Codex:

> **[P2] Valid sequence gaps of 1 µm or less are discarded.**
>
> This uses an absolute `1e-6` m minimum drift. A sequence with a real 0.5 µm entrance gap was shortened to `0.9999995` m and the element was shifted upstream. Positive gaps should be preserved; any roundoff tolerance should only handle numerically insignificant negative residuals.

PR #1276, opened 2026-01-24
Add spin moments to reduced diagnostics

cemitch99:
This PR adds the calculation of spin moments `mean_sx`, `mean_sy`, `mean_sz` and
`sigma_sx`, `sigma_sy`, `sigma_sz` to the list of reduced diagnostics when spin tracking is turned on.

- [x] Select whether to output spin moments in `DiagnosticOutput.cpp`.
- [x] Add calculation of spin moments in `ReducedBeamCharacteristics.cpp`.
- [x] Add benchmark test.
- [x] Update documentation.
- [x] Resolve issue introduced in min/max diagnostics.

Our previous automated tests do not probe the reduced diagnostics output file carefully, so it would be easy for errors to slip through the testing.  Let's take great care to make sure this doesn't break any of the diagnostics calculations by checking for consistency with old results.


cemitch99:
Note:  This PR appears to introduce a problem with `min_x`, `max_x`, `min_y`, `max_y`, `min_t`, `max_t`, while the other diagnostics appear consistent with previous results.

examples/CMakeLists.txt
```diff
@@ -1871,11 +1871,13 @@ add_impactx_test(quad-spin
...
     ON   # ImpactX MPI-parallel
     examples/spin_tracking/analysis_quad_spin.py
+    examples/spin_tracking/analysis_quad_spin_rbc.py
     OFF  # no plot script yet
 )
 add_impactx_test(quad-spin.py
     examples/spin_tracking/run_quad_spin.py
     ON    # ImpactX MPI-parallel
     examples/spin_tracking/analysis_quad_spin.py
+    examples/spin_tracking/analysis_quad_spin_rbc.py
```
ax3l: This syntax here should not work, did you mean to add a new test or replace the old test file?

(Maybe the script right not takes your 2nd analysis script as the plot script and it works because of that xD)

cemitch99: Oh, I was trying to be efficient by avoiding a second run using the same input.  I just split this out into a new test.  (This was already tested locally.)

src/diagnostics/ReducedBeamCharacteristics.cpp
```diff
@@ -49,13 +49,13 @@ namespace impactx::diagnostics
          * https://stackoverflow.com/questions/55136414/constexpr-variable-captured-inside-lambda-loses-its-constexpr-ness
          */
         // numbers of same type reduction operations in first concurrent batch
-        static constexpr std::size_t num_red_ops_1_sum = 7;  // summation
+        static constexpr std::size_t num_red_ops_1_sum = 13;  // summation
```
ax3l: 7+3 new ones for the sum:
```suggestion
        static constexpr std::size_t num_red_ops_1_sum = 10;  // summation
```

src/diagnostics/ReducedBeamCharacteristics.cpp
```diff
@@ -85,11 +90,17 @@ namespace impactx::diagnostics
...
+                const amrex::ParticleReal p_sz_mean = p_sz * p_w;
+
                 return {p_w,
                         p_x_mean, p_y_mean, p_t_mean,
                         p_px_mean, p_py_mean, p_pt_mean,
+                        p_sx_mean, p_sy_mean, p_sz_mean,
+                        p_x, p_y, p_t, p_px, p_py, p_pt,
                         p_x, p_y, p_t, p_px, p_py, p_pt,
-                        p_x, p_y, p_t, p_px, p_py, p_pt};
+                        p_sx, p_sy, p_sz};
```
ax3l: This stays unchanged:
```suggestion
                        p_x, p_y, p_t, p_px, p_py, p_pt};
```

src/diagnostics/ReducedBeamCharacteristics.cpp
cemitch99: This works, but it doesn't resolve the min/max problem.

src/diagnostics/ReducedBeamCharacteristics.cpp
cemitch99: This works, but it doesn't resolve the min/max problem.

cemitch99:
The min/max problem now appears to be fixed.  I did a diff of `diags/reduced_diagnostics`... output for the following tests (comparing before and after the PR), and I see _identical_ results:

- `examples/fodo/input_fodo.in` and `run_fodo.py`
- `examples/chicane/input_chicane.in` and `run_chicane.py`
- `examples/fodo_space_charge/input_fodo_envelope_sc.in`
- `examples/coupled_optics/input_coupled_optics.in`

cemitch99:
Example illustrating spin depolarization in 100 periods of a 1 m FODO channel:

SpinPolarization_FODO_Channel

ax3l (approved):
This looks great and consistent :+1:

PR #1002, opened 2025-06-23
CPU SIMD Support

ax3l:
Until recently (see AMReX PR https://github.com/AMReX-Codes/amrex/pull/4520), AMReX does not have a concept to help users write effective SIMD code on CPU, besides relying on auto-vectorization and pragmas, which are unreliable for any complex enough code. [1]

Lucky enough, C++ `std::datapar`/`std::simd` was just accepted into C++26, which gives an easy in to write portable SIMD/scalar code. Yet, I did not find a compiler/stdlib yet with support for it, so I implemented in AMReX using C++17 `<experimental/simd>` headers a new addition to the `ParallelFor` performance portability construct using the https://github.com/mattkretz/vir-simd library.

This PR vectorizes CPU code in ImpactX for particle tracking, both for NOACC and OpenMP compute backends, in ImpactX for element pushes, keeping our portable, single-source approach.

## Additional background

The implementation here works with any kind of vectorization (if implemented in the respective compiler): SSE(2), AVX(2), AVX512, Neon, ... 512-bit wide vector registers are the largest currently available, being able to push up to 8 (DP) or 16 (SP) particles in parallel, where a non-data-parallel (non-SIMD) implementation would push just 1 (in the worst case, for any complex enough code [1]).

[1] As written in the [story behind Intel's iscp compiler](https://pharr.org/matt/blog/2018/04/18/ispc-origins) and credited to [Tim Foley](http://graphics.stanford.edu/~tfoley/), *auto-vectorization is not a programming model.*

Vectorization is really a technique that primarily benefits us for elements that are very compute heavy (i.e., not memory-bandwidth bound). Not coincidentally, ImpactX focuses on symplectic, (high-order/exact) methods for chromatic and exact tracking, where this is exactly the case. Optimizations like vectorization thus makes a huge impact for our community/users/time-to-solution, and exactly address our most costly tracking methods, making it highly benefitial for our users to rely on them and still get fast results when needed.

### Related Reads for AMD/Intel GPUs with AVX-512

For AVX-512, Intel famously messed up in their early/current chips the clock frequency/throughput for AVX-512. That means in benchmarks one should run long and with many particles for reliable results, but I kept it simple above. The consequence is that sometimes where SIMD "looks meh/bad" it could be (even) better in practice.
AMD Zen5 CPUs do not have that flaw in their AVX-512 implementation.
Details: https://chipsandcheese.com/p/zen-5s-avx-512-frequency-behavior

## To Do

- [x] update install docs
[...]

PR #1279, opened 2026-01-27
`BeamOptics`: Generalize SIMD Logic

ax3l:
Generalize the SIMD logic for load/store with a bit of meta-programming and perfect forwarding, to avoid duplicate code paths to launch the single/simd particle methods of an element.

Exploring this generalization for reuse of the SIMD patterns in other BLAST codes. Generalized versions of `load_pdata` and `store_pdata` might move into `AMReX_SIMD.H` afterwards.

- [x] generalize in AMReX via https://github.com/AMReX-Codes/amrex/pull/4924

Follow-up to #1002

[test_performance.sh](https://github.com/user-attachments/files/25034084/test_performance.sh)
- [x] validate this is performance neutral on CPU with `-DImpactX_SIMD=ON`
- [x] validate this is performance neutral on CPU with `-DImpactX_SIMD=OFF`
- [x] validate this is performance neutral on GPU

[test_performance.sh](https://github.com/user-attachments/files/25037169/test_performance.sh)

### DIFF
--- a/src/elements/mixin/beamoptic.H
+++ b/src/elements/mixin/beamoptic.H
@@ -22,10 +22,124 @@
 #include <type_traits>
 
 
+namespace impactx
+{
+    /** A generalized ParallelFor that dispatches to amrex::ParallelFor or amrex::ParallelForSIMD
+     *  depending on whether the element type T_Element is vectorized.
+     *
+     * @tparam T_Element the element type
+     * @tparam F the functor type
+     * @param n the number of items to iterate over
+     * @param f the functor to execute
+     */
+    template <typename T_Element, typename F>
+    void ParallelFor (int n, F&& f) {
+#ifdef AMREX_USE_SIMD
+        if constexpr (amrex::simd::is_vectorized<T_Element>) {
+            amrex::ParallelForSIMD<T_Element::simd_width>(n, std::forward<F>(f));
+        } else
+#endif
+        {
+            amrex::ParallelFor(n, std::forward<F>(f));
+        }
+    }
+} // namespace impactx
+
 namespace impactx::elements::mixin
 {
 namespace detail
 {
+    /** Load particle data from array pointers
+     *
+     * On GPU and CPU w/o SIMD, this dereferences a particle property at the
+     * index position i.
+     * On CPU with SIMD, this loads a SIMD register at the IndexType::width
+     * SIMD-wide index position i.
+     *
+     * @tparam T data type (amrex::ParticleReal or uint64_t)
+     * @tparam IndexType int or amrex::SIMDindex<SIMD_WIDTH, int>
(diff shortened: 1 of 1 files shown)


src/elements/mixin/beamoptic.H
```diff
@@ -93,100 +184,50 @@ namespace detail
...
-            amrex::ParticleReal & AMREX_RESTRICT y = m_part_y[i];
-            amrex::ParticleReal & AMREX_RESTRICT t = m_part_t[i];
-            amrex::ParticleReal & AMREX_RESTRICT px = m_part_px[i];
-            amrex::ParticleReal & AMREX_RESTRICT py = m_part_py[i];
-            amrex::ParticleReal & AMREX_RESTRICT pt = m_part_pt[i];
-            amrex::ParticleReal & AMREX_RESTRICT sx = m_part_sx[i];
-            amrex::ParticleReal & AMREX_RESTRICT sy = m_part_sy[i];
-            amrex::ParticleReal & AMREX_RESTRICT sz = m_part_sz[i];
-            uint64_t & AMREX_RESTRICT idcpu = m_part_idcpu[i];
+            auto x = load_pdata(m_part_x, i);
```
AlexanderSinn:
```suggestion
            decltype(auto) x = load_pdata(m_part_x, i);
```
You need `decltype(auto)` here if you want it to be a reference for non-simd as before. However, I think it would also be interesting for GPU performance to test with `auto` and reuse the same writeback logic even for non-SIMD.

ax3l: Good catch! Thanks!

src/elements/mixin/beamoptic.H
```diff
@@ -93,100 +207,50 @@ namespace detail
...
-            if constexpr (push_nth_arg_non_const(6))
-                part_sx.copy_to(&m_part_sx[i], stdx::element_aligned);
-            if constexpr (push_nth_arg_non_const(7))
-                part_sy.copy_to(&m_part_sy[i], stdx::element_aligned);
-            if constexpr (push_nth_arg_non_const(8))
-                part_sz.copy_to(&m_part_sz[i], stdx::element_aligned);
-            if constexpr (push_nth_arg_non_const(9))
-                part_idcpu.copy_to(&m_part_idcpu[i], stdx::element_aligned);
-        }
+            if constexpr (amrex::simd::is_vectorized<T_Element>)
```
ax3l, line 235: I still want to get rid of this `if constexpr` to simplify further...

ax3l: `constexpr decltype(auto) P_Method = ..` does the trick for named methods like `spin_and_phasespace_push`, and for `operator()` we have to additonally `static_cast(function_signature)`.

PR #1280, opened 2026-01-29
Require R>0 for nonlinear DipEdge input

cemitch99:
The DipEdge element requires a strict R>0 when the nonlinear model is chosen.

### DIFF
--- a/src/elements/DipEdge.H
+++ b/src/elements/DipEdge.H
@@ -131,7 +131,7 @@ namespace impactx::elements
             if (m_R == 0_prt) {
                 warning_message.append(
                  "In DipEdge user has set R = 0.  This must be a positive value.  Setting R=1. "
-                );  
+                );
                ablastr::warn_manager::WMRecordWarning(
                 "elements::DipEdge",
                    warning_message,
--- a/src/elements/DipEdge.H
+++ b/src/elements/DipEdge.H
@@ -27,6 +27,7 @@
 #include <cmath>
 #include <stdexcept>
 
+
 namespace impactx::elements
 {
     namespace dipedge
--- a/src/elements/DipEdge.H
+++ b/src/elements/DipEdge.H
@@ -26,7 +26,7 @@
 
 #include <cmath>
 #include <stdexcept>
-
+#include <ablastr/warn_manager/WarnManager.H>
 
 namespace impactx::elements
 {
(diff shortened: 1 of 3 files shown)


src/elements/DipEdge.H
```diff
@@ -126,6 +126,20 @@ namespace impactx::elements
 
             Alignment::compute_constants(refpart);
 
+            // verify that the length scale constant R > 0
+            std::string warning_message;
+            if (m_R == 0_prt) {
+                warning_message.append(
+                 "In DipEdge user has set R = 0.  This must be a positive value.  Setting R=1. "
+                );
```
ax3l: We can handle this here, but it is a bit of a nifty location for input validation.

- Maybe rather in `elements.cpp` of Python and in `src/initialization/InitElement.cpp`?
- Do you want to check `R>0` to also capture negative values?
- I think throwing a runtime error is appropriate here.

cemitch99: Yes, throwing an error on `R < 0` is is better.  Will look at moving this check.

src/python/elements.cpp
```diff
@@ -862,6 +862,10 @@ void init_elements(py::module& m)
             std::optional<std::string> name
             )
             {
+                 if (R < 0.0)
+                     throw std::runtime_error(R"(DipEdge parameter R must be > 0.)");
+                     R = 1
```
cemitch99:
```suggestion
                     R = 1;
```

src/python/elements.cpp
```diff
@@ -862,6 +862,9 @@ void init_elements(py::module& m)
             std::optional<std::string> name
             )
             {
+                 if (R < 0.0)
```
```suggestion
                 if (R <= 0.0)
```

src/initialization/InitElement.cpp
```diff
@@ -215,6 +215,15 @@ namespace detail
...
             dipedge::Location const location = amrex::getEnum<dipedge::Location>(location_str);
 
+            if (R < 0) {
+                ablastr::warn_manager::WMRecordWarning(
+                    "ImpactX::read_element",
+                    element_name + " User passed R < 0 to DipEdge.  Please set R > 0. ",
+                    ablastr::warn_manager::WarnPriority::high
+                    );
+                    R = 1;
+            }
```
```suggestion
            if (R <= 0) {
                throw std::runtime_error(element_name + ".R must be >0 but is: " + std::to_string(R));
            }
```

src/python/elements.cpp
ax3l: There is one more location below in `.def_property("R", ...` that needs the same check for the setter part of it.

ax3l (approved):

cemitch99:
Waiting for CI to run.  All CI appears to be on hold at the moment.

PR #1281, opened 2026-01-29
Fix `rotate_spin`

ax3l:
Follow-up to #1262, only a bug with SIMD acceleration.

### DIFF
--- a/src/elements/mixin/spintransport.H
+++ b/src/elements/mixin/spintransport.H
@@ -69,9 +69,9 @@ namespace impactx::elements::mixin
             T_Real w1 = 0_prt;
             amrex::simd::stdx::where(non_zero_angle, w1) = lambdax * c_sin_half;
             T_Real w2 = 0_prt;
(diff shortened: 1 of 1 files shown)


cemitch99 (approved):
Good catch.  I'm surprised this didn't affect the SIMD test in the last spin PR we merged...

ax3l:
I was surprised as well, then I saw:
Our quad spin test is [declared as MPI-only](https://github.com/BLAST-ImpactX/impactx/blob/469fddd33dd18214706c20f6ea6ef3e05b4986b2/examples/CMakeLists.txt#L1872-L1878)... And our only [CI test runner w/ SIMD so far uses MPI=OFF](https://github.com/BLAST-ImpactX/impactx/blob/469fddd33dd18214706c20f6ea6ef3e05b4986b2/.github/workflows/ubuntu.yml#L151-L153) :laughing:

ax3l:
Solution(s):
- either the Python or Inputs test could run as no-MPI #1282
- we should turn on SIMD for at least one MPI runner, too #1283

PR #1290, opened 2026-02-03
Add spin FODO example

cemitch99:
This PR adds an example that illustrates spin depolarization for an otherwise matched beam, occurring in 100 periods of a FODO channel.  It illustrates the use of spin in the following elements:  `ChrQuad`, `ChrDrift`.  As a result, the evolution of the polarization is sensitive to the beam energy spread.


cemitch99:
I messed up the commit history here.  This is a simple PR - let me close this and open a clean one.

PR #1296, opened 2026-02-06
Pip Install: Skip `diags/`

### LINKED ISSUE #1294: Pip Install: Skip `diags/`
When running `--target pip_install` we run `pip wheel` on the source tree.

If one has run in-source for example testing, this bloats up the wheel (zip) file, slows the process, and up to the ZIP limit (>4GB) where it fails to build 💥 

We should categorically exclude `diags` and `diags.*` from `pip wheel`.

ax3l:
Fix #1294

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -15,3 +15,9 @@ recursive-include tests *
 # see .gitignore
 prune cmake-build*
 prune .spack-env*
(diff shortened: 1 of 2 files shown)


cemitch99 (approved):
Wonderful.

PR #1297, opened 2026-02-06
Add user-defined Spin Map

cemitch99:
This PR adds an element that applies a user-defined spin map to the spin variables `(sx, sy, sz)`--a functionality that is similar to that of the user-defined LinearMap element.  The LinearMap element is also updated for spin support--it leaves spin unaffected.

- [x] implement element inputs and input parsing
- [x] implement math for the spin push
- [x] implement Python bindings (esp. for user-facing input) https://github.com/AMReX-Codes/pyamrex/pull/538 #1333
- [x] add C++ benchmark example
- [x] add Python benchmark example
- [x] add analysis script
- [x] update documentation

Follow-up:  Investigate the loss of precision (1e-12 -> 1e-9) that occurs when Python input is used.
See Issue #1342

### DIFF
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -873,7 +873,7 @@ The vector components :math:`v(i)` and the matrix elements :math:`A(i,j)` are in
 The vector :math:`v` and the matrix :math:`A` are defaulted to zero, so only entries that differ from zero need to be specified.
 
 The matrix :math:`A` multiplies the phase space vector :math:`(x,p_x,y,p_y,t,p_t)`, where coordinates :math:`(x,y,t)` have units of m
-and momenta :math:`(px,py,pt)` are dimensionless.  The three components output are dimensionless.  So, for example, :math:`A(1,1)` has units of 1/m, and :math:`A(1,2)` is dimensionless.
+and momenta :math:`(p_x,p_y,p_t)` are dimensionless.  The three components output are dimensionless.  So, for example, :math:`A(1,1)` has units of 1/m, and :math:`A(1,2)` is dimensionless.
 All three components of :math:`v` are dimensionless.
 
 This element requires these additional parameters:
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -866,7 +866,7 @@ Spin maps are specified in the Lie-algebraic form:
 
    \vec{s}_f = M(\zeta)\vec{s}_i where M(\zeta)=e^{v\cdot L}e^{A\Delta\zeta\cdot L}.
 
-Here :math:`v` is a 3-vector that defines the axis and angle of rotation at the phase space design point, and :math:`A` is a 3x6 matrix that defines the spin-orbit coupling for particles not on the design orbit. 
+Here :math:`v` is a 3-vector that defines the axis and angle of rotation at the phase space design point, and :math:`A` is a 3x6 matrix that defines the spin-orbit coupling for particles not on the design orbit.
 Also, :math:`\Delta\zeta=(x,p_x,y,p_y,t,p_t)` denotes the 6-vector of phase space variables as deviations from the design orbit. The quantities :math:`L_x`, :math:`L_y`, and :math:`L_z` are standard 3x3 matrices that define a basis for the Lie algebra :math:`so(3)`.
 
 The vector components :math:`v(i)` and the matrix elements :math:`A(i,j)` are indexed beginning with 1, so that :math:`i=1,2,3` and :math:`j=1,2,3,4,5,6`.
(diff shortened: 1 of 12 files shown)


cemitch99:
The Python bindings for the user-facing inputs, which use SmallMatrix type, need support similar to what we used for the LinearMap element.

examples/spin_tracking/run_spin_map.py
```diff
@@ -0,0 +1,94 @@
...
+    0.1,
+)
+sim.add_particles(bunch_charge_C, distr, npart, spin)
+
+# add beam diagnostics
+monitor = elements.BeamMonitor("monitor", backend="h5")
+
+# initialize the spin map generators
+vmat = Vector3
+Amat = Map3x6
```
cemitch99:
```suggestion
```

ax3l:
Added a type for the 3x6 matrix and 3x1 vector in https://github.com/AMReX-Codes/pyamrex/pull/538 now.

docs/source/usage/parameters.rst
```diff
@@ -855,6 +855,39 @@ Currently, this only supports openPMD files from our ``beam_monitor``.
   Inject particles only for the first lattice period.
 
 
+``spin_map``
+^^^^^^^^^^^^^^
```
```suggestion
``spin_map``
^^^^^^^^^^^^
```

examples/spin_tracking/analysis_spin_map.py
```diff
@@ -0,0 +1,43 @@
...
+sxf = final["spin_x"]
+syf = final["spin_y"]
+szf = final["spin_z"]
+
+dspin2 = (sxf - sxi) ** 2 + (syf - syi) ** 2 + (szf - szi) ** 2
+dspin = np.sqrt(dspin2)
+dspinmax = dspin.max()
+
+print("Change in the spin:")
+print("||delta s||_max", dspinmax)
```
ax3l, line 32: I see in CI:
```
Change in the spin:
||delta s||_max 2.2705360483651895e-09
  atol=1.5e-12
```

cemitch99: Yes.  The version of the test based on the C++ input works within this tolerance.  However, the Python version does not (yet).  Maybe something is being done in single-precision...?

ax3l: The macOS and Tooling CI (Ubuntu) runs that show the larger tolerance are also in double precision.

ax3l: The difference is probably from parallel execution. The failing runs have MPI enabled. So maybe just not converged enough at the current resolution (no. of particles)...?

cemitch99: Hm.  The no. of particles should no be relevant in this example, because the change in the spin should be zero (to machine precision) for every particle.

examples/spin_tracking/run_spin_map.py
```diff
@@ -0,0 +1,98 @@
...
+AmatB1[1, 4] = 1.643140677773437
+AmatB1[2, 1] = 0.236555147919017
+AmatB1[2, 2] = 0.118376237268958
+AmatB1[2, 6] = 0.096052815713841
+AmatB1[3, 4] = -0.765638392807848
+Map2 = elements.SpinMap(v=vmatB1, A=AmatB1)
+
+# line = [monitor, Map1, Q1, Map2, B1, monitor]
+# line = [monitor, Map1, Q1, monitor]
+line = [monitor, Map2, B1, monitor]
```
ax3l: @cemitch99 I think the lattice in Python is incomplete (compared to the `input_spin_map.in`)

examples/spin_tracking/analysis_spin_map.py
ax3l: Makes sense. Also, bot inputs & py are run with MPI as I just checked, but the inputs pass.

I compared the lattices in both and the Python one forgets to comment in the full lattice.

examples/spin_tracking/run_spin_map.py
cemitch99: Oh.  You're right that this is not the full lattice--I used this segment for benchmarking.  However, this segment should also produce 0 spin change, so this is not the problem.

cemitch99:
```suggestion
line = [monitor, Map1, Q1, Map2, B1, monitor]
```

cemitch99: Fixing this now in any case.

examples/spin_tracking/analysis_spin_map.py
```diff
@@ -0,0 +1,43 @@
...
+
+atol = 1.5e-12
```
cemitch99:
```suggestion
atol = 5.0e-9
```

qianglbl (approved):

examples/spin_tracking/analysis_spin_map.py
ax3l: I saw you bumped the tolerance in the end.

ax3l (approved):

PR #1301, opened 2026-02-11
Doc: Perlmutter Python Batch Script

ax3l:
Make HPC batch templates portable to Python & executable usage alike.
Put Python first as in the rest of the docs.
Add fallback if a user forgot to source their profile.

- [x] wait for solution in https://github.com/AMReX-Codes/amrex/issues/4947

### DIFF
--- a/docs/source/install/hpc/lumi-csc/lumi_gpu.sbatch
+++ b/docs/source/install/hpc/lumi-csc/lumi_gpu.sbatch
@@ -17,6 +17,15 @@ INPUTS=run_fodo.py
 #EXE=./impactx
 #INPUTS=input_fodo.in
 
+# enviroment setup
+if [[ -z "${MY_PROFILE}" ]]; then
+    echo "WARNING: FORGOT TO"
+    echo "   source $HOME/lumi_gpu_impactx.profile"
+    echo "before submission. Doing that now."
+
+    source $HOME/lumi_gpu_impactx.profile
+fi
+
 date
 
 # note (12-12-22)
--- a/docs/source/install/hpc/lumi-csc/lumi_gpu.sbatch
+++ b/docs/source/install/hpc/lumi-csc/lumi_gpu.sbatch
@@ -10,12 +10,12 @@
 #SBATCH --ntasks-per-node=8
 #SBATCH --gpus-per-node=8
 
-# executable & inputs file or ...
-EXE=./impactx
-INPUTS=inputs
-# ... python interpreter & PICMI script here
-#EXE=python3
-#INPUTS=run_fodo.py
+# python interpreter & script here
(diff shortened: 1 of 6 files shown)


docs/source/install/hpc/perlmutter-nersc/perlmutter_gpu.sbatch
ax3l: @WeiqunZhang in Python scripts for batch systems, we often need to set `amrex.use_gpu_aware_mpi=1` externally.

Do you think we could define an optional environment variable in AMReX, say `AMREX_USE_GPU_AWARE_MPI` that at runtime changes the default of AMReX if set? The precedence would be as always:

1. CLI option
2. inputs file
3. environment option
4. hard-coded default

https://github.com/AMReX-Codes/amrex/issues/4947

docs/source/install/hpc/lumi-csc/lumi_gpu.sbatch
```diff
@@ -10,12 +10,12 @@
 #SBATCH --ntasks-per-node=8
 #SBATCH --gpus-per-node=8
```
cemitch99, line 11: I suspect we need to activate the appropriate modules here by adding:

```suggestion
#SBATCH --gpus-per-node=8

source ~/perlmutter_gpu_impactx.profile
```

or similar syntax.

docs/source/install/hpc/perlmutter-nersc/perlmutter_gpu.sbatch
ax3l: yay, we can use: https://amrex-codes.github.io/amrex/docs_html/Basics.html?highlight=amrex_default_init#setting-default-via-environment-variable

i.e.
```bash
export AMREX_DEFAULT_INIT="amrex.use_gpu_aware_mpi=1"
```

docs/source/install/hpc/lumi-csc/lumi_gpu.sbatch
ax3l: Usually not, because we source them in our workflow upfront and want to inherit it at submission time.... (i.e. log into a machine, pick a profile for a workflow, compile or submit).

But I can build in a fallback if users forget.

ax3l: done

cemitch99 (commented):
I would like to test this myself, but Perlmutter has been down since yesterday.  Hope to test and approve once services is restored.

cemitch99 (approved):
Testing this, and it works for me.  A good improvement to the documentation.

ax3l:
Thank you for catching this, that was definitely a missed workflow!

PR #1303, opened 2026-02-14
Fix Polygon Aperture Finalize

ax3l:
We need to deallocate cleanly the host and device data of the polygon aperture data when elements are finalized.

This resolves segfaults at the end of the simulation on GPU.


src/elements/PolygonAperture.H
ax3l:
```suggestion
```

PR #1305, opened 2026-02-15
Fix: Python Wake Tests on GPU

ax3l:
Ensure the CSR wake Python tests work also on GPU:
- use proper vector type arenas (device)
- carefully copy to host when needed
- carefully access with cupy when manipulated from Python on the device side
- [x] depends on https://github.com/AMReX-Codes/pyamrex/pull/540
- [x] test: works on CUDA now

### DIFF
--- a/tests/python/test_wake.py
+++ b/tests/python/test_wake.py
@@ -48,8 +48,8 @@ def test_wake(save_png=True):
         bin_size = (bin_max - bin_min) / (num_bins - 1)
 
         # Create charge_distribution and slopes as PODVector_real_std
-        charge_distribution = amr.PODVector_real_std(num_bins + 1)
-        slopes = amr.PODVector_real_std(num_bins)
+        charge_distribution = amr.PODVector_real_arena(num_bins + 1)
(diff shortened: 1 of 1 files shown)


tests/python/test_wake.py
```diff
@@ -48,8 +48,8 @@ def test_wake(save_png=True):
         bin_size = (bin_max - bin_min) / (num_bins - 1)
 
         # Create charge_distribution and slopes as PODVector_real_std
-        charge_distribution = amr.PODVector_real_std(num_bins + 1)
-        slopes = amr.PODVector_real_std(num_bins)
+        charge_distribution = amr.PODVector_real_arena(num_bins + 1)
+        slopes = amr.PODVector_real_arena(num_bins)
```
ax3l: Oh, it's a bit more complicated to make the same interface for CPU compiles. Might need to update the python bindings, too.

ax3l: Needs: https://github.com/AMReX-Codes/pyamrex/pull/540

cemitch99 (approved):
Great!

PR #1307, opened 2026-02-15
[WIP] Element Finalize Logic on Destructor

ax3l:
We nove have a few interactive workflows, e.g.,
in `test_benchmark_elements.py`, where we push elements that hold dynamic device data without ever placing them in `sim.lattice`. This revamps the element `finalize` logic to be called automatically on element destruction. The latter is implied in `sim.lattice.clear()` .

### DIFF
--- a/src/elements/mixin/beamoptic.H
+++ b/src/elements/mixin/beamoptic.H
@@ -434,6 +434,20 @@ namespace detail
     template<typename T_Element>
     struct BeamOptic
     {
+        /** Close and deallocate all data and handles.
+         *
+         * Some elements manager dynamic device memory for input data.
+         * This ensures their finalize() method is called when they
+         * get destroyed, so device memory is deallocated properly.
+         */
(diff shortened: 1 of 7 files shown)


src/elements/mixin/beamoptic.H
```diff
@@ -434,6 +434,20 @@ namespace detail
...
+         *
+         * Some elements manager dynamic device memory for input data.
+         * This ensures their finalize() method is called when they
+         * get destroyed, so device memory is deallocated properly.
+         */
+        AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+        ~BeamOptic () {
+            AMREX_IF_ON_HOST((
+                T_Element& element = *static_cast<T_Element*>(this);
+                element.finalize();
```
ax3l, line 447: Hm, maybe this approach does not work because we need to make a copy before launch....? To debug

PR #1312, opened 2026-02-20
Add Cut to Gaussian distribution type

### LINKED ISSUE #1253: Add cutoff support for initial Gaussian beam distribution
For Gaussian beam distribution types, it will be helpful to support truncated distributions with a cutoff of the form N*sigma for some N.  In some cases, it can be useful to adjust the cutoff value in each phase space variable independently.

cemitch99:
~This PR adds a new initial particle distribution type, `CutGaussian`, that is similar to a 6D Gaussian, but it~ *Expand the `Gaussian` distribution, which now* allows the user to specify an independent cutoff in each phase plane `(x,px)`, `(y,py)`, and `(t,pt)`.  The cut is performed in normalized Courant-Snyder variables corresponding to the user-supplied second moments or Twiss functions.  As a result, this is equivalent to a cut corresponding to the (linearized) action in each plane.

- [x] implement distribution type and sampling algorithm
- [x] add initialization of parameters from user input
- [x] update Python bindings
- [x] add benchmark example
- [x] update documentation
- [x] decide on default parameters & position in argument list
- [x] deduplicate code with `Gaussian`?

If the user would like to apply a circular or elliptical cut in the `(x,y)` plane, this is more appropriately done using the `Aperture` element, which represents the effect of a transverse mask.

The PR closes #1253


src/particles/distribution/CutGaussian.H
```diff
@@ -0,0 +1,198 @@
...
+            // Generate three pairs of truncated normal random variables using a modified Box-Muller transformation:
+
+            u1 = amrex::Random(engine);
+            u2 = amrex::Random(engine);
+            // factors determined by the first cutoff value:
+            f = m_cutX*m_cutX/2.0_prt;
+            ef = std::exp(-f);
+            r2scale = 1_prt - f*ef/(1_prt - ef);
+            r2 = -2_prt * std::log(1_prt - u1*(1_prt - ef));
+            ln1 = std::sqrt(r2 / r2scale);
```
cemitch99: These lines (and the corresponding lines in the code blocks below) define the effect of the cutoff.  As cutX -> infinity, this reduces to the Box-Mueller transformation defining the parameter `ln1` that appears in `Gaussian.H`.

docs/source/usage/python.rst
```diff
@@ -614,6 +614,35 @@ For the input from Twiss parameters in Python, please use the helper function ``
...
 
+.. py:class:: impactx.distribution.CutGaussian(cutX=5.0, cutY=5.0, cutT=5.0, lambdaX, lambdaY, lambdaT, lambdaPx, lambdaPy, lambdaPt, muxpx=0.0, muypy=0.0, mutpt=0.0, meanX=0.0, meanY=0.0, meanT=0.0, meanPx=0.0, meanPy=0.0, meanPt=0.0, dispX=0.0, dispPx=0.0, dispY=0.0, dispPy=0.0)
+
+   A 6D Gaussian distribution with truncation.  The user may specify an independent cutoff in each phase plane (x,px), (y,py), and (t,pt). The cut is performed in normalized
+   Courant-Snyder variables corresponding to the user-supplied second moments or Twiss functions. As a result, this is equivalent to a cut corresponding to the (linearized) action in each
+   plane.
+
+   :param cutX: number of sigma at which to cut the distribution in (x,px) (dimensionless)
+   :param cutY: number of sigma at which to cut the distribution in (y,py) (dimensionless)
+   :param cutT: number of sigma at which to cut the distribution in (t,pt) (dimensionless)
```
ax3l: Do you want these intentionally as first parameters? If you put them to the end, or at least after `lambda*` then you can default them to a value.

Also, is 5.0 a good default or is "no cut" a good default?

src/initialization/InitDistribution.cpp
```diff
@@ -197,6 +197,21 @@ namespace impactx
                         meanx, meany, meant,
                         meanpx, meanpy, meanpt,
                         dispx, disppx, dispy, disppy);
+            } else if (base_dist_type == "cutgaussian") {
+                amrex::ParticleReal cutx = 5.0;
+                amrex::ParticleReal cuty = 5.0;
+                amrex::ParticleReal cutt = 5.0;
```
ax3l: Please see above on defaults.

src/particles/distribution/CutGaussian.H
```diff
@@ -0,0 +1,198 @@
...
+         * @param cutx, cuty, cutt dimensionless cut parameter (number of sigma) in each phase plane
+         * @param lambdax,lambday,lambdat phase space axis intercepts (position);
+         *                                for zero correlation, these are the related RMS sizes (in meters)
+         * @param lambdapx,lambdapy,lambdapt phase space axis intercepts (momentum)
+         * @param muxpx,muypy,mutpt correlation length-momentum
+         * @param meanx,meany,meant offsets of the mean (centroid) positions from those of the reference particle
+         * @param meanpx,meanpy,meanpt offsets of the mean (centroid) momenta from those of the reference particle
+         * @param dispx,disppx,dispy,disppy dispersion and its derivative in horizontal and vertical directions
+         */
+        CutGaussian (
```
ax3l: Not required and does not help a ton, but technically we could probably even derive from the `Gaussian` class... (do not change for now...)

src/particles/distribution/CutGaussian.H
ax3l: Thanks! If the code change is very localized, one could probably deduplicate the code by deriving from `Gaussian` and defining an inlined helper function. But I can take a look at this if desired.

src/python/distribution.cpp
```diff
@@ -187,6 +187,28 @@ void init_distribution(py::module& m)
...
+                amrex::ParticleReal, amrex::ParticleReal, amrex::ParticleReal,
+                amrex::ParticleReal, amrex::ParticleReal, amrex::ParticleReal,
+                amrex::ParticleReal, amrex::ParticleReal, amrex::ParticleReal,
+                amrex::ParticleReal, amrex::ParticleReal, amrex::ParticleReal,
+                amrex::ParticleReal, amrex::ParticleReal, amrex::ParticleReal,
+                amrex::ParticleReal, amrex::ParticleReal, amrex::ParticleReal,
+                amrex::ParticleReal, amrex::ParticleReal,
+                amrex::ParticleReal, amrex::ParticleReal
+             >(),
+             py::arg("cutX"), py::arg("cutY"), py::arg("cutT"),
```
ax3l: See comment above in `python.rst`:

The defaults are documented for the Python API but they do not work if the parameters are defined before the parameters that have no defaults (`lambda*`).

src/particles/distribution/CutGaussian.H
cemitch99: Thanks, but I don't see much advantage in this.  Most of the remaining logic in this function is common to all the core distribution types.  There's not much that is `Gaussian` -specific that would be re-used here.

cemitch99: The other option could be to merge this with `Gaussian` and include an internal conditional based on whether or not cutoff values are provided.  However, if we do this at some point, I think it would be better to do this closer to the user input level (not in the sampling routine).

docs/source/usage/python.rst
cemitch99: "No cut" is currently not defined.  This would correspond to a value of infinity, and in this limit the distribution becomes equivalent to a `Gaussian`.  In practice, the tails decay quickly beyond 5 or 6 sigma, so I think this is a reasonable default.

src/particles/distribution/CutGaussian.H
ax3l: Yes, on the way back today I had the same thought. The additional overhead from the extra parameters & cutting is probably negligible compared to the random number generation in the distribution sampling, thus it would likely not hurt if we just added this as new parameters to the Gaussian distribution.

docs/source/usage/python.rst
ax3l: I unified the logic now and defined 0 as no cut (because dealing with Inf/NaN is not ideal, due to their special meaning and comparison behavior).

ax3l (approved):
Unified the logic now in a separate commit into the existing Gaussian distribution - does this look good to you?

cemitch99:
Looks good to me.  Re-checked the functionality and the documentation.

PR #1326, opened 2026-02-25
Determinism in `initialize_from_array.py`

### LINKED ISSUE #1325: Determinism in `initialize_from_array.py`
The `initialize_from_array.py` test is not deterministic enough (probably from drawing random numbers) and sometimes fails the thresholds in `initialize_from_array.py.analysis`.

ax3l:
Fix #1325: all random number draws need to be deterministic with a fixed seed to avoid fluctuations in CI. This fixes the transverse components, before only z-pz were deterministic.

### DIFF
--- a/examples/initialize_from_array/run_from_array.py
+++ b/examples/initialize_from_array/run_from_array.py
@@ -19,10 +19,12 @@ beam_radius = 2e-3
 sigr = 500e-6
 sigpx = 10
 sigpy = 10
-px = np.random.normal(0, sigpx, N_part)
-py = np.random.normal(0, sigpy, N_part)
-theta = 2 * np.pi * np.random.rand(N_part)
-r = abs(np.random.normal(beam_radius, sigr, N_part))
+# fixed seed for deterministic test results
+rng = np.random.default_rng()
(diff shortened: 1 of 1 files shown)


examples/initialize_from_array/run_from_array.py
```diff
@@ -19,10 +19,12 @@
 sigr = 500e-6
 sigpx = 10
 sigpy = 10
-px = np.random.normal(0, sigpx, N_part)
-py = np.random.normal(0, sigpy, N_part)
-theta = 2 * np.pi * np.random.rand(N_part)
-r = abs(np.random.normal(beam_radius, sigr, N_part))
+# fixed seed for deterministic test results
+rng = np.random.default_rng()
```
ax3l, line 23: Dang, forgot the actual seed here:

```suggestion
rng = np.random.default_rng(seed=42)
```
https://numpy.org/doc/stable/reference/random/generator.html
> If None, then fresh, unpredictable entropy will be pulled from the OS.

ax3l: Fix in #1536

PR #1327, opened 2026-02-26
Examples CMake: Cleaning

ax3l:
Purely unifying style:
- align title blocks to 80 chars
- ensure followed by a single `#` line
- remove spurious hint of with or without space charge: we used this initially but it is not relevant at this point in the test logic


cemitch99 (approved):
Looks good; thanks.

PR #1329, opened 2026-02-27
Docs: Fix `transfer_map`, Link Example

ax3l:
There was a typo in the docs to calculate the transfer map of the elements in the list.

Also links the test as an example how to use it.

Preview: https://impactx--1329.org.readthedocs.build/en/1329/usage/python.html#impactx.elements.KnownElementsList.transfer_map


cemitch99 (approved):
Thanks.

Issue #1343, opened 2026-03-08
ExactQuad has unexpected longitudinal coordinate change


nikitakuklev:
I am either going crazy or this is an ExactQuad map bug. Tracking reference particle creates nonzero coordinates, unlike with Quad (and other elements).

```
import numpy as np
import amrex.space3d as amr

from impactx import Config, ImpactX, elements


KIN_ENERGY_MEV = 400.0
MASS_MEV = 938.27208816
QM_EEV = 1.0 / (MASS_MEV * 1.0e6)


def make_vec(values):
    vec_cls = amr.PODVector_real_arena if Config.have_gpu else amr.PODVector_real_std
    out = vec_cls()
    for vv in values:
        out.push_back(float(vv))
    return out


def run_one(element):
    sim = ImpactX()
    sim.space_charge = False
    sim.diagnostics = False
    sim.slice_step_diagnostics = False
    sim.verbose = 0
    sim.init_grids()

    pc = sim.particle_container()
    ref = pc.ref_particle()
    ref.set_charge_qe(1.0).set_mass_MeV(MASS_MEV).set_kin_energy_MeV(KIN_ENERGY_MEV)

    # One particle exactly on the reference trajectory
    pc.add_n_particles(
        make_vec([0.0]),  # x
        make_vec([0.0]),  # y
        make_vec([0.0]),  # t
        make_vec([0.0]),  # px
        make_vec([0.0]),  # py
        make_vec([0.0]),  # pt
        QM_EEV,
        bunch_charge=0.0,
    )

    sim.lattice.append(element)
    sim.track_particles()

    df = pc.to_df(local=True)
    print(df[[
        "position_x",
        "momentum_x",
        "position_y",
        "momentum_y",
        "position_t",
        "momentum_t",
    ]])

    sim.finalize()


print("Quad")
run_one(elements.Quad(name="q", ds=0.5, k=1.2))

print("\nExactQuad")
run_one(elements.ExactQuad(name="eq",ds=0.5,k=1.2))

print("\nExactMultipole")
run_one(elements.ExactMultipole(name="em",ds=0.5,k_normal=[0,1.2],k_skew=[0.0,0.0]))

# also softquad and other stuff gives ~0
```

```
Quad
       position_x  momentum_x  position_y  momentum_y  position_t  momentum_t
idcpu                                                                        
0             0.0         0.0         0.0         0.0         0.0         0.0

ExactQuad
[...]

PR #1345, opened 2026-03-09
Fix t-update in ExactQuad.

cemitch99:
This PR fixes a bug in the longitudinal _t_-update in the element ExactQuad, resulting from a missing factor of relativistic beta.  The transverse update is not affected.  The effect becomes large when $\beta<<1$. 

- [x] add bug fix
- [x] add benchmark example

This closes Issue #1343.  Thanks to @nikitakuklev.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1970,7 +1970,7 @@ add_impactx_test(spin-map.py
 # Symplectic integration in an ExactQuad on Reference ###################
 add_impactx_test(exact-quad-on-reference.py
     examples/symplectic_integration/run_exact_quad_on_reference.py
-    ON   # ImpactX MPI-parallel
+    OFF   # ImpactX MPI-parallel
     examples/symplectic_integration/analysis_exact_quad_on_reference.py
     OFF  # no additional analysis script needed
 )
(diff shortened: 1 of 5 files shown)


examples/CMakeLists.txt
```diff
@@ -1966,3 +1966,11 @@ add_impactx_test(spin-map.py
     examples/spin_tracking/analysis_spin_map.py
     OFF  # no plot script yet
 )
+
+# Symplectic integration in an ExactQuad on Reference ###################
+add_impactx_test(exact-quad-on-reference.py
+    examples/symplectic_integration/run_exact_quad_on_reference.py
+    ON   # ImpactX MPI-parallel
```
```suggestion
    OFF   # ImpactX MPI-parallel
```

cemitch99: Running this test with MPI poses problems, and is unnecessary.

egstern (approved):
Glad you found it! We should have more tests with low betas!

qianglbl (approved):

examples/symplectic_integration/run_exact_quad_on_reference.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022-2023 ImpactX contributors
+# Authors: Axel Huebl, Chad Mitchell
```
cemitch99:
```suggestion
# Authors: Nikita Kuklev, Axel Huebl, Chad Mitchell
```

PR #1347, opened 2026-03-10
Update naming of on-axis data input options for soft-edge elements

cemitch99:
This PR clarifies the meaning of on-axis data provided for soft-edge elements by replacing the string `field_or_gradient` with either `field_on_axis` or `gradient_on_axis` as appropriate.

- [x] update extension functions
- [x] update examples using these functions
- [x] update documentation

This is a follow-up to PR #1316.


ax3l (approved):
Perfect, thank you for improving this!

PR #1323, opened 2026-02-25
Add spin to ExactQuad.

cemitch99:
This PR adds spin tracking to the `ExactQuad` element, which requires combining the Thomas-BMT equation s-push with the orbit symplectic integration, to the correct order in stepsize _h_.  The basic framework will be reused for the `ExactMultipole` and `ExactCFbend` elements.

- [x] add basic spin structure for ExactQuad
- [x] add Thomas-BMT precession vector evaluation
- [x] add 2nd-order symplectic integration in the presence of spin
- [x] add 4th-6th-order symplectic integration in the presence of spin
- [x] move Thomas-BMT precession vector to `mixin/spin_transport` class
- [x] add a benchmark (e.g. comparison against a linear Quad as a limiting case)
- [x] fix failing CI tests (all works locally)
- [x] fix inline comment on bunch charge in particle container

Follow-up:  Investigate alternative 4th and 6th-order symplectic integration algorithms with spin to improve performance (by reducing the number of required map evaluations).  See Issue #1341

PR #1351, opened 2026-03-12
Elements: Compile Push in Separate Translation Units

ax3l:
Each element's `push_all<T>` instantiation is now compiled in its own translation unit (`src/elements/<Element>.cpp`), enabling parallel compilation of the 33 GPU-heavy kernel instantiations that were previously all compiled together through `Push.cpp`. Two convenience macros in `PushAll.H`:
* `IMPACTX_PUSH_EXTERN_TEMPLATE` and
* `IMPACTX_PUSH_INSTANTIATE`

keep the boilerplate minimal and centralize the `push_all` signature.                                                                                                                                          
                                                            
Each element header declares `extern template` to suppress implicit instantiation, while the corresponding `.cpp` file provides the explicit instantiation definition. `Push.cpp` is unchanged: it still includes `All.H` and uses `std::visit`, but now only emits function calls to the externally-defined symbols instead of instantiating the full template chain (`push_all` → `push_all_particles` → `PushSingleParticle` → `ParallelFor` + element `operator()`). GPU performance is preserved because each per-element TU is compiled with full visibility of the template chain, so the per-particle physics remains fully inlined into `ParallelFor`.

Elements without `BeamOptic` (`Empty`, `Marker`, `Programmable`, `Source`, `BeamMonitor`) are unaffected.


## To Do?

- [x] CPU/OpenMP: measure per TU and complete compile time before/with this PR (ccache turned off)
- [x] GPU/CUDA: measure per TU and complete compile time before/with this PR (ccache turned off)

### DIFF
--- a/src/elements/Aperture.H
+++ b/src/elements/Aperture.H
@@ -247,4 +247,6 @@ namespace impactx::elements
 
 } // namespace impactx
 
+IMPACTX_PUSH_EXTERN_TEMPLATE(impactx::elements::Aperture)
+
 #endif // IMPACTX_APERTURE_H
--- a/src/elements/Aperture.cpp
+++ b/src/elements/Aperture.cpp
@@ -40,3 +40,5 @@ impactx::elements::Aperture::action_name (Action const & action)
             throw std::runtime_error("Unknown action");
     }
 }
+
+IMPACTX_PUSH_INSTANTIATE(impactx::elements::Aperture)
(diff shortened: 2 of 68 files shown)


ax3l:
Hm, I see a big improvement in `Push.cpp` time, also some in `python/elements.cpp` (but this one needs to be split for elements because it has a lot of pybind11 logic).

[ninja_dev_trace.json](https://github.com/user-attachments/files/26173967/ninja_dev_trace.json)
[ninja_multitu_trace.json](https://github.com/user-attachments/files/26173970/ninja_multitu_trace.json)

Overall time went slightly up, but that is more from build noise in pyAMReX (unchanged).

ax3l:
Merging this to help with #1323

PR #1352, opened 2026-03-12
Return linear map for simple nonlinear elements.

cemitch99:
This PR adds the return of the linear map for several core elements that use nonlinear element models.  Here, this includes:

- [x] ChrQuad
- [x] ChrDrift
- [x] ChrPlasmaLens
- [x] ExactDrift
- [x] ExactSbend
- [ ] simplify: call maps of linear elements -- later because of diverged inputs

The linear maps here are identical to those in the corresponding linear element (Quad, Drift, etc.)

Note that other nonlinear elements, including `ExactQuad`, `ExactMultipole` and `ExactCFbend`, are already supported.

PR #1362, opened 2026-03-24
Add support for SI unit input to ChrQuad linear map return.

cemitch99:
The return of the linear map used in the element ChrQuad works correctly for `unit = 0` (normalized gradient).  This PR extends the support to include the case when `unit = 1` (gradient in SI units).

This is a follow-up to PR #1352


ax3l (approved):
Thank you! ✨

PR #1364, opened 2026-03-26
CMake: pyAMReX Compile Selected PCs

ax3l:
Safe compile time when building pyAMReX in superbuilds.

- [x] See https://github.com/AMReX-Codes/pyamrex/pull/550

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -33,6 +33,11 @@ jobs:
 
     - uses: actions/checkout@v6
 
+    - name: Setup Python
+      uses: actions/setup-python@v6
+      with:
+        python-version: "3.11"
+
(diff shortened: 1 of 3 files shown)


EZoni (approved):

EZoni:
I see the CI check `OpenMP / GCC w/ MPI w/ Python` failed, but maybe it's just a glitch that gets fixed by restarting the workflow?

EZoni:
> I see the CI check `OpenMP / GCC w/ MPI w/ Python` failed, but maybe it's just a glitch that gets fixed by restarting the workflow?

Actually, I did find a legitimate error log, though I'm a bit confused:
```
ERROR: Package 'amrex' requires a different Python: 3.10.12 not in '>=3.11'
```

PR #1365, opened 2026-03-26
Add to_dict return of int_order to Exact elements.

cemitch99:
Beamline elements that use symplectic integration do not currently return the value of the `int_order` parameter as part of the `to_dict()` call.  This PR addresses this.

This closes #1332 .

Issue #1332, opened 2026-03-02
int_order parameter is not exported by to_dict() method


egstern:
For the elements ExactMultipole and ExactCFbend, the `to_dict()` method does not give the value of the `int_order` parameter.

```
>>> import impactx

>>> m_elem = impactx.elements.ExactMultipole(ds=1.0, k_normal=[0.0, 0.01, 0.002\
], k_skew=[0.0, 0.002, 0.0004], int_order=6, mapsteps=6)

>>> dir(m_elem)
['__class__', '__delattr__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getstate__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '_pybind11_conduit_v1_', 'aperture_x', 'aperture_y', 'ds', 'dx', 'dy', 'finalize', 'has_name', 'int_order', 'mapsteps', 'name', 'nslice', 'push', 'rotation', 'to_dict', 'unit']
```

The `int_order` parameter is present in the element structure.

```
>>> m_elem.to_dict()
{'aperture_x': 0.0, 'aperture_y': 0.0, 'ds': 1.0, 'dx': 0.0, 'dy': 0.0, 'k_normal': [0.0, 0.01, 0.002], 'k_skew': [0.0, 0.002, 0.0004], 'mapsteps': 6, 'name': None, 'nslice': 1, 'rotation': 0.0, 'type': 'ExactMultipole', 'unit': 0}
```

but the `int_order ` value is not in the dict return by the `to_dict()` method.

The same thing happens with the ExactCFbend element.

egstern:
Same issue with ExactQuad as well.

```
>>> import impactx
>>> elem = impactx.elements.ExactQuad(ds=1.0, k=.01, int_order=6, mapsteps=6)
>>> elem
<impactx.elements.ExactQuad, ds=1.000000, k=0.010000, unit=0>
>>> elem.to_dict()
{'aperture_x': 0.0, 'aperture_y': 0.0, 'ds': 1.0, 'dx': 0.0, 'dy': 0.0, 'k': 0.01, 'name': None, 'nslice': 1, 'rotation': 0.0, 'type': 'ExactQuad', 'unit': 0}
```

[...]

PR #1366, opened 2026-03-26
Lattice Element (De)Serialization Round-Trip

ax3l:
## Add lattice serialization with `to_dicts()` and `from_dicts()`
    
`KnownElementsList` now supports JSON/YAML-compatible [serialization](https://impactx--1366.org.readthedocs.build/en/1366/usage/python.html#impactx.elements.KnownElementsList.to_dicts):
- to_dicts(): serialize lattice elements to list of dictionaries
- from_dicts(): reconstruct lattice from dictionaries

`from_dicts()` works around known radian/degree `.to_dict()` bugs in four elements, see below.

## Add lattice code generation with `to_py()`

`KnownElementsList.to_py()` generates [executable Python code](https://impactx--1366.org.readthedocs.build/en/1366/usage/python.html#impactx.elements.KnownElementsList.to_py):
- Returns complete script with `get_lattice()` function
- Handles `SmallMatrix` parameters for `LinearMap`/`SpinMap`
- Applies radian-to-degree fixes for affected elements (see below)

When do you need this? After importing an external file format (MAD-X, PALS), you might want to save into a ImpactX Python lattice file for further inspection and manipulation. Basically for one-time imports to ImpactX.

# Bug Fixes

Follow-up to #1365 and #1332:
Ensure we expose enough properties in `to_dict` to rebuild elements in the constructor.

## (Loudly) Breaking Changes to `.to_dict()` Results to better match constructor arguments

These breaks cause visible asserts and are thus easy to fix without a silent error / mismatch being introduced:
- `Aperture`: `shape` and `action` are now properly returned as string, not as int
- `QuadEdge`: `flag` is now properly returned as a string, not as int
- `Kicker`: `unit` is now properly returned as a string, not as int
- `PolygonAperture`: `action` is now properly returned as a string, not as int
- `RFCavity`: `cos_coef`/`sin_coef` are now the `cos_coefficients`/`sin_coefficients` keys to match the constructor arguments
- `SoftQuadrupole`: `cos_coef`/`sin_coef` are now the `cos_coefficients`/`sin_coefficients` keys to match the constructor arguments
- `SoftSolenoid`: `cos_coef`/`sin_coef` are now the `cos_coefficients`/`sin_coefficients` keys to match the constructor arguments
- `Source`: `path` is now the `openpmd_path` key to match the constructor argument
- `LinearMap`: `transport_map` is now the `R` key to match the constructor argument
- for thin elements, `nslice` (==1) is omitted, to match that it does not appear in constructor arguments
- for thick elements, `nslice` is omitted if `1` because that is anyway the default argument in constructors

## Known Bugs with Backwards Compatible Fixes

Elements that take an angle in degrees in their constructor but expose radian #1367 in their `.to_dict()` and `.property` accessors:
- `ExactSbend`: `phi`
- `PlaneXYRot`: `angle`
- `PRot`: `phi_in`, `phi_out`
- `ThinDipole`: `theta`

The new higher-level  `sim.lattice` functions `.to_dicts()` and `.to_py()` address this already, by using underneath for these four classes the new `.to_dicts(in_degrees=True)` argument. The default of this argument is, for backwards compatibility, `False` - but if users use it with that they now get a `RuntimeWarning` and are advised of the inconsistency.

We have to fix them in a separate PR, potentially renaming the keys to cause a visible break instead of a silent translation that will lead to wrong results if we change degrees/rad, and implementing fallbacks where possible.

## Not fixed here

- `unit` is inconsistently handled as int or as string in various elements

### DIFF
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -823,6 +823,14 @@ This module provides elements and methods for the accelerator lattice.
       Each element is converted to a dictionary using its ``to_dict()`` method.
       The resulting list can be serialized to JSON, YAML, or other formats.
 
+      .. note::
+
+         This transforms the buggy ``.to_dict()`` keys of
+         ``ExactSbend``, ``PlaneXYRot``, ``PRot`` and ``ThinDipole``
+         to degrees, which by accident are written in radians.
+         See this comment in
+         `issue #1367 <https://github.com/BLAST-ImpactX/impactx/issues/1367#issuecomment-4160236826>`__.
+
       :return: List of element dictionaries
       :rtype: list[dict]
 
@@ -848,13 +856,6 @@ This module provides elements and methods for the accelerator lattice.
 
       Each dictionary should be in the format produced by ``to_dict()``,
       containing at minimum a ``type`` key identifying the element class.
-
-      .. note::
-
-         This transforms the dict keys of ``ExactSbend``, ``PlaneXYRot``,
-         ``PRot`` and ``ThinDipole`` to degrees, which by accident are written
-         in radians. See this comment in
-         `issue #1367 <https://github.com/BLAST-ImpactX/impactx/issues/1367#issuecomment-4160236826>`__.
 
       :param dicts: List of element dictionaries
       :type dicts: list[dict]
@@ -883,8 +884,9 @@ This module provides elements and methods for the accelerator lattice.
 
       .. note::
 
-         Like ``from_dicts()``, this transforms the dict keys of ``ExactSbend``,
-         ``PlaneXYRot``, ``PRot`` and ``ThinDipole`` angles from radians to degrees.
+         Like ``to_dicts()``, this transforms the buggy ``.to_dict()`` keys of
+         ``ExactSbend``, ``PlaneXYRot``, ``PRot`` and ``ThinDipole``
+         from radians to degrees.
 
       :return: Python source code
       :rtype: str
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -816,6 +816,63 @@ This module provides elements and methods for the accelerator lattice.
       :return: The transfer map of all elements in the list.
       :rtype: Map6x6
 
+   .. py:method:: to_dicts()
+
+      Serialize the lattice to a list of dictionaries.
+
+      Each element is converted to a dictionary using its ``to_dict()`` method.
+      The resulting list can be serialized to JSON, YAML, or other formats.
+
+      :return: List of element dictionaries
+      :rtype: list[dict]
+
+      **Example:**
+
+      .. code-block:: python
+
+         import json
+         from impactx import elements
+
+         lattice = elements.KnownElementsList([
+             elements.Drift(ds=1.0, name="d1"),
+             elements.Quad(ds=0.5, k=2.0, name="q1"),
+         ])
+
+         # Serialize to JSON
+         with open("lattice.impactx.json", "w") as f:
+             json.dump(lattice.to_dicts(), f, indent=2)
+
+   .. py:method:: from_dicts(dicts)
+
+      Load and append elements from a list of dictionaries.
+
+      Each dictionary should be in the format produced by ``to_dict()``,
+      containing at minimum a ``type`` key identifying the element class.
+
+      .. note::
+
+         This transforms the dict keys of `ExactSbend`, `PlanXYRot`,
+         `PRot` and `ThinDipole` to degrees, which by accident are written
+         in radians. See this comment in
+         `issue #1367 <https://github.com/BLAST-ImpactX/impactx/issues/1367#issuecomment-4160236826>`__.
+
+      :param dicts: List of element dictionaries
+      :type dicts: list[dict]
+
+      **Example:**
+
+      .. code-block:: python
+
+         import json
+         from impactx import elements
+
+         # Load from JSON
+         with open("lattice.impactx.json") as f:
+             data = json.load(f)
+
+         lattice = elements.KnownElementsList()
+         lattice.from_dicts(data)
+
    .. py:method:: plot_survey(ref=None, ax=None, legend=True, legend_ncols=5)
 
       Plot over s of all elements in the KnownElementsList.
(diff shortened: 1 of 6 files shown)


src/python/elements.cpp
```diff
@@ -194,7 +195,20 @@ namespace
...
+     * arguments where possible, used internally for copies and (de)serialization.
+     *
+     * The dict always includes a ``type`` string (the element class name) for dispatch;
+     * that key is not a constructor argument and must be omitted when unpacking, e.g.:
+     * ```py
+     * dr = elements.Drift(name="drift1", ds=1.0)
+     * d = dr.to_dict()
+     * kwargs = {k: v for k, v in d.items() if k != "type"}
+     * dr2 = elements.Drift(**kwargs)
+     * ```
```
ax3l, line 212: @RemiLehe serialize/deserialize

ax3l: This will get a proper helper later on.

ax3l:
I tested round-trips with this auto-generated file, but it feels a bit too awkward to commit just yet: [test_element_to_dict.py.txt](https://github.com/user-attachments/files/26371123/test_element_to_dict.py.txt)

src/python/elements.cpp
ax3l: Implemented `sim.lattice.to_dicts()` and `.from_dicts()` now.

ax3l:
The auto-generated test now looks like this: [test_element_to_dict.py.txt](https://github.com/user-attachments/files/26372361/test_element_to_dict.py.txt)

Still unsure if I want to commit the test file as is or merge for now w/o it, to decide.

docs/source/usage/python.rst
```diff
@@ -816,6 +816,102 @@ This module provides elements and methods for the accelerator lattice.
...
+
+      Each dictionary should be in the format produced by ``to_dict()``,
+      containing at minimum a ``type`` key identifying the element class.
+
+      .. note::
+
+         This transforms the dict keys of ``ExactSbend``, ``PlaneXYRot``,
+         ``PRot`` and ``ThinDipole`` to degrees, which by accident are written
+         in radians. See this comment in
+         `issue #1367 <https://github.com/BLAST-ImpactX/impactx/issues/1367#issuecomment-4160236826>`__.
```
ax3l: Maybe it is better to move this work-around directly to `.to_dicts()`

docs/source/usage/python.rst
```diff
@@ -816,6 +816,102 @@ This module provides elements and methods for the accelerator lattice.
...
+
+      Generate Python code that recreates this lattice.
+
+      Returns a string containing a complete Python script with imports
+      and a ``get_lattice()`` function that returns a KnownElementsList
+      with all elements.
+
+      .. note::
+
+         Like ``from_dicts()``, this transforms the dict keys of ``ExactSbend``,
```
ax3l: If we move the work-around:
```suggestion
         Like ``to_dicts()``, this transforms the dict keys of ``ExactSbend``,
```

src/python/impactx/extensions/KnownElementsList.py
```diff
@@ -521,6 +521,247 @@ def has_kind(self, kind_pattern) -> bool:
...
+        kwargs["phi"] = _rad2deg(kwargs["phi"])
+    elif element_type == "PlaneXYRot" and "angle" in kwargs:
+        kwargs["angle"] = _rad2deg(kwargs["angle"])
+    elif element_type == "PRot":
+        if "phi_in" in kwargs:
+            kwargs["phi_in"] = _rad2deg(kwargs["phi_in"])
+        if "phi_out" in kwargs:
+            kwargs["phi_out"] = _rad2deg(kwargs["phi_out"])
+    elif element_type == "ThinDipole" and "theta" in kwargs:
+        kwargs["theta"] = _rad2deg(kwargs["theta"])
```
ax3l: Maybe we rather move this work-around to the `.to_dicts()` than `.from_dicts()` method.

src/python/impactx/extensions/KnownElementsList.py
```diff
@@ -521,6 +521,247 @@ def has_kind(self, kind_pattern) -> bool:
...
+            with open("lattice.impactx.json", "w") as f:
+                json.dump(data, f, indent=2)
+
+    Note:
+        Elements with matrix parameters (LinearMap, SpinMap) contain
+        AMReX SmallMatrix objects that require custom JSON encoding.
+        Use :func:`impactx.extensions.ImpactXEncoder` for JSON serialization
+        of such elements.
+    """
+    return [el.to_dict() for el in self]
```
ax3l: Todo: move radian work-arounds here.

src/python/elements.cpp
```diff
@@ -2305,17 +2370,18 @@ void init_elements(py::module& m)
...
                  );
              }
         )
         .def("to_dict",
             [](PRot const & prot) {
                 return element_dict(
                     prot,
-                    std::make_pair("phi_in", prot.m_phi_in),
-                    std::make_pair("phi_out", prot.m_phi_out)
+                    std::make_pair("phi_in", prot.m_phi_in),   // BUG: needs / PRot::degree2rad),
```
ax3l: For buggy methods, a non-breaking change could be:
- add a default parameter `bool in_degrees=false`
- convert to degrees if true
- raise a warning if not `true`
- use with `true` from `.to_dicts()`

docs/source/usage/python.rst
ax3l: simplified

docs/source/usage/python.rst
ax3l: simplified

src/python/impactx/extensions/KnownElementsList.py
ax3l: simplified

src/python/impactx/extensions/KnownElementsList.py
ax3l: simplified

src/python/elements.cpp
ax3l: implemented as described

ax3l:
@cemitch99 @egstern @RemiLehe this is ready now from my end :)

ax3l:
cc @nikitakuklev just a heads-up: in case you use element `.to_dict()` there are necessary breaking changes in this PR

cemitch99 (approved):
Looks great to me.

PR #1369, opened 2026-03-27
Add example of RF cavity phase optimization.

cemitch99:
This PR adds an example that illustrates how to set the RF phase values for thick RFCavity elements by using a combination of the hook callback feature and a small optimization loop (using reference particle tracking) in Python to find the phase of maximum energy gain.

### DIFF
--- a/examples/rfcavity/README.rst
+++ b/examples/rfcavity/README.rst
@@ -160,3 +160,52 @@ We run the following script to analyze correctness:
    .. literalinclude:: analysis_rfcavity_ref_part_hook.py
       :language: python3
       :caption: You can copy this file from ``examples/rfcavity/analysis_rfcavity_ref_part_hook.py``.
+
+
+
+.. _examples-rfcavity-ref-part-opt:
+
+Proton acceleration by RF Cavities (Using RF phase optimization)
+================================================================
+   
+This test is similar to the :ref:`test above <examples-rfcavity-ref-part-hook>`, except that the test is performed for protons (at the same energy).  As a result, there is a large variation of relativistic
+beta (in the first RF cavity), and the analytical methods of the previous example cannot be used.  
+
+To treat this case, the callback hook feature :py:attr:`~impactx.ImpactX.hook` is combined with an optimization loop (using scipy.minimize_scalar) to determine, for each RF cavity, the input phase setting
+that results in maximum energy gain.  The phase is reset to this value (for each cavity).
+
+The functions used for optimization of the RF phase are contained in the file ``phase_opt.py``.
+   
+In this test, the initial and final reference values of :math:`s` and :math:`\gamma` must agree with nominal values.
+          
+
+Run
+---       
+
+This example can be run as:
+
+* **Python** script: ``python3 run_rfcavity_ref_part_opt.py``
+          
+For `MPI-parallel <https://www.mpi-forum.org>`__ runs, prefix these lines with ``mpiexec -n 4 ...`` or ``srun -n 4 ...``, depending on the system.
+
+.. tab-set::
+
+   .. tab-item:: Python: Script
+
+       .. literalinclude:: run_rfcavity_ref_part_opt.py
+          :language: python3
+          :caption: You can copy this file from ``examples/rfcavity/run_rfcavity_ref_part_opt.py``.
+      
+   
+Analyze
+-------
+          
+We run the following script to analyze correctness:
+
+.. dropdown:: Script ``analysis_rfcavity_ref_part_opt.py``
+
+   .. literalinclude:: analysis_rfcavity_ref_part_opt.py
+      :language: python3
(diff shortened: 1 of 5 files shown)


examples/rfcavity/run_rfcavity_ref_part_opt.py
```diff
@@ -0,0 +1,111 @@
...
+        rf,
+        dr2,
+        monitor,
+    ]
+)
+
+
+def hook_before_element(sim):
+    element = sim.tracking_element
+    if element.name == "rf":
```
cemitch99: Here, it would be nice to select based on the element kind.  Something like
`if element.kind == "RFCavity":`

ax3l: That works, you can do this in Python :)
```suggestion
    if type(element) is elements.RFCavity:
```

examples/rfcavity/phase_opt.py
```diff
@@ -0,0 +1,61 @@
...
+    loss = KE_in - KE_fin
+
+    if np.isnan(loss):
+        loss = 1.0e99
+
+    return loss
+
+
+def optimize(ref, element):
+
```
ax3l, line 60: Maybe a tiny function doc string here on what we optimize for: I assume we optimize for maximum energy gain (i.e. `in - out` MeV is a negative number of we gain energy, we try to find the smallest number in MeV)

examples/rfcavity/README.rst
```diff
@@ -160,3 +160,50 @@ We run the following script to analyze correctness:
...
+
+For `MPI-parallel <https://www.mpi-forum.org>`__ runs, prefix these lines with ``mpiexec -n 4 ...`` or ``srun -n 4 ...``, depending on the system.
+
+.. tab-set::
+
+   .. tab-item:: Python: Script
+
+       .. literalinclude:: run_rfcavity_ref_part_opt.py
+          :language: python3
+          :caption: You can copy this file from ``examples/rfcavity/run_rfcavity_ref_part_opt.py``.
```
ax3l, line 197: You could as in other parts either also include `phase_opt.py` for the user to see, or maybe you just move the two functions in `phase_opt.py` into the `run*.py` script and have them in one place?

Preview: https://impactx--1369.org.readthedocs.build/en/1369/usage/examples/rfcavity/README.html#proton-acceleration-by-rf-cavities-using-rf-phase-optimization

ax3l (approved):
This looks great!

Maybe a tiny detail function comment on the optimize function we call from the sim and then this is ready! :)

examples/rfcavity/phase_opt.py
```diff
@@ -0,0 +1,61 @@
...
+    # optimizer specific options
```
cemitch99, line 61:
```suggestion
    # This function is designed to maximize the energy gain of a reference particle
    #  (ref) in an RFCavity (element).  Optimization is performed by minimizing 
    #  -(change in kinetic energy in MeV).
    
    # optimizer specific options
```

examples/rfcavity/phase_opt.py
```diff
@@ -0,0 +1,65 @@
...
+        loss = 1.0e99
+
+    return loss
+
+
+def optimize(ref, element):
+
+    # This function is designed to maximize the energy gain of a reference particle
+    #  (ref) in an RFCavity (element).  Optimization is performed by minimizing
+    #  -(change in kinetic energy in MeV).
```
ax3l: Actually like you do above is best:
```suggestion
    """
    Maximize the energy gain of a reference particle
    
    Using (ref) in an RFCavity (element): the optimization is performed by minimizing
    -(change in kinetic energy in MeV).
    """
```
That way, `help(optimize)` and in IPython/Jupyter notebooks `optimize?` shows you these docs as a user :)

examples/rfcavity/phase_opt.py
```diff
@@ -0,0 +1,67 @@
...
+
+    return loss
+
+
+def optimize(ref, element):
+    """
+    Maximize the energy gain of a reference particle
+
+    Using (ref) in an RFCavity (element): the optimization is performed by minimizing
+    -(change in kinetic energy in MeV).
```
ax3l, line 47:
```suggestion
    -(change in kinetic energy in MeV).

    Parameters
    ----------
    ref:
      the reference particle at RF entry
    element:
      the RF cavity element

    Returns
    -------
    The optimized phase and energy gain at that phase (phase_opt, e_gain).
```

examples/rfcavity/README.rst
cemitch99: I lean toward keeping `phase_opt.py` as a separate file, to simplify re-use by the user without cluttering the Python input.  If this works, I'll add comments and make this visible via the README.

ax3l (approved):
🚀 ✨

PR #1372, opened 2026-03-29
Fix domain segfault before init_grids

### LINKED ISSUE #1371: `sim.domain` segfaults before `init_grids()`
Setting `ImpactX.domain` before `init_grids()` segfaults instead of storing the static box and letting initialization continue later. This is because binding calls `ix.ResizeMesh()` unconditionally, which of course is a bad idea before init. Other properties like `n_cell` and `max_level` already guard against this with `if (ix.initialized()) throw std::runtime_error(...)`, but `domain` does not.

Minimal Reproducer:

```python
from impactx import ImpactX, amr

sim = ImpactX()
sim.domain = amr.RealBox([-1, -1, -1], [1, 1, 1])  # segfault
```

Also need to add to docs warning about requiring prior initialization.

Not contained in https://github.com/BLAST-ImpactX/impactx/issues/1354

nikitakuklev:
Fixes #1371 

Note - the modified `prob_lo`/`prob_hi` and `dynamic_size=false` will get correctly picked up by init_grids() when it is called later. We could also throw here, not sure which is better.

### DIFF
--- a/src/python/ImpactX.cpp
+++ b/src/python/ImpactX.cpp
@@ -166,7 +166,8 @@ void init_ImpactX (py::module& m)
 
                 pp_geometry.add("dynamic_size", false);
 
-                ix.ResizeMesh();
+                if (ix.initialized())
+                    ix.ResizeMesh();
             },
             "The physical extent of the full simulation domain, relative to the reference particle position, in meters."
         )


ax3l (approved):
Looks great! Thank you!

PR #1374, opened 2026-03-29
Fix prob_relative getter for multi-level lists

### LINKED ISSUE #1373: `prob_relative` getter is broken for multi-level lists
`sim.prob_relative` is documented as a list with `amr.max_level + 1` entries, but reading it back fails when more than one entry is present.

### Repro:
```python
from impactx import ImpactX

sim = ImpactX()
sim.max_level = 1
sim.prob_relative = [1.2, 1.1]
print(sim.prob_relative) # raises RuntimeError: syntax error in Parser expression "1.21.1000000000000001"
```

### Claude-cause:
In the Python binding, the setter accepts a `std::vector<amrex::Real>` and stores it with `addarr`, but the getter reads it back as a single `amrex::Real`:

Relevant code:

```cpp
.def_property("prob_relative",
      [](ImpactX & /* ix */) {
          return detail::get_or_throw<amrex::Real>("geometry", "prob_relative");
      },
      [](ImpactX & /* ix */, std::vector<amrex::Real> frac) {
          amrex::ParmParse pp_geometry("geometry");
          pp_geometry.addarr("prob_relative", frac);
      },
```

That getter should likely use `queryarr` / `getarr` and return a vector/list, not a scalar.

nikitakuklev:
Use getarr to return a vector, matching the setter and documentation. The scalar get_or_throw concatenated array entries into a single unparseable string.

Fixes #1373

### DIFF
--- a/src/python/ImpactX.cpp
+++ b/src/python/ImpactX.cpp
@@ -173,7 +173,10 @@ void init_ImpactX (py::module& m)
 
         .def_property("prob_relative",
               [](ImpactX & /* ix */) {
-                  return detail::get_or_throw<amrex::Real>("geometry", "prob_relative");
+                  amrex::ParmParse const pp_geometry("geometry");
+                  std::vector<amrex::Real> frac;
+                  pp_geometry.getarr("prob_relative", frac);
+                  return frac;
               },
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Thank you, good catch!

PR #1368, opened 2026-03-27
Elements Lifetime (GPU)

ax3l:
On GPU, we have to more carefully deal with the lifetime of dynamic (array) data in elements. This PR is an infrastructure update for complex elements that have such dynamic GPU data, to address multiple usage bugs on GPU.

## Part 1: This PR
- [x] ensure we can create elements before an initialized simulation object
- [x] add unit tests
- [x] clean them up more carefully after a simulation ends, incl. no error for elements that were _never added_ to `sim.lattice`
- [x] we can even reuse the same elements between different simulations (with AMReX finalize and another init in between!)
- [x] similar logic for beam monitor/source: do not over-eagerly open files, but lazily on first use. Close on finalize (even if not in lattice): already done in #1178

## Part 2: Follow-up PR

- make `sim.lattice` behave like a Python list: shared instead of copied elements #1381

PR #1381, opened 2026-04-03
Pythonic `sim.lattice` List

ax3l:
# Better `sim.lattice`

`sim.lattice` behaves like a Python list: it holds the elements it is given, rather than copies of them.

```python
q = elements.Quad(ds=0.3, k=2.0)
sim.lattice.append(q)

assert sim.lattice[0] is q
q.k = 3.0                       # applies to tracking
```

Follow-up to #1368, part 2 of the elements lifetime work.

## ⚠️ Behavior change

**Adding the same variable twice now places one element at two positions, where it previously placed independent copies.** Retuning it then affects both.

```python
sim.lattice.extend([q, q])      # one element, two positions
sim.lattice.extend(cell * 3)    # 6 positions, 2 elements
```

That is usually what a periodic channel wants. Where positions are tuned separately, construct one per position or add `element.copy()`. Aliasing is harmless without per-position state — a `BeamMonitor` at head and tail, a drift in a FODO cell — and every in-tree example was checked.

`sim.tracking_element` now yields the element being tracked, so a hook writing to it changes the next push. `impactx.reverse(el)` now reverses the supplied element in place; previously it reversed a temporary copy. `impactx.push(pc, el, …)` already advanced particles, and now also preserves changes to the supplied element's internal state instead of discarding them with a temporary copy.

Migration is in the new **[Upgrade Guide](https://impactx--1381.org.readthedocs.build/en/1381/install/upgrade.html)**, where breaking changes are documented from now on; workflows in **[Manipulate a Lattice](https://impactx--1381.org.readthedocs.build/en/1381/usage/howto/lattice_manipulation.html)**.

## What is new

- `element.copy()` on every element type; keyword arguments override values, so one element supplies many: `[quad.copy(k=k) for k in scan]`.
- Additional list operations: slices, negative indices, `insert()`, `remove()`, `index()`, `count()`, `in`, `del`, `lattice[i] = element`, `reversed()`, `pop_back()`. Membership and lookup match on element identity. Iteration and `reversed()` take a snapshot of element references. This is a list-like sequence, with the supported API documented in the manual.
- Array-valued parameters (Fourier and multipole coefficients, polygon vertices) can be set after construction, individually or as a validated pair.
- Elements keep their Python subclass, attributes and callbacks in the lattice.
- The sequence is fixed while tracking walks it: a hook that adds or removes elements raises. Retuning from a hook is unaffected.
- A selection knows when the lattice changed under it, via a new `lattice.generation` counter.

## Internals

- `GPUDataRegistry` is gone. The six elements with dynamic arrays hold `amrex::Gpu::TrackedVector` members directly, with ordinary RAII. Its `clear()` had no caller, so such elements kept their arrays for the process lifetime: over 800 construct/destroy cycles, peak RSS growth went ~49 MiB → 0.
- Elements split into `Element<T_GPU>`: the state a push kernel is handed (`DriftGPU`, `ChrQuadGPU`, …) and the host-only metadata that never enters one. A compile-time check rejects host-only state in a launch closure.
- `make_particle_pusher()` runs once per element slice instead of once per tile.
- The lattice stores a `std::vector<ElementHandle>` and is held by `shared_ptr`, making indexed access constant-time. A Python view can outlive its simulation, but accessing it after the simulation is destroyed raises.
- Bulk edits rebuild in one pass. `del lattice[:n//2]`, `lattice[a:b] = …` and a filtered `delete()` were quadratic; at 16 000 elements the first went 120 ms → 2.2 ms.
- Filtered edits change only the selected positions and prepare and validate all replacements before installing any. Failed preparation or validation installs no replacements; edits made by user callbacks, property getters or finalizers are not rolled back. Finalizers can also invalidate the returned selection.
- Headers that are not elements moved to `elements/helper/`; `mixin/` holds only classes elements derive from.

## Testing

- `pytest.ImpactX`: 644 passed. `ctest` examples green, including the aperture, envelope, `rfcavity`, `FODO` and `chicane` suites (243/243).
- Physics unchanged under aliasing: one shared `RFCavity` at four positions gives bit-identical results to four copies.
- Ten new files: lattice `identity`, `robustness`, `consistency`, `scaling`, `view_lifetime`, `traversal_guard`; element `copy`, `coefficients`, `array_validation`, `metadata`.

Rebased onto `development` (`ef5141baa`): the 26.09 release, the single lattice traversal (#1643) and the pipe-aperture work (#1644, #1648, #1652). The upgrade-guide entry targets 26.10.


egstern:
Wait a second! Did I misunderstand what is meant here? When an element is appended to the lattice, that should be a copy not a reference because each element represents a separate physical object that can have its own properties independentally of any other beamline element.

src/python/elements.cpp
```diff
@@ -2933,14 +2634,147 @@ void init_elements(py::module& m)
...
+    //
+    // Ref semantics (Python vs this binding): a Python list keeps references to mutable
+    // objects, but here each append/extend materializes a C++ KnownElements value in this
+    // list (copy or move). The original Python handle is a separate object unless we
+    // intentionally bind shared ownership (e.g. std::shared_ptr element types later).
+    //
+    // TODO: register GPU-heavy element types with py::class_<T, std::shared_ptr<T>> and
+    //       store std::shared_ptr in KnownElements (or a thin wrapper) so append/extend
+    //       match Python mutable-object reference semantics.
+    // TODO: optional weak_ptr to a SimulationLifetime/session object on elements for
```
ax3l: This design part is outdated, we now have `TrackedVector`.

Also, we want to get rid of `GPUDataRegistry` altogether now with this redesign.

ax3l:
> Wait a second! Did I misunderstand what is meant here? When an element is appended to the lattice, that should be a copy not a reference because each element represents a separate physical object that can have its own properties independentally of any other beamline element.

Yes, that is still possible by using either in-place construction in `append()` (the common pattern we use now) or by using `.copy()` on a variable or an element. That way, we do use standard Python list semantics and can make sure we can construct elements across simulation instances cleanly without leaking resources.

docs/source/install/upgrade.rst
```diff
@@ -0,0 +1,161 @@
...
+
+   def build():
+       sim = ImpactX()
+       sim.lattice.extend(cell)
+       return sim.lattice        # the simulation ends here
+
+   lattice = build()
+   len(lattice)                  # raises
+
+Return the simulation, or keep a reference to it for as long as you use its lattice.
```
ax3l, line 113:
```suggestion
Return the simulation, or keep a reference to it for as long as you use its lattice.
Alternatively, keep the lattice alive across simulations by storing it in a pure Python list and call `sim.lattice.extend` directly before tracking.
```

docs/source/install/upgrade.rst
```diff
@@ -0,0 +1,161 @@
...
+
+Selections describe the lattice they were taken from
+""""""""""""""""""""""""""""""""""""""""""""""""""""""
```
ax3l:
```suggestion
Selections describe the lattice they were taken from
""""""""""""""""""""""""""""""""""""""""""""""""""""
```

docs/source/install/upgrade.rst
```diff
@@ -0,0 +1,161 @@
...
+The template is copied into each selected position, so a Python subclass must define
+``copy()``. Passing one that does not raises, where it previously inserted plain
+base-class elements.
+
+``finalize()`` always empties the lattice
+"""""""""""""""""""""""""""""""""""""""""
+
+``sim.finalize()`` releases the elements whether or not ``init_grids()`` was ever called.
+They have been finalized by then, so keeping them would leave elements that are done with
+-- a beam monitor with its output closed, say -- still in the lattice.
```
ax3l:
```suggestion
, e.g., a beam monitor with its output closed, still in the lattice.
```

docs/source/install/upgrade.rst
```diff
@@ -0,0 +1,161 @@
...
+
+New in this release
+"""""""""""""""""""
+
+- ``element.copy()`` on every element type, for a new element with the same configuration.
+- The array-valued element parameters -- Fourier coefficients, multipole coefficients and
+  polygon vertices -- can be set after construction, via their properties or the paired
```
ax3l:
```suggestion
- The array-valued element parameters (Fourier coefficients, multipole coefficients and
  polygon vertices) can be set after construction, via their properties or the paired
```

docs/source/usage/python.rst
```diff
@@ -1412,13 +1492,20 @@ This module provides elements and methods for the accelerator lattice.
...
+   parameters moves nothing and leaves selections usable.
+
+   Mutating operations rewrite only the positions selected. Elements at other positions are
+   left exactly as they are, keeping their identity, their Python subclass and their
+   attributes. Every replacement is prepared before any is installed, so an operation that
+   fails leaves the lattice unchanged.
 
    If the selection is empty, ``delete`` is a no-op and ``replace_*`` return
-   an empty ``FilteredElementsList``.
+   an empty ``FilteredElementsList``; neither affects other selections.
```
```suggestion
   an empty ``FilteredElementsList``.
```

cemitch99:
@ax3l  Some significant merge conflicts here.

ax3l:
@cemitch99 thanks, rebased now and resolved :)

This will keep conflicting with other PRs, so I would merge it soon.

docs/source/usage/howto/lattice_manipulation.rst
```diff
@@ -0,0 +1,247 @@
...
+   ``cell.copy()``, ``cell[:]`` and ``list(cell)`` each give a new list holding the same
+   element objects, so this still places the very elements ``cell`` holds:
+
+   .. code-block:: python
+
+      sim.lattice.extend(cell.copy())   # a new list, the same elements
+
+   `copy.deepcopy() <https://docs.python.org/3/library/copy.html#copy.deepcopy>`__ is not
+   an alternative either: elements cannot be pickled.
+   Copy the elements, as above, to get independent ones.
```
ax3l:
```suggestion
   `copy.deepcopy() <https://docs.python.org/3/library/copy.html#copy.deepcopy>`__
   also copies the list, but references the same elements.
```

docs/source/usage/howto/lattice_manipulation.rst
```diff
@@ -0,0 +1,247 @@
...
+
+.. code-block:: python
+
+   square = elements.PolygonAperture(vertices_x=[-1.0, 1.0, 1.0, -1.0, -1.0],
+                                     vertices_y=[-1.0, -1.0, 1.0, 1.0, -1.0])
+   wider = square.copy(vertices_x=[-2.0, 2.0, 2.0, -2.0, -2.0],
+                       vertices_y=[-2.0, -2.0, 2.0, 2.0, -2.0])
+
+Setting a parameter the element does not have raises, so a mistyped name is reported
+rather than quietly ignored.
```
ax3l:
```suggestion
Setting a parameter the element does not have raises an error, e.g., for typos.
```

docs/source/usage/howto/lattice_manipulation.rst
```diff
@@ -0,0 +1,247 @@
...
+   :py:meth:`element.copy() <impactx.elements.Element.copy>` gives a new element with the same configuration.
+   A copy of a beam monitor does not inherit an already-open output file, and an element
+   written as a Python subclass must define ``copy()`` for this to work.
+
+Keeping a Lattice Longer Than a Simulation
+------------------------------------------
+
+A lattice reached through :py:attr:`sim.lattice <impactx.ImpactX.lattice>` belongs to that
+simulation and does not keep it
+alive; using it after the simulation is gone raises.
```
ax3l:
```suggestion
A lattice accessed through :py:attr:`sim.lattice <impactx.ImpactX.lattice>` belongs to that
simulation, using it after the simulation is gone raises an error.
```

docs/source/usage/howto/lattice_manipulation.rst
```diff
@@ -0,0 +1,247 @@
...
+   for run in range(runs):
+       sim = ImpactX()
+       # ... set up the beam ...
+       sim.lattice.extend(cell)  # the same elements each run
+       sim.track_particles()
+       sim.finalize()
+
+The list outlives every simulation, and each run tracks the same element objects.
+Retuning one between runs therefore applies to the runs that follow, which is what a
+sequence of dependent runs wants; use :py:meth:`element.copy() <impactx.elements.Element.copy>` where the runs
```
ax3l:
```suggestion
Retuning a lattice list between runs therefore applies to the runs that follow.
Use :py:meth:`element.copy() <impactx.elements.Element.copy>` where the runs
```

docs/source/usage/python.rst
```diff
@@ -1658,9 +1801,40 @@ Lattice Elements
...
+Use ``copy()`` when a second, separately tunable element is wanted:
+
+.. code-block:: python
+
+   focus = elements.Quad(ds=0.3, k=2.0)
+   defocus = focus.copy()
+   defocus.k = -2.0
+
+The elements carrying array-valued parameters -- the soft-edge elements, the exact multipole
+and combined-function bend, and the polygon aperture -- expose those arrays as properties,
```
```suggestion
The elements carrying array-valued parameters (e.g., the soft-edge elements, the exact multipole
and combined-function bend, and the polygon aperture) expose those arrays as properties,
```

examples/rfcavity/run_rfcavity_ref_part_opt.py
```diff
@@ -52,6 +53,7 @@
     mapsteps=100,
     nslice=4,
 )
+cavities = [rf] + [rf.copy(name=f"rf{i}") for i in range(2, 5)]
```
ax3l:
```suggestion
rf = [rf_tpl.copy(name=f"rf{i}") for i in range(1, 5)]
```

examples/rfcavity/run_rfcavity_ref_part_opt.py
```diff
@@ -39,9 +39,10 @@
 dr1 = elements.Drift(name="dr1", ds=0.4, nslice=1)
 dr2 = elements.Drift(name="dr2", ds=0.032997, nslice=1)
 
-#   RF cavity element
+#   RF cavity elements: four physically distinct cavities, built from one template so
+#   that the on-axis field is fitted once
 rf = elements.RFCavity(
```
ax3l:
```suggestion
rf_tpl = elements.RFCavity(
```

examples/rfcavity/run_rfcavity_ref_part_opt.py
```diff
@@ -62,16 +64,16 @@
...
-        rf,
+        cavities[1],
         dr2,
         dr2,
-        rf,
+        cavities[2],
         dr2,
         dr2,
-        rf,
+        cavities[3],
```
ax3l:
```suggestion
        rf[0],
        dr2,
        dr2,
        rf[1],
        dr2,
        dr2,
        rf[2],
        dr2,
        dr2,
        rf[3],
```

examples/rfcavity/run_rfcavity_ref_part_opt.py
```diff
@@ -39,9 +39,10 @@
 dr1 = elements.Drift(name="dr1", ds=0.4, nslice=1)
 dr2 = elements.Drift(name="dr2", ds=0.032997, nslice=1)
 
-#   RF cavity element
+#   RF cavity elements: four physically distinct cavities, built from one template so
+#   that the on-axis field is fitted once
 rf = elements.RFCavity(
-    name="rf",
+    name="rf1",
```
ax3l, line 45:
```suggestion
    name="rf",
```

docs/source/install/upgrade.rst
```diff
@@ -0,0 +1,176 @@
...
+.. code-block:: python
+
+   q = elements.Quad(ds=0.3, k=2.0)
+   sim.lattice.append(q)
+
+   assert sim.lattice[0] is q
+   q.k = 3.0                       # applies to tracking
+
+**What to check in your scripts:** adding the same variable more than once now places one
+element at several positions, where it used to place independent copies. Retuning it
```
cemitch99: The first sentence is a little confusing.  Maybe better to make explicit:  What is the old behavior, and what is the new behavior?

docs/source/usage/howto/lattice_manipulation.rst
```diff
@@ -0,0 +1,244 @@
+.. _usage-howto-lattice-manipulation:
+
+Manipulate a Lattice
+====================
+
+:py:attr:`sim.lattice <impactx.ImpactX.lattice>` is a sequence of elements that behaves like a Python list.
+It **holds references to** the elements it is given rather than copies of them, so an element you keep a variable to is the element that is tracked, and the same element may sit at several positions.
```
cemitch99: By "you keep a variable to", do you mean "you pass a variable to"?

ax3l: Yes, maybe writing it inverse, "a lattice that holds an element that you keep a variable to" is the long form... Let me see how to rephrase this, the examples below are showing it clearer than the prose.

docs/source/install/upgrade.rst
```diff
@@ -0,0 +1,176 @@
...
+afterwards affects every one of those positions.
```
ax3l:
```suggestion
**What to check in your scripts:** previously, when storing an element in a variable and adding that variable more than once used to place independent copies.
Now, this *places the same element at several positions*.
Changing the variable (tuning the element) afterwards affects every one of those positions.
```

docs/source/install/upgrade.rst
ax3l: Like this? :)
https://github.com/BLAST-ImpactX/impactx/pull/1381/changes#r3981361386

docs/source/usage/howto/lattice_manipulation.rst
ax3l: Is this clearer?
```suggestion
When you add elements to ``sim.lattice`` it will **hold references to** the Python variables it is given rather than copies of them.
Concretely: if you create an element as a Python variable, append it to ``sim.lattice``, you can continue to change and manipulate the element that now sits inside `sim.lattice` through the variable (see below).
_The same element (variable) may sit at several lattice positions_, making tuning and ramping operations for whole channels, ring arcs, etc. easy.
```

docs/source/usage/howto/lattice_manipulation.rst
```diff
@@ -0,0 +1,244 @@
...
+
+Adding an element borrows the object to the lattice.
+Changing it afterwards changes what is tracked:
+
+.. code-block:: python
+
+   q = elements.Quad(ds=0.3, k=2.0)
+   sim.lattice.append(q)
+
+   q.k = 3.0                     # the lattice tracks q, so this applies
```
ax3l:
```suggestion
   q.k = 4.0              # sim.lattice references variable q, this changes the lattice

   sim.track_particles()  # tracks through Quad(ds=0.3, k=4.0)
```

docs/source/usage/howto/lattice_manipulation.rst
```diff
@@ -0,0 +1,244 @@
...
+.. code-block:: python
+
+   sim.lattice.extend([q, q])    # one element, two positions
+
+For elements that are tuned separately, construct one per position or add a copy:
+
+.. code-block:: python
+
+   sim.lattice.append(elements.Quad(ds=0.3, k=2.0))
+   sim.lattice.append(elements.Quad(ds=0.3, k=2.0))   # two elements
```
ax3l:
```suggestion
   sim.lattice.append(elements.Quad(ds=0.3, k=2.0))   # two distinct elements
```

docs/source/usage/howto/lattice_manipulation.rst
```diff
@@ -0,0 +1,244 @@
...
+
+   template = elements.Quad(ds=0.3, k=2.0)
+   sim.lattice.extend([template.copy(), template.copy()])
```
ax3l:
```suggestion
   sim.lattice.extend([template.copy(), template.copy()])  # also two distinct elements
```

docs/source/usage/howto/lattice_manipulation.rst
```diff
@@ -0,0 +1,244 @@
...
+----------------
+
+:py:meth:`~impactx.elements.KnownElementsList.extend` holds every element of the list it is given, so multiplying a list
+repeats its elements rather than duplicating them:
+
+.. code-block:: python
+
+   cell = [elements.Drift(ds=0.25), elements.Quad(ds=1.0, k=1.0)]
+
+   sim.lattice.extend(cell * 3)          # 6 positions, 2 elements
```
ax3l:
```suggestion
   sim.lattice.extend(cell * 3)          # 6 positions, 2 elements used repeatedly
```

cemitch99 (approved):

src/Lattice.H
```diff
@@ -0,0 +1,335 @@
...
+                throw std::runtime_error(
+                    std::string("Lattice: cannot ") + what + " while tracking through it. "
+                    "Element parameters may be changed during tracking, but the sequence of "
+                    "elements may not.");
+            }
+        }
+
+        storage_type m_elements;
+        std::uint64_t m_generation = 0;
+        mutable std::size_t m_active_traversals = 0;
```
ax3l: These three members need inline one-line Doxygen comments.

src/particles/wakefields/CSRBendElement.H
```diff
@@ -40,22 +40,19 @@ namespace impactx::particles::wakefields
         amrex::Real R = 0;
         bool element_has_csr = false;
 
-        if (std::holds_alternative<elements::Sbend>(element_variant))
+        if (auto const * sbend = elements::get_if<elements::Sbend>(element_variant))
```
ax3l: For a follow-up PR: We should create a CSR and ISR mixing class for elements that support it, which is then also providing the helpers we need to query `R(ref)`.

PR #1382, opened 2026-04-03
Increase bithermal space charge test resolution

### LINKED ISSUE #1379: Extreme rtol in bithermal SC test
See https://github.com/BLAST-ImpactX/impactx/blob/278fb0fa49d92c3f60973e01d1e2db1e49917f20/examples/epac2004_benchmarks/analysis_bithermal.py#L83

Maybe rtol=6000/sqrt(N) was meant to be 6.0/sqrt(N), like for thermal case?

cemitch99:
This PR increases the resolution of the space charge test in: `examples/epac2004_benchmarks/run_bithermal.py`.  

- [x] adjust numerical resolution and tolerance
- [x] mark test as "slow" -- not needed anymore, now only 3.4x slower than `development`

This fixes #1379, where we had a 6000% tolerance. Now it is 7%.

### DIFF
--- a/examples/epac2004_benchmarks/analysis_bithermal.py
+++ b/examples/epac2004_benchmarks/analysis_bithermal.py
@@ -41,7 +41,7 @@ initial = series.iterations[1].particles["beam"].to_df()
 final = series.iterations[last_step].particles["beam"].to_df()
 
 # compare number of particles
-num_particles = 10000
+num_particles = 1000000
 assert num_particles == len(initial)
 assert num_particles == len(final)
 
@@ -80,7 +80,8 @@ print(
 )
 
 atol = 0.0  # ignored
-rtol = 6000 * num_particles**-0.5  # from random sampling of a smooth distribution
+#rtol = 6000 * num_particles**-0.5  # from random sampling of a smooth distribution
+rtol = 10.0 * num_particles**-0.5  # from random sampling of a smooth distribution
 print(f"  rtol={rtol} (ignored: atol~={atol})")
 
 assert np.allclose(
--- a/examples/epac2004_benchmarks/analysis_bithermal.py
+++ b/examples/epac2004_benchmarks/analysis_bithermal.py
@@ -80,7 +80,7 @@ print(
 )
 
 atol = 0.0  # ignored
-#rtol = 6000 * num_particles**-0.5  # from random sampling of a smooth distribution
+# rtol = 6000 * num_particles**-0.5  # from random sampling of a smooth distribution
 rtol = 10.0 * num_particles**-0.5  # from random sampling of a smooth distribution
 print(f"  rtol={rtol} (ignored: atol~={atol})")
 
--- a/examples/epac2004_benchmarks/analysis_bithermal.py
+++ b/examples/epac2004_benchmarks/analysis_bithermal.py
@@ -41,7 +41,7 @@ initial = series.iterations[1].particles["beam"].to_df()
 final = series.iterations[last_step].particles["beam"].to_df()
 
 # compare number of particles
-num_particles = 1000000
+num_particles = 10000
 assert num_particles == len(initial)
 assert num_particles == len(final)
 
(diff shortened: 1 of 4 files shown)


examples/epac2004_benchmarks/analysis_bithermal.py
```diff
@@ -80,7 +80,8 @@ def get_moments(beam):
 )
 
 atol = 0.0  # ignored
-rtol = 6000 * num_particles**-0.5  # from random sampling of a smooth distribution
+# rtol = 6000 * num_particles**-0.5  # from random sampling of a smooth distribution
+rtol = 10.0 * num_particles**-0.5  # from random sampling of a smooth distribution
```
ax3l: The old tolerance was `rtol=6000 / sqrt(10'000)` aka 6000% (anything passes).
The new tolerance is `1%` w/ `1'000'000` (good, but as you note very costly).

Do you think we can achieve something in the range of 5-10% tolerance to the analytical solution using maybe 100 instead of 400 slices and maybe around 100'000 particles instead of 1M?

I will automate this relaxation process and see what I find.

ax3l: I pushed an update that converges to 7% and does need less runtime:

  |                       | `development` | full resolution  | <7% tolerance |                      
  |-----------------------|---------------|-----------|-------------|                      
  | `npart`               | 10,000        | 1,000,000 | 10,000      |                      
  | `nslice`              | 50            | 400       | 120         |                      
  | `prob_relative`       | 3.0           | 1.5       | 1.1         |                      
  | `poisson_solver`      | default       | fft       | fft         |
  | rtol (final)          | 6000%         | 1%        | 6.0%          |
  | runtime (4 threads)   | 11s           | very long | 27s         |
  | slow label            | no            | yes       | no          |

ax3l: Sweep I ran: (w/ 4 OMP threads on my laptop, about 2x slower on 2 OMP threads as in CI)

  | `npart` | `nslice` | runtime | max init err | max final err |                         
  |--------:|---------:|--------:|-------------:|--------------:|                          
  |  10,000 |      120 |  26.6s  |        2.7%  |        6.0%   |
  |  10,000 |      150 |  32.8s  |        2.7%  |        6.6%   |                          
  |  10,000 |      175 |  38.0s  |        2.7%  |        5.0%   |                          
  |  20,000 |      120 |  26.5s  |        1.6%  |        5.9%   |                         
  |  20,000 |      150 |  32.7s  |        1.6%  |        6.8%   |                          
  |  20,000 |      175 |  38.4s  |        1.6%  |        5.4%   |
  |  50,000 |      120 |  27.0s  |        1.3%  |        6.6%   |                          
  |  50,000 |      150 |  33.5s  |        1.3%  |        6.9%   |
  |  50,000 |      175 |  39.0s  |        1.3%  |        5.9%   |                          

Notes:
  - `nslice` dominates the error: `nslice=50` is catastrophic regardless of `npart`             
  - `npart` barely affects runtime (grid-solve dominated) but reduces init beam noise

ax3l: Just for comparison:

Here's the convergence toward 1% (100k particles, ncell=64, prob_relative=1.1):

  | `nslice` | OMP=2 err | OMP=4 err | worst | runtime (OMP=4) |
  |---------:|----------:|----------:|------:|----------------:|
  |      200 |     4.0%  |     4.6%  |  4.6% |            32s  |
  |      300 |     2.1%  |     2.8%  |  2.8% |            70s  |
  |      400 |     1.7%  |     2.3%  |  2.3% |            60s  |
  |      600 |     1.4%  |     2.0%  |  2.0% |            91s  |

The error is plateauing around 1.5-2% — the grid resolution (ncell=64) becomes the floor, not integration. Getting to 1% would likely require ncell=128 (8x more expensive grid solve per slice) on top of high nslice. That would push runtimes to many minutes.

Also notable: OMP=4 consistently gives ~0.5-0.8% worse results than OMP=2, probably from different floating-point accumulation order.

Bottom line: 1% tolerance is expensive with this test. It would need something like ncell=128, nslice=400+, 100k+ particles — probably 10+ minutes runtime. Not practical for CI.

ax3l (approved):

examples/epac2004_benchmarks/analysis_bithermal.py
cemitch99: @ax3l Very nice!  At some point, I would like to explore OMP=2 vs 4 differences a bit more.  It should be possible to further improve runtime by reducing `nslice` once the Strang splitting PR is ready.

ax3l: Thanks!

Oh yes, good point!

PR #1384, opened 2026-04-05
Python: `ref.set_species(species_name)`

ax3l:
Simplify reference particle generation in Python by sharing the same lookup table of charge, mass and gyromagnetic anomaly as the `.in` parameter files logic.

- [x] non-breaking change, everything still works as before if desired
- [x] follow-up PR: update all examples & tests #1385
- [x] show internally used constants in docs ([Python](https://impactx--1384.org.readthedocs.build/en/1384/usage/python.html#impactx.RefPart.set_species), [Inputs file](https://impactx--1384.org.readthedocs.build/en/1384/usage/parameters.html#initial-beam-distributions))
- [x] clarify `Hminus` treatment

### DIFF
--- a/src/initialization/InitDistribution.cpp
+++ b/src/initialization/InitDistribution.cpp
@@ -45,43 +45,10 @@ namespace impactx
         std::string particle_type;  // Particle type
         pp_dist.get("particle", particle_type);
 
-        amrex::ParticleReal qe;     // charge (elementary charge)
-        amrex::ParticleReal massE;  // MeV/c^2
-        constexpr amrex::ParticleReal m_e = ablastr::constant::SI::m_e_v<amrex::ParticleReal>;
-        constexpr amrex::ParticleReal m_p = ablastr::constant::SI::m_p_v<amrex::ParticleReal>;
-        constexpr amrex::ParticleReal MeV_inv_c2 = ablastr::constant::SI::MeV_inv_c2_v<amrex::ParticleReal>;
-        amrex::ParticleReal gyromagnetic_anomaly;  // anomalous magnetic dipole moment
-        if (particle_type == "electron") {
-            qe = -1.0;
-            massE = m_e / MeV_inv_c2;
-            gyromagnetic_anomaly = 0.00115965218062;
-        } else if (particle_type == "positron") {
-            qe = 1.0;
-            massE = m_e / MeV_inv_c2;
-            gyromagnetic_anomaly = 0.00115965218062;
-        } else if (particle_type == "proton") {
-            qe = 1.0;
-            massE = m_p / MeV_inv_c2;
-            gyromagnetic_anomaly = 1.7928473446;
-        } else if (particle_type == "Hminus") {
-            qe = -1.0;
-            massE = 939.294308;  // value used in TraceWin
-            gyromagnetic_anomaly = 1.7928473446; // this value is difficult to find, and needs to be checked
-        }
-        else {  // default to electron
-            ablastr::warn_manager::WMRecordWarning(
-                    "ImpactX::initBeamDistributionFromInputs",
-                    "No beam.particle specified, defaulting to electrons.",
-                    ablastr::warn_manager::WarnPriority::low
-            );
-            qe = -1.0;
-            massE = m_e / MeV_inv_c2;
-            gyromagnetic_anomaly = 0.00115965218062;
-        }
-
         // configure a new reference particle
         RefPart ref;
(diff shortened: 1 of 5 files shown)


src/particles/ReferenceParticle.H
```diff
@@ -74,6 +76,53 @@ namespace impactx
...
+                massE = m_e / MeV_inv_c2;
+                g_anomaly = 0.00115965218062;
+            } else if (species_name == "proton") {
+                qe = 1.0;
+                massE = m_p / MeV_inv_c2;
+                g_anomaly = 1.7928473446;
+            } else if (species_name == "Hminus") {
+                qe = -1.0;
+                massE = 939.294308;
+                g_anomaly = 1.7928473446;
```
ax3l: @cemitch99 we should discuss if we want to use this value from TraceWin for ImpactX (as we encoded before this PR for `.in` parameter files):
```suggestion
                massE = 939.294308;  // value used in TraceWin
                g_anomaly = 1.7928473446;
```

My suggestion: if there is a test that specifically wants to compare 1:1 with tracewin, we can do: `ref.set_species("Hminus").set_mass("939.294308")` and add a comment. Otherwise, let us rely on the PDG, CODATA or other more agreed upon values?

For instance, some tests in ImpactX use for the Hydrogen anion the proton mass of 938.27 MeV (e.g., `run_sol_spin.py`). Let us make those all consistent, too.

src/initialization/InitDistribution.cpp
```diff
@@ -45,43 +45,19 @@ namespace impactx
...
-            qe = 1.0;
-            massE = m_e / MeV_inv_c2;
-            gyromagnetic_anomaly = 0.00115965218062;
-        } else if (particle_type == "proton") {
-            qe = 1.0;
-            massE = m_p / MeV_inv_c2;
-            gyromagnetic_anomaly = 1.7928473446;
-        } else if (particle_type == "Hminus") {
-            qe = -1.0;
-            massE = 939.294308;  // value used in TraceWin
```
ax3l, line 68: I just remembered about this value... See comment in new location.

src/initialization/InitDistribution.cpp
```diff
@@ -45,43 +45,19 @@ namespace impactx
...
-            gyromagnetic_anomaly = 1.7928473446; // this value is difficult to find, and needs to be checked
-        }
-        else {  // default to electron
+        // configure a new reference particle
+        RefPart ref;
+        try {
+            ref.set_species(particle_type);
+        } catch (std::runtime_error const &) {
```
ax3l: Clean up: remove the try-catch, throw on unknown species.

src/particles/ReferenceParticle.H
ax3l: I think the current value of ~939.2941 MeV/c^2 (to a to-be-determined precision) is sensible for proton mass + electron mass + a small fraction binding energy, we might just want to update the comment.

ax3l: **Hydrogen anion H-**
- Simple estimate: 938.272089 proton mass + 0.51099895*2 electron mass + negligible binding energy in the eV range = 939.294 GeV / c^2
- TraceWin: 939.294308 GeV/c^2
- [NIST](https://webbook.nist.gov/cgi/cbook.cgi?ID=C12184882&Mask=1): 1.00849 g/mol -> 939.402 GeV/c^2
- [PubChem 2.2](https://pubchem.ncbi.nlm.nih.gov/compound/166653): 1.0080 g/mol -> 938.946 GeV/c^2
- PDG: ...
- CODATA: ...

src/particles/ReferenceParticle.H
```diff
@@ -74,6 +76,59 @@ namespace impactx
...
+                qe = 1.0;
+                massE = m_e / MeV_inv_c2;
+                g_anomaly = 0.00115965218062;
+            } else if (species_name == "proton") {
+                qe = 1.0;
+                massE = m_p / MeV_inv_c2;
+                g_anomaly = 1.7928473446;
+            } else if (species_name == "Hminus") {
+                qe = -1.0;
+                massE = (m_p + 2.0 * m_e) / MeV_inv_c2;  // neglect ~14 eV/c^2 binding energy
```
ax3l, line 112: This is up to the 6th digit the same value that we used before (from TraceWin). Still neglects the 14 eV because they make no difference for tracking.

cemitch99 (approved):
Looks great.  This change will definitely be appreciated by users who work via Python.

PR #1396, opened 2026-04-14
`LinearMap` Property Check

ax3l:
Add a property to check if the linear map is symplectic.

### DIFF
--- a/src/particles/CovarianceMatrix.H
+++ b/src/particles/CovarianceMatrix.H
@@ -31,6 +31,9 @@ namespace impactx
      * properly called Omega to avoid confusion with complex structures.
      *
      * @see https://en.wikipedia.org/wiki/Symplectic_matrix
+     * Refs.:
+     * - K. R. Meyer and D. C. Offin, Introduction to Hamiltonian Dynamical Systems and the N-Body Problem, Springer-Verlag NY, 2009.
+     * - A. J. Dragt, Lie Methods for Nonlinear Dynamics with Applications to Accelerator Physics, University of Maryland, 2020, http://www.physics.umd.edu/dsat/
      */
     inline Map6x6 symplectic_form ()
     {
--- a/src/particles/CovarianceMatrix.H
+++ b/src/particles/CovarianceMatrix.H
@@ -27,8 +27,8 @@ namespace impactx
      * diag([[0,1],[-1,0]], [[0,1],[-1,0]], [[0,1],[-1,0]])
      * for the phase space ordering (x, px, y, py, t, pt).
      *
-     * Note: commonly referred to as J in accelerator physics, but
-     * properly called Omega to avoid confusion with complex structures.
+     * Note: commonly referred to as J (especially in accelerator physics), but
+     * sometimes called Omega to avoid confusion with complex structures.
      *
      * @see https://en.wikipedia.org/wiki/Symplectic_matrix
      * Refs.:
--- a/src/particles/CovarianceMatrix.H
+++ b/src/particles/CovarianceMatrix.H
@@ -22,6 +22,23 @@ namespace impactx
     /** this is the 6x6 covariance matrix */
     using CovarianceMatrix = Map6x6;
 
+    /** Return the 6x6 skew-symmetric matrix Omega (symplectic form)
+     *
+     * Omega = diag([[0,1],[-1,0]], [[0,1],[-1,0]], [[0,1],[-1,0]])
+     * for the phase space ordering (x, px, y, py, t, pt).
+     *
+     * @see https://en.wikipedia.org/wiki/Symplectic_matrix
+     */
+    inline Map6x6 symplectic_form ()
+    {
+        Map6x6 Omega = Map6x6::Zero();
+        for (int i = 0; i < 3; ++i) {
+            Omega(2*i+1, 2*i+2) =  1;
+            Omega(2*i+2, 2*i+1) = -1;
+        }
+        return Omega;
+    }
+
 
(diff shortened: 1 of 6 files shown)


src/particles/CovarianceMatrix.H
```diff
@@ -22,6 +22,23 @@ namespace impactx
     /** this is the 6x6 covariance matrix */
     using CovarianceMatrix = Map6x6;
 
+    /** Return the 6x6 skew-symmetric matrix Omega (symplectic form)
```
cemitch99: I would have chosen the notation J, since this is more widely used in the accelerator physics literature, and it is used in the book by A. Dragt and the notes by R. Ryne on which most of the element formalism is based.

src/particles/CovarianceMatrix.H
```diff
@@ -22,6 +22,23 @@ namespace impactx
...
+     *
+     * Omega = diag([[0,1],[-1,0]], [[0,1],[-1,0]], [[0,1],[-1,0]])
+     * for the phase space ordering (x, px, y, py, t, pt).
+     *
+     * @see https://en.wikipedia.org/wiki/Symplectic_matrix
```
cemitch99, line 33: There are good references that can be cited here (TBD).

src/particles/CovarianceMatrix.H
ax3l: I know... I had that initially, but then [read on wikipedia](https://en.wikipedia.org/wiki/Symplectic_matrix#The_matrix_%CE%A9):
> Sometimes the notation J is used instead of Ω for the skew-symmetric matrix. This is a particularly unfortunate choice as it leads to confusion with the notion of a [complex structure](https://en.wikipedia.org/wiki/Linear_complex_structure), which often has the same coordinate expression as Ω but represents a very different structure.

So I startled and ran away from J :D

cemitch99: Haha yes.  Accelerator physicists essentially never use the abstract notion of a linear complex structure referenced here, so this is not an issue for us.  I think the documented description is clear enough that the symbol doesn't matter much either way.

ax3l: We can add a comment of the form:
```suggestion
    /** Return the 6x6 skew-symmetric matrix Omega (symplectic form)
     *
     * Note: In most accelerator literature, Omega is referred to as J.
```

Or the other way around. Intentionally, I did call the method `symplectic_form`, so this ends up to become a detail of the narrative/doc string we choose :)

ax3l: > Haha yes. Accelerator physicists essentially never use the abstract notion of a linear complex structure referenced here, so this is not an issue for us. I think the documented description is clear enough that the symbol doesn't matter much either way.

Hehe, ok ;D Let me rename to J and add the Omega note just once somewhere.

ax3l: Updated :)

src/particles/CovarianceMatrix.H
ax3l: Def! Happy to add now or later.

src/particles/CovarianceMatrix.H
```diff
@@ -22,6 +22,26 @@ namespace impactx
     /** this is the 6x6 covariance matrix */
     using CovarianceMatrix = Map6x6;
 
+    /** Return the 6x6 skew-symmetric symplectic form
+     *
+     * diag([[0,1],[-1,0]], [[0,1],[-1,0]], [[0,1],[-1,0]])
+     * for the phase space ordering (x, px, y, py, t, pt).
+     *
+     * Note: commonly referred to as J in accelerator physics, but
+     * properly called Omega to avoid confusion with complex structures.
```
cemitch99:
```suggestion
     * sometimes called Omega to avoid confusion with complex structures.
```

src/particles/CovarianceMatrix.H
```diff
@@ -22,6 +22,26 @@ namespace impactx
     /** this is the 6x6 covariance matrix */
     using CovarianceMatrix = Map6x6;
 
+    /** Return the 6x6 skew-symmetric symplectic form
+     *
+     * diag([[0,1],[-1,0]], [[0,1],[-1,0]], [[0,1],[-1,0]])
+     * for the phase space ordering (x, px, y, py, t, pt).
+     *
+     * Note: commonly referred to as J in accelerator physics, but
```
cemitch99:
```suggestion
     * Note: commonly referred to as J (especially in accelerator physics), but
```

src/particles/CovarianceMatrix.H
```diff
@@ -22,6 +22,26 @@ namespace impactx
...
+     * properly called Omega to avoid confusion with complex structures.
+     *
+     * @see https://en.wikipedia.org/wiki/Symplectic_matrix
```
cemitch99, line 33:
```suggestion
     * @see https://en.wikipedia.org/wiki/Symplectic_matrix
     * Refs.:
     * - K. R. Meyer and D. C. Offin, Introduction to Hamiltonian Dynamical Systems and the N-Body Problem, Springer-Verlag NY, 2009.
     * - A. J. Dragt, Lie Methods for Nonlinear Dynamics with Applications to Accelerator Physics, University of Maryland, 2020, http://www.physics.umd.edu/dsat/
```

ax3l:
Thanks for the refs, @cemitch99 ! All suggestions applied

cemitch99:
Larger issue in CI; I have the same failing tests in PR #1380.

cemitch99 (approved):

PR #1401, opened 2026-04-16
More Element Tests

### LINKED ISSUE #1349: Unit Tests on Elements
Add a method to every element to inverse/reverse the element. Negate `ds` (thick) or negate the _strength_ (thin), then all `compute_constants()`.

## Corner Cases

- `PRot`: swap `phi_in` ↔ `phi_out`
- `DipEdge`: negate `psi / K` coefficients, swap `entry`↔`exit`
- `RFCavity` (thick, energy-changing) needs `-ds` plus negating escale or shifting phase by 180° to reverse the acceleration
- Apertures: nothing for transmitted particles, not inversible for absorbed particles

ax3l:
Test our elements more systematically.

Close #1349

### DIFF
--- a/src/elements/Drift.H
+++ b/src/elements/Drift.H
@@ -74,6 +74,9 @@ namespace impactx::elements
         {
         }
 
+        /** Reverse the element (negate ds) */
+        void reverse () { Thick::reverse(); }
+
         /** Push all particles */
         using BeamOptic::operator();
 
--- a/src/elements/LinearMap.H
+++ b/src/elements/LinearMap.H
@@ -100,8 +100,8 @@ namespace impactx::elements
         /** Reverse the element (compute symplectic inverse of transport map)
          *
          * For a symplectic matrix R, the inverse is
-         *   R^{-1} = Omega^T R^T Omega = -Omega R^T Omega
-         * where Omega is the standard 6x6 symplectic form.
+         *   R^{-1} = J^T R^T J = -J R^T J
+         * where J is the standard 6x6 symplectic form.
          *
          * @see https://en.wikipedia.org/wiki/Symplectic_matrix
          * Throws if the transport map is not symplectic.
--- a/src/elements/LinearMap.H
+++ b/src/elements/LinearMap.H
@@ -182,7 +182,7 @@ namespace impactx::elements
         AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
         void operator() (RefPart & AMREX_RESTRICT refpart) const
         {
-            if (m_ds > 0)  // Drift
+            if (m_ds != 0)  // Drift forward or backward
             {
                 using namespace amrex::literals; // for _rt and _prt
                 using amrex::Math::powi;
--- a/src/elements/LinearMap.H
+++ b/src/elements/LinearMap.H
@@ -97,6 +97,31 @@ namespace impactx::elements
             return max_err <= tol;
         }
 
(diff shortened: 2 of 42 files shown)


src/elements/Drift.H
```diff
@@ -74,6 +74,9 @@ namespace impactx::elements
         {
         }
 
+        /** Invert the element (negate ds) */
+        void invert () { Thick::invert(); }
```
ax3l: I am not yet sure if I like this as in-place method or in practice returning a fresh, inverted copy is better. I tend to change it to the latter for practical use cases. But then, we can always implement `.copy()` too and do `.copy().invert()`

src/elements/RFCavity.H
```diff
@@ -363,7 +366,7 @@ namespace impactx::elements
 
             // specify constants
             using ablastr::constant::math::pi;
-            amrex::ParticleReal const zlen = m_ds;
+            amrex::ParticleReal const zlen = std::abs(m_ds);
```
ax3l, line 373: This is the fix I pushed into #1397

src/elements/SoftQuad.H
```diff
@@ -362,7 +365,7 @@ namespace impactx::elements
 
             // specify constants
             using ablastr::constant::math::pi;
-            amrex::ParticleReal const zlen = m_ds;
+            amrex::ParticleReal const zlen = std::abs(m_ds);
```
ax3l, line 372: See #1397

src/elements/SoftSol.H
```diff
@@ -371,7 +374,7 @@ namespace impactx::elements
 
             // specify constants
             using ablastr::constant::math::pi;
-            amrex::ParticleReal const zlen = m_ds;
+            amrex::ParticleReal const zlen = std::abs(m_ds);
```
ax3l, line 381: See #1397

tests/python/test_reversibility_elements.py
```diff
@@ -0,0 +1,476 @@
...
+    cols = PHASE_COLS + (SPIN_COLS if spin else [])
+    any_changed = False
+    for c in cols:
+        if not np.allclose(df[c].to_numpy(), initial[c], atol=1e-14, rtol=0):
+            any_changed = True
+            break
+    assert any_changed, "Forward push did not change any particle coordinate"
+
+
+def check_roundtrip(pc, initial, atol=1e-10, spin=False):
```
ax3l: We gotta check the expected tolerance we want (absolute/relative). The current atol is just a draft and seems to need a larger value for w/ spin. But not sure if atol w/o an rtol is sensible at all here.

src/elements/LinearMap.H
```diff
@@ -97,6 +97,31 @@ namespace impactx::elements
             return max_err <= tol;
         }
 
+        /** Invert the element (compute symplectic inverse of transport map)
+         *
+         * For a symplectic matrix R, the inverse is
+         *   R^{-1} = Omega^T R^T Omega = -Omega R^T Omega
+         * where Omega is the standard 6x6 symplectic form.
```
ax3l: Renamed in the other PR to:
```suggestion
         *   R^{-1} = J^T R^T J = -J R^T J
         * where J is the standard 6x6 symplectic form.
```

cemitch99 (approved):
This looks great to me, provided tests pass.  I will try to understand the SpinMap case offline.  It's ok with me to merge this one and continue ongoing testing of special cases.

PR #1409, opened 2026-04-19
[WIP] Linear Transport: Centroid Tracking

ax3l:
Add a centroid 6D phase vector to our envelope tracking logic.

The goal is to support off-axis envelope propagation and feed-down effects in linear optics from misalignment (translation & rotation) and off-axis reference particle.

This lets transfer maps, and currently developed map traces and Twiss optics see the same feed-down effects that particles experience through the element.

Reusable helper methods are added to the `mixin::Alignment` class that we can use in other elements as well.

## Multipole (Illustrative Example)

Higher-order thin multipoles contribute the correct linear focusing when the reference orbit is off axis or the element is misaligned, including the normal/skew mixing from transverse roll.

Adds regression coverage for sextupole feed-down (e.g. Quads kicks) and periodic lattice optics.

### DIFF
--- a/src/elements/Multipole.H
+++ b/src/elements/Multipole.H
@@ -32,7 +32,7 @@ namespace impactx::elements
     struct Multipole
     : public mixin::Named,
       public mixin::BeamOptic<Multipole>,
-      public mixin::LinearTransport<Multipole, false>,
+      public mixin::LinearTransport<Multipole>,
       public mixin::Thin,
       public mixin::Alignment,
       public mixin::NoFinalize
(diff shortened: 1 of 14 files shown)


src/elements/Multipole.H
```diff
@@ -195,15 +195,52 @@ namespace impactx::elements
...
+                Complex const alpha(m_Kn, m_Ks);
+
+                amrex::ParticleReal const inv_factorial_nm1 =
+                    static_cast<amrex::ParticleReal>(n) /
+                    static_cast<amrex::ParticleReal>(m_mfactorial);
+
+                // Linearize the kick around the incoming reference orbit in
+                // the element frame. For n == 1 this reproduces the standard
+                // thin quadrupole matrix, while n >= 2 gives feed-down.
+                Complex const beta = amrex::pow(zeta, n - 1) * alpha * inv_factorial_nm1;
```
ax3l, line 239: Need to double check `pow(0, 0)` [behavior](https://en.wikipedia.org/wiki/Zero_to_the_power_of_zero) in AMReX (quadrupole with on-axis ref particle)....

ax3l: In IEEE 754, the function `std::pow(base, ±0)` is required to return 1 for any base, including zero or NaN. But in [ISO C++](https://en.cppreference.com/cpp/numeric/math/pow) it is actually kept free for the implementation. So most compilers will implement 1 here, but maybe we want to make it explicit/safe...

ax3l: Ok, but since we use `amrex::pow`: AMReX's integer-exponent GpuComplex uses binary exponentiation, which returns 1+0i for exponent 0, so the quadrupole (n=1) on-axis branch is well-defined. We actually also cover this in the tests added to this PR. All good
https://github.com/AMReX-Codes/amrex/blob/9219ba416b7ba2073dd1b12bf19fdce27391f17b/Src/Base/AMReX_GpuComplex.H#L440

ax3l:
While implementing the feed-down logic, I think I stumbled on a convention-caused issue that we might want to check in ImpactX. Correct me if I got something wrong in my assumptions.

1. We want to use one (ideal?) reference particle update implementation for all three tracking modes, otherwise this gets very confusing.
2. By convention so far, we model all element alignment/rotation errors on the beam/map, not modifying the reference particle.

For the current implementation in this PR regarding feed-down effects in the linear maps from misalignment (something that MAD-X and Xsuite and others can model), we thus lack the capability to model ideal dipole kicks from an off-axis higher-order multipoles (or anything that depends on propagating the constant orbit offset): The current 6x6 map only carries linear transport maps (Jacobians), not orbit offsets.

  - we **do** capture the linear focusing/coupling effect around the current reference orbit
  - we **do not** propagate the orbit offset caused by the constant kick to later elements

With this PR we can model:

- Local feed-down at the current reference orbit.
- Normal/skew mixing from roll.
- Misalignment/rotation effects in the element’s own linearized map.
- Composition of those local linear maps through the lattice.

There are probably a few ways forward, but we need to discuss them:

a) accept & document the limitations of what feed-down effects we can model
b) flip the alignment & rotation generally in the reference particle: possible but not always what people want when they search for an ideal reference orbit -- also the most intrusive change
c) evolve the reference particle (using the ideal kicks) for envelope tracking

PR #1410, opened 2026-04-19
Linear Transport: Calc `transfer_map` Once

ax3l:
Just in case a compiler optimization misses this:
We want to calculate `R` exactly once and then transpose it. Before, this was potentially calculated twice.


ax3l:
This is a trivial refactor and tests pass. I will merge it because I want to change `LinearTransport::operator()` more in #1409 for feed-down logic.

PR #1411, opened 2026-04-19
Fix `transfer_map` for edge- and slice-dependent linear optics

ax3l:
`transfer_map` now follows the reference orbit with the correct element entrance `s` and per-slice reference update, so the accumulated linear map matches the actual single-particle optics through the lattice.

This fixes wrong end-to-end maps for elements whose linear response depends on the evolving reference state, especially RF cavities and soft field elements such as SoftQuadrupole and SoftSolenoid.

Also add an element-by-element map trace, so the build-up of focusing and coupling can be inspected along `s` instead of only at the lattice exit. Both implementations share the same logic / do not repeat themselves. We will use the map trace in follow-up PRs that implement twiss optics/diagnostics.

### DIFF
--- a/src/diagnostics/LinearMap.cpp
+++ b/src/diagnostics/LinearMap.cpp
@@ -1,0 +1,248 @@
+/* Copyright 2026 The Regents of the University of California, through Lawrence
+ *           Berkeley National Laboratory (subject to receipt of any required
+ *           approvals from the U.S. Dept. of Energy). All rights reserved.
+ *
+ * This file is part of ImpactX.
+ *
+ * Authors: Axel Huebl
+ * License: BSD-3-Clause-LBNL
+ */
+#include "LinearMap.H"
+
+#include <ablastr/warn_manager/WarnManager.H>
+
+#include <AMReX_BLProfiler.H>
+#include <AMReX_REAL.H>
+
+#include <set>
+#include <stdexcept>
+#include <string>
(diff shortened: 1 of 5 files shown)


src/diagnostics/LinearMap.cpp
```diff
@@ -0,0 +1,248 @@
...
+            ref.sedge = ref.s;
+
+            std::visit([&](auto && element)
+            {
+                using E = std::decay_t<decltype(element)>;
+
+                int const nslice = element.nslice();
+                for (int slice_i = 0; slice_i < nslice; ++slice_i)
+                {
+                    element(ref);
```
ax3l, line 178: Here are the two bug fixes: the old logic in `src/python/lattice.cpp` (see below) forgot to update `ref.sedge` on element entry and it also only pushed the reference particle once per element, which kicks it only 1/n_slice updates far... 

So yay, I think in total I managed to code 3 bugs in #1256 in January (the first one already fixed in #1338). Admittedly, we wrote very poor tests so far for all the linear optics, so all the affected elements were not tested. This will improve with the new upcoming twiss diagnostics, that will exercise all this.

Sorry :D

ax3l:
This is a relatively trivial bug fix and tests pass. I need this for more complex work for better feed-down book-keeping in #1409 and dependent PRs, so I will merge it. Prior tests continue to pass.

PR #1412, opened 2026-04-19
[WIP] Linear-optics Twiss functions

ax3l:
Two modes are supported:

* **Periodic ring** (``init=None``): the one-turn map is computed from the current lattice, the Wolski 6D eigendecomposition provides the matched initial Courant-Snyder / Mais-Ripken Twiss, and the closed-orbit dispersion is obtained from the periodic equation $$(I - M_{4 \times 4}) D = M[0{:}4, 5]$$. Fractional tunes are extracted from the eigenvalue phases and returned in the ``tunes`` key.
* **Transfer line / linac** (``init`` given): the user supplies initial uncoupled Courant-Snyder parameters; Twiss is then propagated from the entrance through the lattice without matching.

### DIFF
--- a/src/python/impactx/__init__.py
+++ b/src/python/impactx/__init__.py
@@ -28,6 +28,7 @@ __author__ = impactx_pybind.__author__
 
 from .distribution_input_helpers import twiss  # noqa
 from .fourier import fourier_coefficients  # noqa
+from .twiss_lattice import compute_twiss as _compute_twiss_lattice
 from .extensions.KnownElementsList import (
     FilteredElementsList,
     register_KnownElementsList_extension,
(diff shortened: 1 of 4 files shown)


src/python/impactx/__init__.py
```diff
@@ -65,3 +66,23 @@
...
+#                                                of the one-turn map
+#   sim.twiss(init={"beta_x": ..., ...})      -> transfer line / linac: propagate
+#                                                uncoupled Courant-Snyder Twiss
+#                                                from the given initial conditions
+#
+# Returns a dict of numpy arrays (one entry per element boundary) with beta,
+# alpha, phase advance mu, and dispersion along s, plus the fractional tunes
+# for rings. The underlying lattice-level building blocks are available as
+# ``sim.lattice.transfer_map(ref)`` (end-to-end map) and
+# ``sim.lattice.map_trace(ref)`` (per-element cumulative trace).
```
ax3l, line 87: This is useful doc, but a bit ill placed here. Belongs in `compute_twiss` or other user-facing docs.

src/python/impactx/twiss_lattice.py
```diff
@@ -0,0 +1,821 @@
...
+    J is block-diagonal with three 2x2 blocks ``[[0, 1], [-1, 0]]``. A
+    matrix ``A`` is symplectic iff ``A^T J A = J``; the Poisson bracket
+    of two phase-space functions f, g satisfies
+    ``{f, g} = (grad f)^T J (grad g)``.
+    """
+    J = np.zeros((6, 6))
+    for i in range(3):
+        J[2 * i, 2 * i + 1] = 1.0
+        J[2 * i + 1, 2 * i] = -1.0
+    return J
```
ax3l, line 140: We have a value for this now in `CovarianceMatrix.H` and `SmallMatrix.H` that we can bind.

PR #1419, opened 2026-04-20
pybind11: 3.0.4

EZoni:
## Related PRs

- https://github.com/AMReX-Codes/pyamrex/pull/562
- https://github.com/BLAST-WarpX/warpx/pull/6792

### DIFF
--- a/cmake/dependencies/pybind11.cmake
+++ b/cmake/dependencies/pybind11.cmake
@@ -37,7 +37,7 @@ function(find_pybind11)
             mark_as_advanced(FETCHCONTENT_UPDATES_DISCONNECTED_FETCHEDpybind11)
         endif()
     else()
-        find_package(pybind11 3.0.3 CONFIG REQUIRED)
+        find_package(pybind11 3.0.4 CONFIG REQUIRED)
         message(STATUS "pybind11: Found version '${pybind11_VERSION}'")
     endif()
 endfunction()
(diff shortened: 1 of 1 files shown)


ax3l (approved):
LGTM, ty!

cmake/dependencies/pybind11.cmake
```diff
@@ -37,7 +37,7 @@ function(find_pybind11)
             mark_as_advanced(FETCHCONTENT_UPDATES_DISCONNECTED_FETCHEDpybind11)
         endif()
     else()
-        find_package(pybind11 3.0.3 CONFIG REQUIRED)
+        find_package(pybind11 3.0.4 CONFIG REQUIRED)
```
ax3l, line 40: @EZoni I overlooked something here: You do not want to update the version in `find_package` higher than needed, because that way we cannot build with version 3.0.0, 3.0.1, 3.0.2, and 3.03 in package managers anymore - even if those versions are totally fine. (e.g., conda-forge is still to ship 3.0.4, stands at 3.0.3. Spack has very good version control, too.)

So in other words: keep this here on 3.0.0 and below just update the version we download to the latest.

Applies also to all other dependencies in our build logic.

EZoni: Okay, understood. I can open a PR to fix that.

EZoni: Fix implemented in https://github.com/BLAST-ImpactX/impactx/pull/1432.

PR #1425, opened 2026-04-21
Linear Transport: Programmable

ax3l:
Enable to set the linear transfer map of the programmable element, if desired. Simply accepts a ref-particle dependent function generator for a 6x6 matrix. This is a bit more general than `SpinMap` element, which has a static map that does not depend on `ref`.

### DIFF
--- a/src/elements/Programmable.H
+++ b/src/elements/Programmable.H
@@ -88,17 +88,34 @@ namespace impactx::elements
 
         /** This function returns the linear transport map.
          *
+         * By default, @c Programmable provides no linear map because its
+         * particle push is an arbitrary user callback. Users who know the
+         * Jacobian of their push around the reference trajectory can opt
+         * in by setting the @c m_transport_map hook: a callable that
+         * returns a 6x6 transport matrix as a function of the current
+         * reference particle. When set, this hook is used by envelope
+         * tracking, @c lattice.transfer_map, @c map_trace, and Twiss.
+         *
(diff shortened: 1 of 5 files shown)


src/elements/Programmable.H
```diff
@@ -87,18 +87,35 @@ namespace impactx::elements
...
+        transport_map (RefPart const & AMREX_RESTRICT refpart) const
         {
-            // TODO: can/should we fuse the two elements?
-            throw std::runtime_error("The Programmable element cannot be used for linear transport maps. Use the LinearMap element instead.");
-
-            return Map6x6::Identity();
+            if (m_transport_map == nullptr) {
+                throw std::runtime_error(
+                    "Programmable: no linear transport map hook was set. "
+                    "Set Programmable.transport_map_hook = <callable(refpart)->6x6> "
```
ax3l, line 113:
```suggestion
                    "Set Programmable.linear_map = <callable(refpart)->6x6> "
```

cemitch99 (commented):
One previous suggestion awaits to be applied.

src/elements/Programmable.H
```diff
@@ -87,18 +87,35 @@ namespace impactx::elements
...
+                    "for envelope/linear-optics tracking, or use the LinearMap "
+                    "element for a purely linear map."
+                );
+            }
+            return m_transport_map(refpart);
```
ax3l, line 118: Needs #1506 applied, too - or is this never `mixin::Aligned`?

PR #1426, opened 2026-04-21
Linear Transport: PRot

ax3l:
Add an ideal `transport_map` for `PRot`.

### DIFF
--- a/src/elements/PRot.H
+++ b/src/elements/PRot.H
@@ -188,7 +188,8 @@ namespace impactx::elements
          *
          * Reference: MaryLie 3.0 Users' Manual, Sect. 6.5 ("Plane Rotation");
          * A. J. Dragt, Lie Methods for Nonlinear Dynamics with Applications
-         * to Accelerator Physics, Appendix V ("PROT without and in the Presence of a Magnetic Field").
+         * to Accelerator Physics, v. Nov. 2020,
+         * Appendix V ("PROT without and in the Presence of a Magnetic Field").
          *
          * Ideal behavior only: alignment and rotation errors are
          * ignored; the linearization is at (x, y, t) = (0, 0, 0) and
--- a/src/elements/PRot.H
+++ b/src/elements/PRot.H
@@ -33,7 +33,7 @@ namespace impactx::elements
     struct PRot
     : public mixin::Named,
       public mixin::BeamOptic<PRot>,
-      public mixin::LinearTransport<PRot, false>,
+      public mixin::LinearTransport<PRot>,
       public mixin::Thin,
       public mixin::NoFinalize,
       public amrex::simd::Vectorized<amrex::simd::native_simd_size_particlereal>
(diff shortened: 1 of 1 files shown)


src/elements/PRot.H
```diff
@@ -168,15 +168,65 @@ namespace impactx::elements
...
+         * Symplecticity of the reduced (x, px, t, pt) block was
+         * verified with SymPy (R^T J R = J).
+         *
+         * The (y, py) sub-block is the identity: pole-face rotation in
+         * the x-z plane does not mix the vertical plane at linear order.
+         *
+         * Reference: MaryLie 3.0 Users' Manual, Sect. 6.4 ("General
+         * Bends, Geometric Corrections"); A. J. Dragt, Lie Methods for
+         * Nonlinear Dynamics with Applications to Accelerator Physics,
+         * Sect. 12.5 (exact bend geometry / pole-face rotations).
```
ax3l, line 192: This needs to be checked.

src/elements/PRot.H
```diff
@@ -168,15 +168,65 @@ namespace impactx::elements
...
+            (void) sin_phi_in;
+            auto const [sin_phi_out, cos_phi_out] = amrex::Math::sincos(m_phi_out);
+            (void) sin_phi_out;
+
+            Map6x6 R = Map6x6::Identity();
+            R(1,1) = cos_phi_in / cos_phi_out;
+            R(2,2) = cos_phi_out / cos_phi_in;
+            R(2,6) = -sin_theta / (beta * cos_phi_in);
+            R(5,1) = +sin_theta / (beta * cos_phi_out);
+
```
cemitch99, line 223: Validated against the linear map obtained in a Mathematica notebook that was during the original implementation of the full map.  This coincides with the linear map appearing in Appendix V of Alex's book (Nov. 2020 version).

src/elements/PRot.H
```diff
@@ -168,15 +168,65 @@ namespace impactx::elements
...
+         *
+         * Reference: MaryLie 3.0 Users' Manual, Sect. 6.4 ("General
+         * Bends, Geometric Corrections"); A. J. Dragt, Lie Methods for
+         * Nonlinear Dynamics with Applications to Accelerator Physics,
+         * Sect. 12.5 (exact bend geometry / pole-face rotations).
+         *
+         * Sign convention: with ImpactX's pt = -Delta(gamma)/(beta0
+         * gamma0) (opposite sign to MAD-X's PT = +Delta(E)/(p0*c)), the
+         * R(2,6) and R(5,1) signs agree with the sibling Sbend element
+         * in ImpactX but differ overall from the MAD-X PROT formulas.
```
cemitch99: I'm not sure we need to comment on the PT sign convention of MAD-X here, since this is not specific to this element, but applies more generally and dates back to the origins of MaryLie and MAD.

src/elements/PRot.H
cemitch99: Yes, the sources are right, but those are not the right sections (or refer to outdated versions?).  The first reference should be to Sect. 6.5 ("Plane Rotation").  The second reference should be Appendix V ("PROT without and in the Presence of a Magnetic Field").

cemitch99 (commented):
The map is correct; just some comments on the comments.

src/elements/PRot.H
ax3l: Yes I am not sure if outdated or slightly halluscinated on those. I also looked up a Wiedemann ref the other day that it found and was wondering as well if it was from a new revision or guessed it vaguely.

Generally, referencing section titles rather than numbers is more stable I find, for books & manuals with multiple editions.

src/elements/PRot.H
ax3l: Agreed, user-facing differences are probably more important than the implementation detail here.

ax3l:
Thank you for checking! I left comments on your comments on comments, which concur. Will update the comments.

src/elements/PRot.H
ax3l:
```suggestion
         * Reference: MaryLie 3.0 Users' Manual, Sect. 6.5 ("Plane Rotation");
         * A. J. Dragt, Lie Methods for Nonlinear Dynamics with Applications
         * to Accelerator Physics, Appendix V ("PROT without and in the Presence of a Magnetic Field").
```

src/elements/PRot.H
ax3l:
```suggestion
```

src/elements/PRot.H
```diff
@@ -168,15 +168,65 @@ namespace impactx::elements
...
+         *
```
ax3l, line 193:
```suggestion
```

src/elements/PRot.H
```diff
@@ -168,15 +168,59 @@ namespace impactx::elements
...
+         * which simplifies the x-denominator pz|_0 -> cos(phi_out).
+         * Symplecticity of the reduced (x, px, t, pt) block was
+         * verified with SymPy (R^T J R = J).
+         *
+         * The (y, py) sub-block is the identity: pole-face rotation in
+         * the x-z plane does not mix the vertical plane at linear order.
+         *
+         * Reference: MaryLie 3.0 Users' Manual, Sect. 6.5 ("Plane Rotation");
+         * A. J. Dragt, Lie Methods for Nonlinear Dynamics with Applications
+         * to Accelerator Physics, Appendix V ("PROT without and in the Presence of a Magnetic Field").
```
cemitch99:
```suggestion
         * to Accelerator Physics, v. Nov. 2020,
         * Appendix V ("PROT without and in the Presence of a Magnetic Field").
```

cemitch99 (approved):

PR #1428, opened 2026-04-21
Doc: AI Skill - Update Element Benchmarks

ax3l:
This skill should be used when the user asks to "update element benchmarks", "add missing element tests", "audit element test coverage", "add a benchmark for <element>", or to check that every lattice element under `src/elements/` has matching entries in the three Python element-coverage tests (benchmark, serialization, reversibility) and that spin-capable elements carry the spin/nospin parametrization. Also covers re-running those three tests after edits.

Tested for developing #1429, works well.

Issue #1328, opened 2026-02-27
Treat translation/rotation errors with spin


cemitch99:
Since the spin push has been separated into a separate function per element, we need to be sure that the shift_in and shift_out functions are applied at the correct location to capture the effect of errors on the spin push.

ax3l:
In case this helps for testing: we can use `tests/python/test_reversibility_elements.py` now.

I noticed there that the `Quad`, `Sbend`, `Sol` and `ChrQuad` elements seem to break or had a higher tolerance when they use off-center alignment with spin-tracking.
The relevant patch to exclude them from tests so far reads as `kwargs = {} if sim.spin else PIPE_KWARGS` in those 4 elements.

ax3l:
Thought about it and I think I know how to solve this cleanly:
We can move the shift in/out to the BeamOptics logic, so it brackets the phase space or spin+phase space loops correctly.

Will submit a PR: #1455

PR #1429, opened 2026-04-21
Test: Element Coverage Update

ax3l:
Add new elements and new spin support to our PyTest benchmarks. Co-prepared with #1428

## Added benchmarks (`tests/python/test_benchmark_elements.py`)

- `test_ExactCFbend` (params from `examples/symplectic_integration/run_exact_cfbend.py`)
- `test_PolygonAperture` (params from `examples/polygon_aperture/run_polygon_aperture.py`)

## Added spin/nospin parametrize

New elements that inherit `mixin::SpinTransport` now exercise both paths:

- benchmark: `ExactQuad`
- reversibility: `ExactQuad`, `LinearMap`

Both round-trip cleanly at machine precision under spin tracking.

## CFbend: spin reversibility known-broken (#1328)

`CFbend` also inherits `mixin::SpinTransport`, but `forward + reverse + forward` leaves O(1e-5) residual in the spin components even with weak k/rc and no aperture (phase space still round-trips to 3e-20). The C++ `spin_and_phasespace_push` applies the spin rotation on pre-push coordinates, so the reverse does not compose to identity on spin.

Left `test_CFbend` as nospin-only with a comment pointing to #1328 so this is visible in the next fix pass. No tolerance was loosened.


tests/python/test_benchmark_elements.py
```diff
@@ -234,6 +249,7 @@ def test_Kicker(benchmark, sim):
     benchmark.pedantic(el.push, setup=partial(pc_setup, sim), rounds=rounds)
 
 
+@pytest.mark.parametrize("sim", [True, False], indirect=True, ids=["spin", "nospin"])
```
ax3l: Should not use spin.

```suggestion
# Note: has no spin support; see SpinMap
```

cemitch99:
@ax3l Is there still an issue regarding CFbend for this PR?  A few conflicts need to be resolved.

cemitch99 (approved):

Issue #745, opened 2024-10-17
Clarify the effect of errors on the reference trajectory


cemitch99:
In elements with alignment shifts or rotations, the alignment shifts and rotations do not affect the push of the reference particle.  (These are treated as errors, while the reference orbit is treated as nominal).

It would be good to add a flag that allows the user to specify whether the shift/rotation provided for an element is to be treated as an error (no modification to the reference orbit) or as a design feature (with modification of the reference orbit).

PR #1431, opened 2026-04-22
Add the option to apply DipEdge to the reference particle.

### LINKED ISSUE #1395: Allow reference trajectory shift in DipEdge
The nonlinear model used in the element `DipEdge` includes a nonzero horizontal translation (depending on the field integral $K_0$) that is present even for a particle that begins on the ideal "hard-edge" reference trajectory.  For a beam, this will result in a centroid offset that will produce centroid oscillations in the downstream beamline.  In practice, this can be avoided by aligning the downstream elements with the true horizontal position (after including the effect of the fringe field).

To model this correction, it would be good to allow two options in `DipEdge`:

1) the existing model, in which the shift is applied to the beam particles but not to the reference particle -- this model makes sense if the shift due to the fringe field is not considered in the baseline design, so that downstream elements are aligned with the "idealized" reference trajectory

2) a model in which the shift is applied to the reference particle, but not to the beam particles -- this model makes sense if the shift due to the fringe field is considered as part of the baseline design, so that downstream elements are aligned with the "shifted" reference trajectory

This option is similar to what was done in [PR#1167](https://github.com/BLAST-ImpactX/impactx/pull/1167).

cemitch99:
This PR closes #1395 

- [x] implement modification to the beam particle push
- [x] implement modification to the reference particle push (validate)
- [x] add benchmark example
- [x] update documentation

### DIFF
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -381,14 +381,14 @@ By comparison, note that the MAD-X DIPEDGE element uses as input the half-gap ``
 
 Note that the nonlinear model includes a nonzero horizontal translation (depending on the field integral values) that is present even for a particle that begins on the ideal "hard-edge" reference trajectory.
 
-For a beam, this will result in a centroid offset that will produce centroid oscillations in the downstream beamline. In practice, this can be avoided by aligning the downstream elements with the true 
+For a beam, this will result in a centroid offset that will produce centroid oscillations in the downstream beamline. In practice, this can be avoided by aligning the downstream elements with the true
 
 horizontal position (after including the effect of the fringe field).  To model this correction, we allow two options in the dipedge model:
 
-* the default model, in which the shift due to the fringe field is applied to each beam particle phase space vector but not to the reference particle phase space vector -- 
+* the default model, in which the shift due to the fringe field is applied to each beam particle phase space vector but not to the reference particle phase space vector --
 this model makes sense if the shift due to the fringe field is not considered in the baseline design, so that downstream elements are aligned with the "idealized" reference trajectory
 
-* an option in which the shift due to the fringe field is applied to the reference particle phase space vector, but not to the beam particle phase space vector -- 
+* an option in which the shift due to the fringe field is applied to the reference particle phase space vector, but not to the beam particle phase space vector --
 this model makes sense if the shift due to the fringe field is considered as part of the baseline design, so that downstream elements are aligned with the "shifted" reference trajectory
 
 This element requires these additional parameters:
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -379,7 +379,19 @@ when expanded to first order in ``g/rc`` (gap / radius of curvature).
 
 By comparison, note that the MAD-X DIPEDGE element uses as input the half-gap ``HGAP = g/2``, and sets the default value ``FINT = 0`` (while the corresponding default value of ``K2`` is set to 1).
 
-This requires these additional parameters:
+Note that the nonlinear model includes a nonzero horizontal translation (depending on the field integral values) that is present even for a particle that begins on the ideal "hard-edge" reference trajectory.
+
+For a beam, this will result in a centroid offset that will produce centroid oscillations in the downstream beamline. In practice, this can be avoided by aligning the downstream elements with the true 
+
+horizontal position (after including the effect of the fringe field).  To model this correction, we allow two options in the dipedge model:
+
+* the default model, in which the shift due to the fringe field is applied to each beam particle phase space vector but not to the reference particle phase space vector -- 
+this model makes sense if the shift due to the fringe field is not considered in the baseline design, so that downstream elements are aligned with the "idealized" reference trajectory
+
+* an option in which the shift due to the fringe field is applied to the reference particle phase space vector, but not to the beam particle phase space vector -- 
+this model makes sense if the shift due to the fringe field is considered as part of the baseline design, so that downstream elements are aligned with the "shifted" reference trajectory
+
+This element requires these additional parameters:
 
 * ``<element_name>.psi`` (``float``, in radians) the pole face rotation angle
 * ``<element_name>.rc`` (``float``, in meters) the bend radius
@@ -394,6 +406,7 @@ This requires these additional parameters:
 * ``<element_name>.K6`` (``float``, dimensionless) normalized field integral for fringe field (default: ``0``)
 * ``<element_name>.model`` (``string``) the fringe field model: ``linear`` (default) or ``nonlinear``
 * ``<element_name>.location`` (``string``) the fringe field edge location: ``entry`` (default) or ``exit``
+* ``<element_name>.modify_ref_part`` (``boolean``) apply fringe field to the reference particle ``true`` or ``false`` (default)
 * ``<element_name>.dx`` (``float``, in meters) horizontal translation error
 * ``<element_name>.dy`` (``float``, in meters) vertical translation error
 * ``<element_name>.rotation`` (``float``, in degrees) rotation error in the transverse plane
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -381,14 +381,14 @@ By comparison, note that the MAD-X DIPEDGE element uses as input the half-gap ``
 
 Note that the nonlinear model includes a nonzero horizontal translation (depending on the field integral values) that is present even for a particle that begins on the ideal "hard-edge" reference trajectory.
 
-For a beam, this will result in a centroid offset that will produce centroid oscillations in the downstream beamline. In practice, this can be avoided by aligning the downstream elements with the true
+For a beam, this will result in a centroid offset that will produce centroid oscillations in the  downstream beamline.
+In practice, this can be avoided by aligning the downstream elements with the true horizontal position (after including the effect of the fringe field).
+To model this correction, we allow two options in the dipedge model:
 
-horizontal position (after including the effect of the fringe field).  To model this correction, we allow two options in the dipedge model:
-
-* the default model, in which the shift due to the fringe field is applied to each beam particle phase space vector but not to the reference particle phase space vector --
+* the option ``modify_ref_part = False`` (default), in which the shift due to the fringe field is applied to each beam particle phase space vector but not to the reference particle phase space vector --
 this model makes sense if the shift due to the fringe field is not considered in the baseline design, so that downstream elements are aligned with the "idealized" reference trajectory
 
-* an option in which the shift due to the fringe field is applied to the reference particle phase space vector, but not to the beam particle phase space vector --
+* the option ``modify_ref_part = True`` in which the shift due to the fringe field is applied to the reference particle phase space vector, but not to the beam particle phase space vector --
 this model makes sense if the shift due to the fringe field is considered as part of the baseline design, so that downstream elements are aligned with the "shifted" reference trajectory
 
 This element requires these additional parameters:
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -1195,7 +1195,7 @@ This module provides elements and methods for the accelerator lattice.
 
       focusing t strength in 1/m
 
-.. py:class:: impactx.elements.DipEdge(psi, rc, g, R=1, K0=pi**2/6, K1=0, K2=1, K3=1/6, K4=0, K5=0, K6=0, model="linear", location="entry", dx=0, dy=0, rotation=0, name=None)
+.. py:class:: impactx.elements.DipEdge(psi, rc, g, R=1, K0=pi**2/6, K1=0, K2=1, K3=1/6, K4=0, K5=0, K6=0, model="linear", location="entry", modify_ref_part=False, dx=0, dy=0, rotation=0, name=None)
 
    Edge focusing associated with bend entry or exit
 
@@ -1221,6 +1221,16 @@ This module provides elements and methods for the accelerator lattice.
    By comparison, note that the MAD-X DIPEDGE element uses as input the half-gap ``HGAP = g/2``, and sets the default value ``FINT = 0`` (while
    the corresponding default value of ``K2`` is set to 1).
 
+   Note that the nonlinear model includes a nonzero horizontal translation (depending on the field integral values) that is present even for a particle that begins on the ideal "hard-edge" reference 
+   trajectory.  For a beam, this will result in a centroid offset that will produce centroid oscillations in the downstream beamline. In practice, this can be avoided by aligning the downstream elements with
+   the true horizontal position (after including the effect of the fringe field).  To model this correction, we allow two options in the dipedge model:
+
+   * the default model, in which the shift due to the fringe field is applied to each beam particle phase space vector but not to the reference particle phase space vector --
+   this model makes sense if the shift due to the fringe field is not considered in the baseline design, so that downstream elements are aligned with the "idealized" reference trajectory
+
+   * an option in which the shift due to the fringe field is applied to the reference particle phase space vector, but not to the beam particle phase space vector --
+   this model makes sense if the shift due to the fringe field is considered as part of the baseline design, so that downstream elements are aligned with the "shifted" reference trajectory
+
    :param psi: Pole face angle [radians]
    :param rc: Radius of curvature [m]
    :param g: Gap parameter [m]
@@ -1234,6 +1244,7 @@ This module provides elements and methods for the accelerator lattice.
    :param K6: Fringe field integral [unitless]
    :param model: the fringe field model: ``linear`` (default) or ``nonlinear``
    :param location: the fringe field edge location: ``entry`` (default) or ``exit``
+   :param modify_ref_part: apply fringe field to the reference particle ``True`` or ``False`` (default)
    :param dx: horizontal translation error [m]
    :param dy: vertical translation error [m]
    :param rotation: rotation error in the transverse plane [degrees]
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -1225,10 +1225,10 @@ This module provides elements and methods for the accelerator lattice.
    trajectory.  For a beam, this will result in a centroid offset that will produce centroid oscillations in the downstream beamline. In practice, this can be avoided by aligning the downstream elements with
    the true horizontal position (after including the effect of the fringe field).  To model this correction, we allow two options in the dipedge model:
 
-   * the default model, in which the shift due to the fringe field is applied to each beam particle phase space vector but not to the reference particle phase space vector --
+   * the option ``modify_ref_part = False`` (default), in which the shift due to the fringe field is applied to each beam particle phase space vector but not to the reference particle phase space vector --
    this model makes sense if the shift due to the fringe field is not considered in the baseline design, so that downstream elements are aligned with the "idealized" reference trajectory
 
-   * an option in which the shift due to the fringe field is applied to the reference particle phase space vector, but not to the beam particle phase space vector --
+   * the option ``modify_ref_part = True``, in which the shift due to the fringe field is applied to the reference particle phase space vector, but not to the beam particle phase space vector --
    this model makes sense if the shift due to the fringe field is considered as part of the baseline design, so that downstream elements are aligned with the "shifted" reference trajectory
 
    :param psi: Pole face angle [radians]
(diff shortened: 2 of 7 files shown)


src/elements/DipEdge.H
```diff
@@ -277,6 +287,42 @@ namespace impactx::elements
         /** This pushes the reference particle. */
         using Thin::operator();
 
+        /** This pushes the reference particle.
+         *
```
cemitch99, line 295: This is now implemented, but I'm not sure it is currently called.  As a Thin element, the reference push defaults to the operator in `mixin/thin.H` (which does nothing).  Note:  This situation occurs also in `ThinDipole.H`.  TODO:  clarify the special case when a non-trivial reference particle push is needed for a Thin element.

ax3l: If you implement the push again downstream then you overwrite what the `mixin/Thin.H` defined. So it should do the right thing, but we should add a test.

Also, you can remove this line in that case to be clear:
```C++
/** This pushes the reference particle. */
using Thin::operator();
```

docs/source/usage/parameters.rst
```diff
@@ -384,7 +379,19 @@ when expanded to first order in ``g/rc`` (gap / radius of curvature).
 
 By comparison, note that the MAD-X DIPEDGE element uses as input the half-gap ``HGAP = g/2``, and sets the default value ``FINT = 0`` (while the corresponding default value of ``K2`` is set to 1).
 
-This requires these additional parameters:
+Note that the nonlinear model includes a nonzero horizontal translation (depending on the field integral values) that is present even for a particle that begins on the ideal "hard-edge" reference trajectory.
+
+For a beam, this will result in a centroid offset that will produce centroid oscillations in the downstream beamline. In practice, this can be avoided by aligning the downstream elements with the true
+
+horizontal position (after including the effect of the fringe field).  To model this correction, we allow two options in the dipedge model:
```
ax3l: Has an extra empty line mid sentence:
```suggestion
For a beam, this will result in a centroid offset that will produce centroid oscillations in the  downstream beamline.
In practice, this can be avoided by aligning the downstream elements with the true horizontal position (after including the effect of the fringe field).
To model this correction, we allow two options in the dipedge model:
```

docs/source/usage/parameters.rst
```diff
@@ -384,7 +379,19 @@ when expanded to first order in ``g/rc`` (gap / radius of curvature).
...
+
+* the default model, in which the shift due to the fringe field is applied to each beam particle phase space vector but not to the reference particle phase space vector --
+this model makes sense if the shift due to the fringe field is not considered in the baseline design, so that downstream elements are aligned with the "idealized" reference trajectory
+
+* an option in which the shift due to the fringe field is applied to the reference particle phase space vector, but not to the beam particle phase space vector --
```
ax3l: I would mention the option we are referring to here. This element has 12+ options... mention `modify_ref_part` explicitly?

docs/source/usage/python.rst
```diff
@@ -1227,6 +1221,16 @@ This module provides elements and methods for the accelerator lattice.
...
    the corresponding default value of ``K2`` is set to 1).
 
+   Note that the nonlinear model includes a nonzero horizontal translation (depending on the field integral values) that is present even for a particle that begins on the ideal "hard-edge" reference
+   trajectory.  For a beam, this will result in a centroid offset that will produce centroid oscillations in the downstream beamline. In practice, this can be avoided by aligning the downstream elements with
+   the true horizontal position (after including the effect of the fringe field).  To model this correction, we allow two options in the dipedge model:
+
+   * the default model, in which the shift due to the fringe field is applied to each beam particle phase space vector but not to the reference particle phase space vector --
+   this model makes sense if the shift due to the fringe field is not considered in the baseline design, so that downstream elements are aligned with the "idealized" reference trajectory
+
+   * an option in which the shift due to the fringe field is applied to the reference particle phase space vector, but not to the beam particle phase space vector --
```
ax3l: Same suggestions as above in `parameters.rst` :)

src/initialization/InitElement.cpp
```diff
@@ -188,6 +188,7 @@ namespace detail
             amrex::ParticleReal R = 1;
             std::string model_str = "linear";    // default
             std::string location_str = "entry";  // default
+            bool modify_ref_part = false;        // default
```
ax3l, line 191: Read from ini file is missing. Need to  add:
```C++
pp_element.queryAdd("modify_ref_part", modify_ref_part);
```

ax3l (changes requested):
Looks great up to small fixes.

src/initialization/InitElement.cpp
ax3l: fixed

docs/source/usage/parameters.rst
```suggestion
* the option ``modify_ref_part = True`` in which the shift due to the fringe field is applied to the reference particle phase space vector, but not to the beam particle phase space vector --
```

docs/source/usage/parameters.rst
```diff
@@ -384,7 +379,19 @@ when expanded to first order in ``g/rc`` (gap / radius of curvature).
...
 By comparison, note that the MAD-X DIPEDGE element uses as input the half-gap ``HGAP = g/2``, and sets the default value ``FINT = 0`` (while the corresponding default value of ``K2`` is set to 1).
 
-This requires these additional parameters:
+Note that the nonlinear model includes a nonzero horizontal translation (depending on the field integral values) that is present even for a particle that begins on the ideal "hard-edge" reference trajectory.
+
+For a beam, this will result in a centroid offset that will produce centroid oscillations in the downstream beamline. In practice, this can be avoided by aligning the downstream elements with the true
+
+horizontal position (after including the effect of the fringe field).  To model this correction, we allow two options in the dipedge model:
+
+* the default model, in which the shift due to the fringe field is applied to each beam particle phase space vector but not to the reference particle phase space vector --
```
```suggestion
* the option ``modify_ref_part = False`` (default), in which the shift due to the fringe field is applied to each beam particle phase space vector but not to the reference particle phase space vector --
```

docs/source/usage/python.rst
```suggestion
   * the option ``modify_ref_part = True``, in which the shift due to the fringe field is applied to the reference particle phase space vector, but not to the beam particle phase space vector --
```

docs/source/usage/python.rst
```diff
@@ -1227,6 +1221,16 @@ This module provides elements and methods for the accelerator lattice.
    By comparison, note that the MAD-X DIPEDGE element uses as input the half-gap ``HGAP = g/2``, and sets the default value ``FINT = 0`` (while
    the corresponding default value of ``K2`` is set to 1).
 
+   Note that the nonlinear model includes a nonzero horizontal translation (depending on the field integral values) that is present even for a particle that begins on the ideal "hard-edge" reference
+   trajectory.  For a beam, this will result in a centroid offset that will produce centroid oscillations in the downstream beamline. In practice, this can be avoided by aligning the downstream elements with
+   the true horizontal position (after including the effect of the fringe field).  To model this correction, we allow two options in the dipedge model:
+
+   * the default model, in which the shift due to the fringe field is applied to each beam particle phase space vector but not to the reference particle phase space vector --
```
```suggestion
   * the option ``modify_ref_part = False`` (default), in which the shift due to the fringe field is applied to each beam particle phase space vector but not to the reference particle phase space vector --
```

src/elements/DipEdge.H
```diff
@@ -274,8 +284,49 @@ namespace impactx::elements
...
+                amrex::ParticleReal const sec_psi = 1_prt/cos_psi;
+                amrex::ParticleReal const c2_ref = powi<2>(m_g/m_rc) * m_K0 * sin_psi * powi<3>(sec_psi)/2_prt;
+                amrex::ParticleReal const c3_ref = loc * powi<2>(m_g)/m_rc * m_K0 * powi<2>(sec_psi);
+                amrex::ParticleReal const c5_ref = tan_psi / (2_prt * m_rc);
+                amrex::ParticleReal const c13_ref = powi<2>(sin_psi) / (2_prt*powi<3>(cos_psi)) * powi<2>(m_g)/(m_rc*m_R) * m_K4;
+
+                amrex::ParticleReal const pnorm = std::sqrt(pt*pt-1_prt);
+                amrex::ParticleReal const dx_ref = -c3_ref;
+                amrex::ParticleReal const dpx_ref = c13_ref - c2_ref - c3_ref*c5_ref;
+                amrex::ParticleReal const dpz_ref = -1_prt + std::sqrt(1_prt - powi<2>(dpx_ref/pnorm));
```
ax3l: @cemitch99 Codex thinks this misses a `pnorm` factor:
>   /tmp/impactx-pr1431-review/src/elements/DipEdge.H:319: dpz_ref is missing a
  factor of pnorm. RefPart::px/pz are normalized momenta, so after applying
  dpx_ref at fixed energy the longitudinal increment should be:
>
>  amrex::ParticleReal const dpz_ref =
      pnorm * (std::sqrt(1_prt - powi<2>(dpx_ref/pnorm)) - 1_prt);
>
>  As written, the code adds a unitless fractional change directly to pz, so for
  pnorm != 1 the reference particle no longer satisfies px^2 + py^2 + pz^2 ==
  pt^2 - 1. The new pytest does not catch this because it only checks x, px, y,
  and py, and the chosen kick is tiny enough that dpz_ref rounds to zero.
>
>  I’d request adding a test assertion for the reference momentum norm, ideally
  with parameters large enough to make the pz correction observable.

cemitch99:
```suggestion
                amrex::ParticleReal const dpz_ref = pnorm*(-1_prt + std::sqrt(1_prt - powi<2>(dpx_ref/pnorm)));
```

tests/python/test_dipedge_ref.py
```diff
@@ -0,0 +1,188 @@
...
+            rbc["mean_t"],
+            rbc["mean_pt"],
+        ]
+    )
+    vec_part_pred = [x_shift, px_shift, 0.0, 0.0, 0.0, 0.0]
+
+    # access reference particle
+    vec_ref = np.array([ref.x, ref.px, ref.y, ref.py])
+    vec_ref_pred = [0, 0, 0, 0]
+
```
cemitch99, line 149:
```suggestion
vec_ref_on_shell = (ref.px)**2 + (ref.py)**2 + (ref.pz)**2 - (ref.pt)**2 + 1.0
```

tests/python/test_dipedge_ref.py
```diff
@@ -0,0 +1,188 @@
...
+            rbc["mean_x"],
+            rbc["mean_px"],
+            rbc["mean_y"],
+            rbc["mean_py"],
+            rbc["mean_t"],
+            rbc["mean_pt"],
+        ]
+    )
+    vec_part_pred = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
+
```
cemitch99, line 188:
```suggestion
vec_ref_on_shell = (ref.px)**2 + (ref.py)**2 + (ref.pz)**2 - (ref.pt)**2 + 1.0
```

tests/python/test_dipedge_ref.py
```diff
@@ -0,0 +1,188 @@
...
+    # access reference particle
+    vec_ref = np.array([ref.x, ref.px, ref.y, ref.py])
+    vec_ref_pred = [x_shift, px_shift, 0, 0]
+
```
cemitch99:
```suggestion
vec_ref_on_shell = (ref.px)**2 + (ref.py)**2 + (ref.pz)**2 - (ref.pt)**2 + 1.0
```

tests/python/test_dipedge_ref.py
```diff
@@ -0,0 +1,188 @@
...
+    vec_part_pred = [x_shift, px_shift, 0.0, 0.0, 0.0, 0.0]
+
+    # access reference particle
+    vec_ref = np.array([ref.x, ref.px, ref.y, ref.py])
+    vec_ref_pred = [0, 0, 0, 0]
+
+    atol = 1.0e-12
+    np.testing.assert_allclose(vec_part, vec_part_pred, atol=atol)
+    np.testing.assert_allclose(vec_ref, vec_ref_pred, atol=atol)
+
```
cemitch99, line 154:
```suggestion
np.testing.assert_allclose(vec_ref_on_shell, 0.0, atol=atol)
```

tests/python/test_dipedge_ref.py
```diff
@@ -0,0 +1,188 @@
...
+    )
+    vec_part_pred = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
+
+    # access reference particle
+    vec_ref = np.array([ref.x, ref.px, ref.y, ref.py])
+    vec_ref_pred = [x_shift, px_shift, 0, 0]
+
+    atol = 1.0e-12
+    np.testing.assert_allclose(vec_part, vec_part_pred, atol=atol)
+    np.testing.assert_allclose(vec_ref, vec_ref_pred, atol=atol)
```
cemitch99, line 197:
```suggestion
    np.testing.assert_allclose(vec_ref, vec_ref_pred, atol=atol)
    np.testing.assert_allclose(vec_ref_on_shell, 0.0, atol=atol)
```

src/elements/DipEdge.H
ax3l: fixed :)

src/elements/DipEdge.H
```diff
@@ -262,6 +264,14 @@ namespace impactx::elements
                tout = t - dDelta_dpt * inv_delta1 * ((c2 + c3*m_c5)*x + (c10*m_c5 + c8)*powi<3>(x) - c7*powi<2>(y) - c12*powi<4>(y)
                       + (c10*m_c5 - c11*m_c5 - c9)*x*powi<2>(y) + pxout*(-c3 + (c11 + c10)*powi<2>(y) - c10*powi<2>(x)) + 2_prt*pyout*c10*x*y);
 
+               // optional removal of reference particle shift
+               amrex::ParticleReal const c2_ref = m_c2_times_1plusdelta;
+               amrex::ParticleReal const c3_ref = m_c3_times_1plusdelta;
+               T_Real const dx_ref = (m_modify_ref_part) ? -c3_ref : 0_prt;
+               T_Real const dpx_ref = (m_modify_ref_part) ? m_c13 - c2_ref - c3_ref*m_c5 : 0_prt;
+               xout -= dx_ref;
+               pxout -= dpx_ref;
```
ax3l, line 278: Codex wonders if modifying `tout` would be sensible here, too:
>   - [P2] /tmp/impactx-pr1431/src/elements/DipEdge.H:267: modify_ref_part=True
    removes only the zero-order x/px shift from beam particles, but the
    nonlinear map also creates a zero-order t term at /tmp/impactx-pr1431/src/
    elements/DipEdge.H:264. The reference push updates only x/z/px/pz at /tmp/
    impactx-pr1431/src/elements/DipEdge.H:322, so a particle starting exactly on
    the reference orbit can exit with nonzero relative t while refpart.t is
    unchanged. The new test uses g=1e-3, making this O(g^4) residual fall below
    1e-12; with realistic finite gaps it should be visible. I’d either move the
    same dt_ref onto the reference and subtract it from tout, or document/test
    that modify_ref_part intentionally only handles transverse coordinates.

src/elements/DipEdge.H
ax3l, line 119: Codex:
>   - [P2] /tmp/impactx-pr1431/src/elements/DipEdge.H:115: reverse() does not
    invert the new modify_ref_part reference-particle push. The forward
    reference kick uses dpx_ref = c13 - c2 - c3*c5 at /tmp/impactx-pr1431/src/
    elements/DipEdge.H:318; after reverse() flips psi and swaps entry/exit, the
    reversed kick is not -dpx_ref, so a forward plus reverse push leaves
    residual reference momentum. With the test parameters I get a residual px ~=
    -8e-9; with the existing reversibility-test gap g=0.058, it grows to about
    -2.7e-5. Please add a modify_ref_part=True case to the DipEdge reversibility
    test and adjust the reference push/reverse logic accordingly.

src/elements/DipEdge.H
```diff
@@ -262,6 +264,14 @@ namespace impactx::elements
                tout = t - dDelta_dpt * inv_delta1 * ((c2 + c3*m_c5)*x + (c10*m_c5 + c8)*powi<3>(x) - c7*powi<2>(y) - c12*powi<4>(y)
                       + (c10*m_c5 - c11*m_c5 - c9)*x*powi<2>(y) + pxout*(-c3 + (c11 + c10)*powi<2>(y) - c10*powi<2>(x)) + 2_prt*pyout*c10*x*y);
 
+               // optional removal of reference particle shift
```
ax3l, line 271: Codex:
>  - [P2] /tmp/impactx-pr1431/src/elements/DipEdge.H:267: modify_ref_part=True
    cancels the zero-order x/px shift before shift_out(). For a rolled DipEdge
    (rotation != 0), the element-frame fringe shift should rotate into global x/
    y and px/py, while the reference shift is the unrolled design shift. Current
    code removes the whole local shift first, so the roll-induced relative y/py
    effect disappears. Move the reference-shift subtraction to the global frame
    after shift_out(), or intentionally apply the same alignment to the
    reference push and document/test that behavior.

src/elements/DipEdge.H
```diff
@@ -274,8 +284,49 @@ namespace impactx::elements
...
+                amrex::ParticleReal const loc = m_location == dipedge::Location::entry ? 1_prt : -1_prt;
+                auto const [sin_psi, cos_psi] = amrex::Math::sincos(m_psi);
+                amrex::ParticleReal const tan_psi = sin_psi/cos_psi;
+                amrex::ParticleReal const sec_psi = 1_prt/cos_psi;
+                amrex::ParticleReal const c2_ref = powi<2>(m_g/m_rc) * m_K0 * sin_psi * powi<3>(sec_psi)/2_prt;
+                amrex::ParticleReal const c3_ref = loc * powi<2>(m_g)/m_rc * m_K0 * powi<2>(sec_psi);
+                amrex::ParticleReal const c5_ref = tan_psi / (2_prt * m_rc);
+                amrex::ParticleReal const c13_ref = powi<2>(sin_psi) / (2_prt*powi<3>(cos_psi)) * powi<2>(m_g)/(m_rc*m_R) * m_K4;
+
+                amrex::ParticleReal const pnorm = std::sqrt(pt*pt-1_prt);
```
ax3l, line 323: Codex on reference particle push/tracking mode:
>   - [P2] /tmp/impactx-pr1431/src/elements/DipEdge.H:316: the new reference-
    particle push assumes the reference has no vertical momentum. It computes
    pnorm = sqrt(pt*pt - 1) and leaves py unchanged, but then applies the
    horizontal kick using px/pnorm and pz/pnorm. If an incoming reference has
    py != 0, the update no longer preserves px^2 + py^2 + pz^2 = pt^2 - 1, and
    the x/z displacement is also scaled by the total 3D momentum instead of the
    x-z projection. Use the x-z momentum norm for this planar update, or perform
    a full orthonormal 3D update that preserves the shell.

src/elements/DipEdge.H
cemitch99: I think reversibility implementation should be done as a follow-up PR.

src/elements/DipEdge.H
cemitch99: This is a more general issue, not specific to this PR.  (The treatment of errors/rotations for the reference particle, Issue #745 )

src/elements/DipEdge.H
cemitch99: Again, this is because bending is assumed to occur in the x-z plane (as discussed above).  The effect of errors must be addressed separately.

src/elements/DipEdge.H
cemitch99: This is a good catch, and it will be addressed.

src/elements/DipEdge.H
ax3l: Makes sense, so also #745

src/elements/DipEdge.H
ax3l: I leave it up to you.
Personally, I would probably prefer to not break it with a new feature and address it in the same PR (also because we have a simple test already for it in `tests/python/test_reversibility_elements.py`).

cemitch99: Ok, understanding reversibility correctly for this element (which relies on a mixed-variable generating function, composed with a nonzero phase space translation) requires a fair bit of careful thought.  This is not the highest priority, so we may set this PR aside for a while.

ax3l: Ah understood, in that case let us add just an inline FIXME comment? We don't rely on this property just yet.

ax3l: added :)

ax3l (approved):

PR #1432, opened 2026-04-23
Relax pybind11 version requirement for package managers

EZoni:
## Overview

Fix the issue raised in https://github.com/BLAST-ImpactX/impactx/pull/1419#discussion_r3128341519:

> @EZoni I overlooked something here: You do not want to update the version in `find_package` higher than needed, because that way we cannot build with version `3.0.0`, `3.0.1`, `3.0.2`, and `3.0.3` in package managers anymore - even if those versions are totally fine. (e.g., conda-forge is still to ship `3.0.4`, stands at `3.0.3`. Spack has very good version control, too). So in other words: keep this here on `3.0.0` and below just update the version we download to the latest.

## Related PRs

- https://github.com/AMReX-Codes/pyamrex/pull/565
- https://github.com/BLAST-WarpX/warpx/pull/6805


ax3l:
Thank you! Please the same on ImpactX & pyAMReX :pray: 

(openPMD-api updates apply similar in future workflows)

ax3l (approved):
Thank you! 👍

PR #1436, opened 2026-04-27
Simple Element & Lattice Comparisons

ax3l:
When loading PALS or MAD-X files, we often need to compare them to a manually validated ground truth.

This adds new comparison and `isclose` operators to our elements and lattice classes, simplifying this process.

[Docs preview here.](https://impactx--1436.org.readthedocs.build/en/1436/usage/python.html#common-comparison-methods-on-lattice-elements)

## Example

```py
from impactx import elements

a = elements.Drift(ds=1.0, name="d1")
b = elements.Drift(ds=1.0 + 1e-15, name="d2")

a == b      # False — name differs and ds differs by float noise
a.isclose(b)                              # False — name still differs
a.isclose(b, ignore_attributes="name")    # True

# Compare a Drift to its exact-physics counterpart on common fields.
lin = elements.Drift(ds=1.0)
ex = elements.ExactDrift(ds=1.0)
lin.isclose(ex, ignore_attributes=["type"])  # True

# Lattice-wide comparison forwards ignore_attributes to each element.
lattice_a.isclose(lattice_b, ignore_attributes=["name"])

# also as free-standing function
elements.isclose(a, b, ignore_attributes="name")    # True
elements.isclose(lattice_a, lattice_b, ignore_attributes=["name"])
```

## Review Guidance

Easiest to check we like the [syntax in the docs](https://impactx--1436.org.readthedocs.build/en/1436/usage/python.html#common-comparison-methods-on-lattice-elements) and the various cases in the tests (`tests/python/test_element_value_semantics.py`).

### DIFF
--- a/src/python/impactx/extensions/KnownElementsList.py
+++ b/src/python/impactx/extensions/KnownElementsList.py
@@ -1043,7 +1043,7 @@ def _lattice_eq(self, other):
     ``list`` convention.
     """
     if not hasattr(other, "__len__") or not hasattr(other, "__iter__"):
-        return NotImplemented
+        raise ValueError(f"{other} does not implement __len__ or __iter__")
     if len(self) != len(other):
         return False
     for a, b in zip(self, other):
@@ -1074,7 +1074,7 @@ def _lattice_isclose(self, other, *, rtol=1e-12, atol=0.0, ignore_attributes=Non
         element's ``isclose``.
     """
     if not hasattr(other, "__len__") or not hasattr(other, "__iter__"):
-        return False
+        raise ValueError(f"{other} does not implement __len__ or __iter__")
     if len(self) != len(other):
         return False
     for a, b in zip(self, other):
--- a/src/python/impactx/extensions/KnownElementsList.py
+++ b/src/python/impactx/extensions/KnownElementsList.py
@@ -1015,6 +1015,92 @@ def from_dicts(self, dicts: list[dict]):
     self.extend([_element_from_dict(d) for d in dicts])
 
 
+# ---------------------------------------------------------------------------
+# Element-wise value comparison (== and isclose) for lattice containers.
+#
+# Duck-typed across container types: a KnownElementsList compares equal to a
+# plain Python list of elements or to a FilteredElementsList as long as the
+# element sequences match. This mirrors the typical select() workflow where
+# users compare a filtered view against a hand-built reference list.
+#
+# No value-based __hash__ is defined: containers are mutable, and elements
+# inside are mutable. The default identity-based hash inherited from ``object``
+# is preserved (KnownElementsList from pybind11; FilteredElementsList from
+# Python) so the existing weak-reference tracking of selections keeps working.
+# ---------------------------------------------------------------------------
+
+
+def _lattice_eq(self, other):
+    """Element-wise equality with any iterable of elements.
+
+    Duck-typed across container types: a ``KnownElementsList`` compares
+    equal to a plain Python ``list`` of elements or to a
+    ``FilteredElementsList`` as long as their lengths match and every
+    pair of elements compares equal under ``==``. Returns
+    ``NotImplemented`` for non-iterable operands so Python's
+    reflected-equality fallback applies. Mutable containers are
+    deliberately unhashable (``__hash__ = None``), matching the Python
+    ``list`` convention.
+    """
+    if not hasattr(other, "__len__") or not hasattr(other, "__iter__"):
+        return NotImplemented
+    if len(self) != len(other):
+        return False
+    for a, b in zip(self, other):
+        if not (a == b):
+            return False
+    return True
+
+
+def _lattice_isclose(self, other, *, rtol=1e-12, atol=0.0, ignore_attributes=None):
+    """Tolerant element-wise comparison with any iterable of elements.
+
+    For each pair of elements at the same index, calls the element's own
+    ``isclose(rtol=..., atol=..., ignore_attributes=...)``. Lengths must
+    match; foreign or non-iterable operands return ``False``. Defaults
+    match the per-element ``isclose`` (``rtol=1e-12``, ``atol=0.0``).
+
+    Parameters
+    ----------
+    other : iterable of elements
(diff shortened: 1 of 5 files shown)


src/python/impactx/extensions/KnownElementsList.py
```diff
@@ -1015,6 +1015,92 @@ def from_dicts(self, dicts: list[dict]):
...
+    rtol, atol : float
+        Forwarded to each element's ``isclose``.
+    ignore_attributes : str or iterable of str, optional
+        ``to_dict()`` keys to skip when comparing each pair of elements.
+        Includes the special key ``"type"`` to compare across element
+        variants (e.g., ``Drift`` vs ``ExactDrift``). Forwarded to each
+        element's ``isclose``.
+    """
+    if not hasattr(other, "__len__") or not hasattr(other, "__iter__"):
+        return False
```
EZoni: This returns `False` and fails silently if the type of `other` is not "correct". Do you intentionally avoid raising an error?

tests/python/test_element_value_semantics.py
```diff
@@ -0,0 +1,489 @@
...
+def test_lattice_uses_identity_hash():
+    # Containers do not get a value-based __hash__: the inherited identity
+    # hash is preserved (so internal weak-reference tracking of filtered
+    # selections keeps working). Two value-equal containers therefore may
+    # hash differently — they're not meant as dict/set keys for value-based
+    # deduplication.
+    a = _build_lattice()
+    b = _build_lattice()
+    assert a == b
+    assert hash(a) != hash(b)  # identity-based, distinct instances
```
EZoni, line 514: I let Claude Opus 4.7 review this diff and it commented the following:

> This asserts that two distinct objects have different hashes. With CPython's default `id()`-based hash this is almost always true, but hash collisions are theoretically possible (e.g., on 32-bit platforms, or if CPython changes its implementation). The test should instead assert that containers are hashable and that identity-based hashing is used, without asserting distinctness:
> ```python
> # Test that hashing works and is identity-based (not value-based)
> # Don't assert hash(a) != hash(b): coincidental collision is valid
> assert hash(a) == hash(a)  # stable
> assert hash(b) == hash(b)
> ```
> If the intent is to verify that `a` and `b` are not the same object, use `assert a is not b`.

Do you agree that the assertion `assert hash(a) != hash(b)` is fragile?

tests/python/test_element_value_semantics.py
```diff
@@ -0,0 +1,489 @@
...
+    a = elements.Drift(ds=float("nan"))
+    hash(a)
+
+
+# ----------------------------------------------------------------------
+# Sweep over every patched element class
+# ----------------------------------------------------------------------
+
+
+_SKIP = {"KnownElementsList", "FilteredElementsList"}
```
EZoni, line 209: This is duplicated in src/python/impactx/extensions/Elements.py, lines 34-38:
```python
# Names in the elements module that should not receive these methods.
# KnownElementsList / FilteredElementsList are containers, not elements;
# they do not currently expose a singular ``to_dict``, but we list them
# explicitly so the intent is unambiguous.
_SKIP = frozenset({"KnownElementsList", "FilteredElementsList"})
```

Would it be at all possible to use `_SKIP` from there instead of redefining it here? If not, I would at least add a similar inline comment here for clarity.

src/python/impactx/extensions/KnownElementsList.py
ax3l: Good point. `np.isclose()` raises a `ValueError` (`TypeError`, or `DTypePromotionError`, depending on various inputs) in those cases, we could mirror that behavior

tests/python/test_element_value_semantics.py
ax3l: Good point. And we need those exclusion types also in the dashboard, we could make this a global definition in `impactx.elements.` so that we can rely on it.

tests/python/test_element_value_semantics.py
ax3l: While technically true, we do not compute the amount of elements (or number of CI runs) that makes hash collisions in this specific test a nuisance on 32bit platforms.

On 64bit systems, the likelihood of this test to fail is very low... to negligible.

In other words: I think it is being pedantic here, and the suggested alternative makes no real sense besides checking stability of the hash.

tests/python/test_element_value_semantics.py
ax3l: Needs more work, this goes in a follow-up PR: #1459

src/python/impactx/extensions/KnownElementsList.py
ax3l: updated

ax3l:
@EZoni review addressed :)

cemitch99 (commented):
This is a nice feature.  Is the comparison operation symmetric with respect to lattice a and lattice b?  If so, I wonder if the syntax should reflect that symmetry.

ax3l:
Thanks!

>  Is the comparison operation symmetric with respect to lattice a and lattice b?

You mean if
```py
lattice_a.isclose(lattice_b, ignore_attributes=["name"])
```
is equivalent to
```py
lattice_b.isclose(lattice_a, ignore_attributes=["name"])
```
?

yes it is (and same for the `==` operator).

There is a small difference for `LinearMap`/`SpinMap`, which go for their `SmallMatrix` entries through `numpy.allclose`, which is asymmetric: `|a−b| ≤ atol + rtol·|b|`. But it is quite sensible to inherit that behavior (same as `numpy.isclose`). Not a big issue in practice unless for rtols against near-zero comparisons.

> If so, I wonder if the syntax should reflect that symmetry.

I do not understand this part :)

Do you mean this should (also) be a free function `impactx.elements.isclose()` instead of a member function? `math.isclose` / `numpy.isclose` happen to be free functions for unrelated historical reasons, while Python intrinsics like `dict.__eq__`, `list.__eq__`, etc. are methods. I picked a member function / method because it is easier to discover. But we can have both if we want.

cemitch99 (approved):

PR #1441, opened 2026-04-30
CI: SIMD in Release, Fix GPUDataRegistry w/ Python LTO

ax3l:
The SIMD CI runner often runs out of RAM in the 7GB limited GitHub actions. The reason is that large, complex TUs of elements can take around 700MB+ to compile w/ debug symbols on.

Compiling in release mode does not add `-g` and keeps the RAM peak per TU a bit smaller, around 600MB.

Needed for #1264

This uncovered a bug in Python from #1368 that is fixed, too:

## Fix `GPUDataRegistry<T>`

`GPUDataRegistry<T>` stored its registry map and ID counter as function‑local statics inside inline accessor methods (`next_id()::id`, `registry()::reg`). The `extern template struct GPUDataRegistry<T>` declaration in element headers does not fully suppress instantiation of variadic template members (like `emplace<Args...>`), so `pyImpactX/elements.cpp` ended up implicit‑instantiating emplace locally. With pyImpactX built with `CXX_VISIBILITY_PRESET hidden` + LTO (`ImpactX_PYTHON_IPO=ON`, default), the inlined `registry()`/`next_id()` body inside `emplace` materialized hidden copies of `reg` and `id` inside `pyImpactX.so`, isolated from lib's default‑visibility copies.

The GPU data registry is itself temporary until we implement #1381

### DIFF
--- a/src/elements/mixin/dynamicdata.H
+++ b/src/elements/mixin/dynamicdata.H
@@ -37,9 +37,12 @@ namespace impactx::elements::mixin
      * using MyElementData = GPUDataRegistry<MyCoefficients>;
      * @endcode
      *
-     * @warning This template uses function-local static variables. To avoid ODR
-     *          violations (separate registry instances per translation unit), you
-     *          MUST add explicit instantiation control:
+     * @warning This template holds its registry map and ID counter in exported
+     *          inline static data members (`s_next_id`, `s_registry`). To
+     *          guarantee a single shared registry across every translation
+     *          unit and shared library, including downstream DSOs built with
+     *          `-fvisibility=hidden` and/or LTO (e.g. the Python bindings
+     *          module), you MUST add explicit instantiation control:
      *          - In the element header: `IMPACTX_GPUDATA_EXTERN(impactx::elements::MyElement)`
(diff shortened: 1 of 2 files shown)


src/elements/mixin/dynamicdata.H
```diff
@@ -50,15 +50,23 @@ namespace impactx::elements::mixin
...
+        // per element type. Combined with `extern template struct` in element
+        // headers, this guarantees a single shared registry across every TU,
+        // including downstream shared libraries built with `-fvisibility=hidden`
+        // and/or LTO (e.g. the Python bindings module), where the previous
+        // function-local-static design produced a hidden duplicate per DSO and
+        // caused emplace/get to disagree across module boundaries.
+        IMPACTX_EXPORT inline static int s_next_id = 0;
+        IMPACTX_EXPORT inline static std::map<int, std::shared_ptr<T>> s_registry{};
+
+        static int& next_id () { return s_next_id; }
```
ax3l, line 71: Fix: Follow-up to #1368

ax3l:
The speedup above is just the difference between a `Debug` and a `Release` build.

PR #1445, opened 2026-05-01
Unify mapsteps defaults at `10`

ax3l:
Set the default mapsteps value to `10` for `RFCavity`, `SoftSolenoid`, and `SoftQuadrupole` across C++ constructors, Python bindings, extension helpers, and user documentation.

This keeps input-file parsing, Python APIs, and docs in sync.

Before this PR, the defaults where:
- inputs files / C++:
  - `InitElement.cpp` implements 10 (picked as source of truth)
  - `parameters.rst` documents 1 (doc bug)
  - element `.H` header files implement 1, but we always overwrite it explicitly on the call site. thus, mostly notional; currently has no influence on users
- Python:
  - `elements.cpp` implements 5 (inconsistency - actual behavior change for Python users)
  - `python.rst` documents 1 (doc bug)


cemitch99 (approved):
Thanks

PR #1448, opened 2026-05-02
Spin push for ExactSbend

cemitch99:
This PR adds the spin push for the `ExactSbend` element.  Consistent with the phase space push, which is analytically known in closed form, the spin map is analytically determined in closed form. 

This is the most important and nontrivial example of an analytically-determined spin map with fully nonlinear dependence on the phase space variables.

- [x] add draft of spin push, including evaluation of the first spin map factor
- [x] add evaluation of the second spin map factor
- [x] add benchmarks

Suggestions:  consistency comparisons against `Sbend` and `ExactCFbend` with spin.
In particular, scaling tests can be performed (with respect to the phase space deviation and the map stepsize, respectively for the two comparisons above).


qianglbl (approved):

src/elements/ExactSbend.H
```diff
@@ -363,13 +370,91 @@ namespace impactx::elements
...
+            [[maybe_unused]] T_Real & AMREX_RESTRICT py,
+            [[maybe_unused]] T_Real & AMREX_RESTRICT pt,
+            [[maybe_unused]] T_Real & AMREX_RESTRICT sx,
+            [[maybe_unused]] T_Real & AMREX_RESTRICT sy,
+            [[maybe_unused]] T_Real & AMREX_RESTRICT sz,
+            [[maybe_unused]] T_IdCpu & AMREX_RESTRICT idcpu,
+            [[maybe_unused]] RefPart const & AMREX_RESTRICT refpart
+        ) const
+        {
+            using namespace amrex::literals; // for _rt and _prt
```
ax3l, line 404: C (double) used when compling in single precision for `sqrt`:
```suggestion
            using namespace amrex::literals; // for _rt and _prt
            using namespace std; // for cmath(float)
```

src/elements/ExactSbend.H
```diff
@@ -363,13 +370,91 @@ namespace impactx::elements
...
+
+            // For this element, note that the time-of-flight from entry to
+            // exit is needed for evaluation of the spin map, so the phase
+            // space push is applied first, in order to obtain the change in t.
+
+            // phase space push
+            (*this)(x, y, t, px, py, pt, idcpu, refpart);
+
+            // particle transit time through the slice
+            T_Real dt = t - ti + m_slice_ds*m_ibeta;
```
ax3l: Codex:
> • Found one issue I’d raise on the PR:
>
>  - src/elements/ExactSbend.H:427: the new spin push uses m_slice_ds to
    reconstruct the particle transit time and reference-frame rotation (dt = ...
    + m_slice_ds*m_ibeta, then m_slice_ds/m_rc - theta). That is only equivalent
    to the bend angle when B == 0, because then rc = ds / phi. But ExactSbend
    explicitly supports B != 0, where m_rc comes from rigidity/B while the map
    still bends by m_slice_phi. For those cases, including existing cyclotron-
    style uses, the design-particle spin rotation will scale with ds/rc instead
    of the actual bend angle m_slice_phi. The new test uses default B=0, so it
    will not catch this branch. I’d ask to replace the reference-angle/time
    pieces with the same geometric quantity used by the phase-space map, or add
    a B != 0 regression test if the current formula is intentional.

src/elements/ExactSbend.H
ax3l: fixed.

src/elements/ExactSbend.H
ax3l: Same phrasing in another session of Codex:
>   - P2: src/elements/ExactSbend.H:440 uses m_slice_ds/m_rc for the second spin-
    map factor, but the phase-space map’s sector geometry is cached separately
    as m_slice_phi at src/elements/ExactSbend.H:128. For the default B=0 path
    these are consistent, but ExactSbend exposes B as an independent way to set
    rc; if B and phi/ds differ, the spin frame rotation can silently use a
    different bend angle than the orbit map. Either use m_slice_phi here or
    explicitly reject/enforce inconsistent B, phi, and ds.

and its comment on tests
>  - P2: The new regression test does not exercise several terms in the new spin
    map. examples/spin_tracking/run_exact_sbend_spin_scaling.py:39 reuses
    initial_coords.csv, whose rows keep px=0 and pt=0; that means the lambdax
    term at src/elements/ExactSbend.H:432 and the off-energy gamma/theta
    dependence at src/elements/ExactSbend.H:414 are effectively untested. I’d
    add at least one scaling set with nonzero px and pt.

cemitch99: Codex is correct:  There are two branches here that need to be captured, and I'll check the treatment of the second branch now.  Regarding the test - I agree.  I will modify initial_coords.csv, which will also provide a stronger spin test for the ExactQuad also.

cemitch99: Update:  The branch using the magnetic field also works.  I updated the test to include this case, and I also modified initial_coords.csv -> initial_coords_sbend.csv to test nonzero dependence on all 6 phase space variables.  This is a strong test.

ax3l: That is awesome, thank you for the improvements!

examples/spin_tracking/README.rst
```diff
@@ -457,3 +457,48 @@ The analysis can be run using the following script:
...
+
+
+.. _examples-exact-sbend-spin-scaling:
+
+Scaling Test of Spin in an Exact Sector Bend
+============================================
+
+This example tests the propagation of spin through the :py:class:`~impactx.elements.ExactSbend` element.
+
+The :py:class:`~impactx.elements.ExactSbend` element incorportates the full nonlinear phase space dependence appearing in both the Hamiltonian and the Thomas-BMT equations for tracking.
```
```suggestion
The :py:class:`~impactx.elements.ExactSbend` element incorporates the full nonlinear phase space dependence appearing in both the Hamiltonian and the Thomas-BMT equations for tracking.
```

src/elements/ExactSbend.H
```diff
@@ -363,13 +370,92 @@ namespace impactx::elements
...
+
+            lambdax = m_gyro * theta * (gamma - 1_prt) * uy * ux;
+            lambday = m_gyro * theta * ( -gamma + (gamma - 1_prt) * uy * uy );
+            lambdaz = m_gyro * theta * (gamma - 1_prt) * uy * uz;
+
+            // push the spin vector using the generator just determined (first factor)
+            rotate_spin(lambdax,lambday,lambdaz,sx,sy,sz);
+
+            lambdax = 0_prt;
+            lambday = m_slice_ds/m_rc - theta;
```
ax3l: Codex has a comment on how we use `phi` in the spin update:
>   1. src/elements/ExactSbend.H:428 and src/elements/ExactSbend.H:441 use
     m_slice_ds / m_rc for the spin transit/frame angle. That only matches the
     actual bend angle when ds / rc == phi. ExactSbend supports B != 0, where rc
     = rigidity / B while phi remains the user-provided bend angle, so these can
     differ. The phase-space map itself uses m_slice_phi * m_rc as the bend
     reference distance, so spin should likely use m_slice_phi * m_rc * m_ibeta
     for the transit-time offset and m_slice_phi - theta for the second
     rotation. Existing cases like examples/cyclotron/run_cyclotron.py:52 rely
     on B-backed ExactSbend with nontrivial ds, phi, and B.

with corresponding comment on the test:
>   2. examples/spin_tracking/initial_coords_sbend.csv:2 initializes every
     particle with spin (0, 1, 0), parallel to the bend’s y-axis rotation. That
     makes the reference y-axis spin rotation, including the questionable angle
     above, invisible to the new test. At least one row should use a non-y spin
     direction, preferably normalized, and the test should include a B-backed
     case where ds / rc != phi.

The suggested change to match what we do in the phase space push would be:

First add a pass-through for B=0:
```C++
  // zero field: exact drift, spin is unaffected
  if (m_phi == 0_prt && m_B == 0_prt) {
      (*this)(x, y, t, px, py, pt, idcpu, refpart);
      return;
  }
```

Then change here:
```suggestion
            // particle transit time through the slice
            T_Real dt = t - ti + m_slice_phi * m_rc * m_ibeta;

            // advance in cyclotron phase
            T_Real theta = dt / m_rc * m_beta_gamma / gamma;

            lambdax = m_gyro * theta * (gamma - 1_prt) * uy * ux;
            lambday = m_gyro * theta * ( -gamma + (gamma - 1_prt) * uy * uy );
            lambdaz = m_gyro * theta * (gamma - 1_prt) * uy * uz;

            // push the spin vector using the generator just determined (first factor)
            rotate_spin(lambdax, lambday, lambdaz, sx, sy, sz);

            lambdax = 0_prt;
            lambday = m_slice_phi - theta;
```
Rationale: for `B == 0, m_rc = m_ds / m_phi`, so `m_slice_phi * m_rc == m_slice_ds` and behavior is unchanged. For `B != 0`, the phase-space map’s geometry is set by `m_slice_phi` and `m_rc`, so the spin map should derive the reference arc/time from the same pair.

ax3l (commented):
Just a small fix on transit time and trivial case / drift treatment.

src/elements/ExactSbend.H
cemitch99: Addressed.

PR #1449, opened 2026-05-04
Elements: Consistent Defaults

ax3l:
In our current workflow, we have to repeat ourselves quite a bit when it comes to defaults of element parameters:
- C++ constructor: purely informational, not used anywhere unless someone uses ImpactX as C++ library and does not pass explicit values (internally, we do pass explicit values in init)
- InitElements: the ParmParse/inputs defaults
- elements.cpp: the Python defaults
- parameters.rst: the ParmParse/inputs docs
- python.rst: the Python docs

This PR simplifies this for the first three to a single, authoritative source: the C++ implementation of the element.

PR #699 will then also unify the `python.rst` file into the same authorative source, leaving only the `parameters.rst` to be manually (or LLM-y) kept right.

Result, less repetition: 5x same value -> 2x same value

[DRY programming](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) like this is desirable where possible (not a hard rule*), since repetition leads quickly inconsistencies over time (e.g., #1445).
(*Programming _everything_ DRY can also lead to hard-to-extend, rigid code, so we strike a balance.)

## Changed Defaults

Same as #1445, the `ExactCFbend`, `ExactMultipole`, `ExactQuad` now use 10 mapsteps as the default.
    
Old C++ default was 10, too (init elements). But:
- C++ docs were off to 5.
- Python & its docs where off to 5.


src/elements/Buncher.H
```diff
@@ -43,6 +43,11 @@ namespace impactx::elements
         static constexpr auto type = "Buncher";
         using PType = ImpactXParticleContainer::ParticleType;
 
+        static constexpr amrex::ParticleReal DEFAULT_dx = 0;
+        static constexpr amrex::ParticleReal DEFAULT_dy = 0;
+        static constexpr amrex::ParticleReal DEFAULT_rotation_degree = 0;
```
ax3l: Could move those 3 into `mixin::Alignment`

src/elements/Buncher.H
```diff
@@ -43,6 +43,11 @@ namespace impactx::elements
...
+        static constexpr std::nullopt_t DEFAULT_name = std::nullopt;
```
ax3l: Could move this one into `mixin::Named`

src/elements/CFbend.H
```diff
@@ -47,6 +47,14 @@ namespace impactx::elements
         static constexpr auto type = "CFbend";
         using PType = ImpactXParticleContainer::ParticleType;
 
+        static constexpr amrex::ParticleReal DEFAULT_dx = 0;
+        static constexpr amrex::ParticleReal DEFAULT_dy = 0;
+        static constexpr amrex::ParticleReal DEFAULT_rotation_degree = 0;
+        static constexpr amrex::ParticleReal DEFAULT_aperture_x = 0;
+        static constexpr amrex::ParticleReal DEFAULT_aperture_y = 0;
```
ax3l: Can move those into `mixin::PipeAperture`

src/elements/CFbend.H
```diff
@@ -68,13 +76,13 @@ namespace impactx::elements
...
-            amrex::ParticleReal aperture_x = 0,
-            amrex::ParticleReal aperture_y = 0,
-            int nslice = 1,
-            std::optional<std::string> name = std::nullopt
+            amrex::ParticleReal dx = DEFAULT_dx,
+            amrex::ParticleReal dy = DEFAULT_dy,
+            amrex::ParticleReal rotation_degree = DEFAULT_rotation_degree,
+            amrex::ParticleReal aperture_x = DEFAULT_aperture_x,
+            amrex::ParticleReal aperture_y = DEFAULT_aperture_y,
+            int nslice = DEFAULT_nslice,
```
ax3l, line 75: Can move this into `mixin::Thick`

ax3l: done.

src/elements/ExactCFbend.H
```diff
@@ -61,6 +61,10 @@ namespace impactx::elements
 
         using DynamicData = mixin::GPUDataRegistry<CFbendCoefficients>;
 
+        static constexpr int DEFAULT_unit = 0;
```
ax3l, line 66: In a follow-up PR, these deserve their own helper/mixin and `AMREX_ENUM` because we just support two options. Needs to be clarified and unified.

ax3l: - magnetic elements: scale by magnetic rigidity or not
  - raw SI: direct machine parameters
  - scaled units (by mag. rig.): scale with energy
- RFCavity: peak field on axis
  - raw SI: direct machine parameters
  - scaled units: for $$\partial \gamma / \partial s$$ gamma change wrt to rest energy, independent of charge/mass ratio (eqv. a bit tomag rig, but for E fields)

ax3l: Moved to #1451

src/elements/ExactCFbend.H
```diff
@@ -61,6 +61,10 @@ namespace impactx::elements
...
+        static constexpr int DEFAULT_int_order = 2;
+        static constexpr int DEFAULT_mapsteps = 1;
```
ax3l: Checked: before this PR the default input was `10` in C++, `5` in Python bindings.

Thus, should probably be:
```suggestion
        static constexpr int DEFAULT_mapsteps = 10;
```

src/elements/ExactMultipole.H
```diff
@@ -68,6 +67,10 @@ namespace impactx::elements
 
         using DynamicData = mixin::GPUDataRegistry<MultipoleCoefficients>;
 
+        static constexpr int DEFAULT_unit = 0;
+        static constexpr int DEFAULT_int_order = 2;
+        static constexpr int DEFAULT_mapsteps = 1;
```
ax3l:
```suggestion
        static constexpr int DEFAULT_mapsteps = 10;
```

cemitch99 (commented):
I'm taking my time a bit with this one.  I have a slight concern that this increases the complexity of the element code by introducing a large number of DEFAULT_ parameters, most of which are zero.  Otherwise, I agree it is a good way to ensure consistency.

ax3l:
Yeah, I found the consistency to be worth the slight verbosity. Let me know where you arrive at @cemitch99

cemitch99 (approved):
Ok, let's do it.

cemitch99:
@ax3l Some conflicts appear now.

PR #1402, opened 2026-04-16
Example to run the simple Booster expressed as ImpactX script

egstern:
an python fragment to instantiate the equivalent ImpactX lattice.

## To Do

- [x] doc: scaffolding @ax3l 
- [x] doc: details on the booster @egstern 
- [x] doc: check we are happy with the [preview](https://impactx--1402.org.readthedocs.build/en/1402/usage/examples/simple_booster/README.html) @egstern @ax3l 
- [x] analysis script: decide if we want to add a beam monitor and/or compare any beam moments for correctness @egstern @ax3l 
- [x] merge `development` in after #1454 was merged

PR #1454, opened 2026-05-06
Fix dashboard lattice state sync

ax3l:
- Normalize pybind types like `typing.SupportsFloat[...]` and `typing.SupportsInt[...]` before validation.
- Recompute lattice statistics explicitly after lattice add, delete, reorder, import, and parameter updates.
- Refresh lattice parameters when dashboard variables change, so `sim_input` resolves correctly.
- Reset derived lattice/stat state on dashboard reset.
- Make the dashboard test helper wait for lattice list growth after clicking `ADD`.

First seen in #1402


ax3l:
@egstern @cemitch99 the dashboard seems to fail multiple PRs now, merging this fix to unbreak them

Issue #1453, opened 2026-05-06
Fix Alignment in Spin Push


ax3l:
Isolated from #1433

I noticed a general issue:
In the phase space push, we do `shift_in`/`shift_out` for alignment errors. But we do not do them for the spin push (generally) yet in the same manner... Not related to this element only, but to the pattern we use.

ax3l:
Oopsi, duplicate of #1328

PR #1455, opened 2026-05-07
Handle Alignment in Spin Push

### LINKED ISSUE #1328: Treat translation/rotation errors with spin

ax3l:
Fix #1328 #1453

Move out alignment shift and pipe aperture from individual elements into beam optics logic.

Rotate spin transverse vector in/out.

Enable all tests that we had to skip so far.

- [x] rebase after #1457

### DIFF
--- a/tests/python/test_reversibility_elements.py
+++ b/tests/python/test_reversibility_elements.py
@@ -270,25 +270,17 @@ def test_ChrPlasmaLens(sim, unit, k):
 
 @pytest.mark.parametrize("sim", [True, False], indirect=True, ids=["spin", "nospin"])
 def test_ChrQuad(sim):
-    kwargs = {} if sim.spin else PIPE_KWARGS
     roundtrip(
-        elements.ChrQuad(ds=1.0, k=1.0, unit=0, nslice=nslice, **kwargs),
+        elements.ChrQuad(ds=1.0, k=1.0, unit=0, nslice=nslice, **PIPE_KWARGS),
         sim,
         spin=sim.spin,
     )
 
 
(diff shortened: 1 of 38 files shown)


tests/python/test_reversibility_elements.py
```diff
@@ -264,26 +264,18 @@ def test_ChrPlasmaLens(sim, unit, k):
...
+        elements.ChrQuad(ds=1.0, k=1.0, unit=0, nslice=nslice, **PIPE_KWARGS),
         sim,
         spin=sim.spin,
     )
 
 
+@pytest.mark.parametrize("sim", [True, False], indirect=True, ids=["spin", "nospin"])
 def test_ChrQuad_marylie_unit(sim):
     roundtrip(
         elements.ChrQuad(ds=1.0, k=3.5, unit=1, nslice=nslice, **PIPE_KWARGS),
```
ax3l, line 281: The newly enabled MaryLie spin test is now exercising the defocusing branch/stronger spin case, and shows low precision...

It is weird that it passes locally on my PC and in half of the runners...

ax3l: fixed in #1457

cemitch99 (approved):

cemitch99:
The approach looks reasonable to me, but do those failing tests need some investigation?

ax3l:
The failing performance tests I finally understood and I need to change the reference-loaded CPU (serial) and GPU code helpers to explicit copies to/from registers, similar to how we do SIMD code.

PR #1456, opened 2026-05-07
ChrQuad: Clean `lambdax` Double Update w/ Spin

ax3l:
- `lambdax` was updated twice (unnecessary, result was the same)
- formatting improved


cemitch99 (approved):
Good catch, thanks.  Looks good, pending tests.

PR #1460, opened 2026-05-10
Diagnostics File Prefix

ax3l:
Currently, we write all diagnostics files, incl. beam monitors into a hard-coded directory `diags/`.

In WarpX, we made this configurable a while ago and this adds a similar, closely-named option here:
* Inputs file: `diag.file_prefix = diags`
* Python: `sim.diag_file_prefix = "diags"`

The option supports sub-folders, no-folders (flat in PWD) and does sensible checks in terms of "back up an existing diags folder", [pls see docs](https://impactx--1460.org.readthedocs.build/en/1460/usage/python.html#impactx.ImpactX.diag_file_prefix).

Recently wondered about by @SchroederSa 

To check:
- [ ] maybe update dashboard too, mostly for file import... (I have a local commit for this ready, but it is noisy and not a central feature we need to expose in the GUI)

### DIFF
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -238,7 +238,7 @@ If the same element name is used multiple times, then an output series is create
 * ``<element_name>.name`` (``string``, default value: ``<element_name>``)
 
   The output series name to use.
-  By default, output is created under ``diags/openPMD/<element_name>.<backend>``.
+  By default, output is created under ``<diag.file_prefix>/openPMD/<element_name>.<backend>``.
 
 * ``<element_name>.backend`` (``string``, default value: ``default``)
 
@@ -1356,13 +1356,25 @@ Diagnostics and output
   By default, diagnostics are computed and written at the beginning and end of the simulation.
   Enabling this flag will write diagnostics at every step and slice step.
 
+* ``diag.file_prefix`` (``string``, optional, default: ``diags``)
+
+  Root directory for diagnostic output.
+  By default, diagnostics are written in the folder ``diags/``.
+
+  Set to an empty string or ``.`` to write diagnostics in the current working directory.
+
+  By default, existing diagnostic directories are moved when a new simulation initializes.
+  Prefixes that resolve to the current working directory, the root directory,
+  or an ancestor of the current working directory are written as requested,
+  but are not moved out of the way during initialization.
+
 * ``diag.file_min_digits`` (``integer``, optional, default: ``6``)
 
   The minimum number of digits used for the step number appended to the diagnostic file names.
 
 * ``diag.backend`` (``string``, default value: ``default``)
 
-  Diagnostics for particles lost in apertures, stored as ``diags/openPMD/particles_lost.*`` at the end of the simulation.
+  Diagnostics for particles lost in apertures, stored as ``<diag.file_prefix>/openPMD/particles_lost.*`` at the end of the simulation.
   See the ``beam_monitor`` element for backend values.
 
 * ``diag.eigenemittances`` (``boolean``, optional, default: ``false``)
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -1363,10 +1363,11 @@ Diagnostics and output
 
   Set to an empty string or ``.`` to write diagnostics in the current working directory.
 
-  By default, existing diagnostic directories are moved when a new simulation initializes.
-  Prefixes that resolve to the current working directory, the root directory,
-  or an ancestor of the current working directory are written as requested,
-  but are not moved out of the way during initialization.
+  If a directory at ``diag.file_prefix`` already exists when a simulation starts,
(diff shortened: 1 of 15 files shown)


src/tracking/envelope.cpp
```diff
@@ -82,10 +83,11 @@ namespace impactx
             pp_diag.queryAddWithParser("file_min_digits", file_min_digits);
 
             // print initial reference particle to file
-            diagnostics::DiagnosticOutput(ref, "diags/ref_particle");
+            diagnostics::DiagnosticOutput(ref, diagnostics::FilePrefixPath("ref_particle"));
 
             // print the initial values of reduced beam characteristics
-            diagnostics::DiagnosticOutput(cm, ref, "diags/reduced_beam_characteristics");
+            diagnostics::DiagnosticOutput(
+                cm, ref, diagnostics::FilePrefixPath("reduced_beam_characteristics"));
```
ax3l: Hm, maybe I move the `diagnostics::FilePrefixPath` wrapping into `diagnostics::DiagnosticOutput`...

docs/source/usage/parameters.rst
```diff
@@ -1356,13 +1356,25 @@ Diagnostics and output
...
   Enabling this flag will write diagnostics at every step and slice step.
 
+* ``diag.file_prefix`` (``string``, optional, default: ``diags``)
+
+  Root directory for diagnostic output.
+  By default, diagnostics are written in the folder ``diags/``.
+
+  Set to an empty string or ``.`` to write diagnostics in the current working directory.
+
+  By default, existing diagnostic directories are moved when a new simulation initializes.
```
SchroederSa: I have to admit, this is paragraph is not super clear to me.

- what do you mean by 'existing directories are moved'? -> I guess, you mean the 'mv x y' command, but  'renamed to diags.xxx' might be more human-readable.
- what do you mean by 'resolve to...' and 'moved out of the way'?

docs/source/usage/python.rst
```diff
@@ -246,6 +246,17 @@ Collective Effects & Overall Simulation Parameters
       By default, diagnostics is performed at the beginning and end of the simulation.
       Enabling this flag will write diagnostics every step and slice step.
 
+   .. py:property:: diag_file_prefix
+
+      Root directory for diagnostic output (default: folder named ``"diags"``).
+
+      Set to ``""`` or ``"."`` to write diagnostics in the current working directory.
+
+      By default, existing diagnostic directories are moved when a new simulation initializes.
```
SchroederSa: Same comment as above

src/diagnostics/FilePrefix.H
```diff
@@ -0,0 +1,33 @@
...
+#define IMPACTX_DIAGNOSTICS_FILE_PREFIX_H
+
+#include <string>
+
+
+namespace impactx::diagnostics
+{
+    /** Root directory for diagnostic output.
+     *
+     * This queries ParmParse every time so runtime changes to diag.file_prefix
```
SchroederSa: I was stumbling over this sentence a little.

Making a coordination conjunction by breaking action and result by 'so that' or an additional semicolon to break action and result (', so') might simplify the read :).

src/diagnostics/FilePrefix.H
```diff
@@ -0,0 +1,33 @@
...
+     * are honored.
+     * Empty string and "." mean the current working directory.
+     */
+    std::string FilePrefix ();
+
+    /** Path to a diagnostic file or directory under diag.file_prefix.
+     */
+    std::string FilePrefixPath (std::string const & name);
+
+    /** Whether diag.file_prefix can be moved out of the way during initialization.
```
SchroederSa: 'moving out of the way' sounds nebulous

SchroederSa (approved):
Fantastic! You guys are amazing - thanks for this fast turnaround! It's exactly what I was hoping for. My sub-directory-clutter is having an end!

-Sarah

docs/source/usage/parameters.rst
ax3l: yeah exactly we `mv diags diags.<unix-seconds>` out of the way if one is present at the start of the simulation.

src/tracking/envelope.cpp
ax3l: Improved.

docs/source/usage/parameters.rst
ax3l: Wording improved! :)

src/diagnostics/FilePrefix.H
ax3l: Good point, applied as you suggested.

docs/source/usage/python.rst
ax3l: Wording improved! :)

src/diagnostics/FilePrefix.H
ax3l: Clarified :)

cemitch99 (approved):

PR #1465, opened 2026-05-13
Catch `double/float` to `int` Narrowing Bugs

ax3l:
Adds compiler and linter coverage for the implicit `double/float → int` narrowing class of bug fixed in #1457 (forgetting `using namespace std;` in templated scalar/SIMD code, causing `abs()` to resolve to C `::abs(int)` and silently truncate). `-Wall -Wextra` does not catch this.

  - **CMake:** add `-Wfloat-conversion` to the GCC/Clang/AppleClang warning set in `impactx_set_compile_warnings`.
  - **clang-tidy:** enable `bugprone-narrowing-conversions`, scoped via `CheckOptions` to only the floating→integer class (skips `int↔int`, sign, and equivalent-bit-width to avoid noise).
  - **Fixes uncovered by the new flags:** two Python property setters in `src/python/elements.cpp` accepted `amrex::ParticleReal` and assigned to an `int` field, silently truncating from Python. Small bug, but did not cause wrong results.

Note: AMReX is SYSTEM-included, so warnings from its templates are not surfaced here; an AMReX-side audit is left for a follow-up.


cemitch99 (approved):

atmyers (approved):

src/elements/ExactCFbend.H
```diff
@@ -384,8 +384,8 @@ namespace impactx::elements
                //amrex::Print() << "n, kn, ks before scaling: " << n << ", " << kn << ", " << ks << "\n";
 
                // define dimensionless multipole coefficients (scaled by radius of curvature)
-               kn = std::pow(m_rc, m) * kn;
-               ks = std::pow(m_rc, m) * ks;
+               kn = std::pow(m_rc, static_cast<amrex::ParticleReal>(m)) * kn;
+               ks = std::pow(m_rc, static_cast<amrex::ParticleReal>(m)) * ks;
```
ax3l, line 389: This will be a case for https://github.com/AMReX-Codes/amrex/pull/5446

PR #1295, opened 2026-02-05
New MAD-X Parser

ax3l:
## Summary

This PR overhauls two files:
- `MADXParser.py`: a tokenizer and parser that tries to truthfully reimplement the MAD-X C parser, ultimately returning a flattened MAD-X lattice. This file is best written with a powerful LLM (mostly Claude 4.6 was used). It is covered by unit tests for central MAD-X input language features and should not require much updates in the future.
- `madx_to_impactx.py`: the actual translator from MAD-X element to ImpactX element(s). This file we need to be able to maintain "by hand" as we add new elements and handle physics translation corner-cases.

## To Do

- [x] sudo Vibe me a MAD-X parser in Python
  - [x] iterate, test, iterate, test, ...
- [x] Existing MAD-X examples still work.
- [x] Do more complex examples parse now?
  - [x] [simple booster](https://github.com/egstern/impactx/blob/parse_madx_lattice/examples/parse_madx_lattice/sbbooster-cooked.madx)
  - [x] Nov 2024 Booster file `ring_seq_bb_spch_thin (2).madx` (GDrive)
  - [x] Eric's pretty complete `pip-ii-booster-combined.madx` and `sbbooster-cooked.madx` #1271
  - [x] IOTA v8.3 / v84_2 (2019-01-01)
  - [x] LHC: `acc-models-lhc/scenarios/cycle/pp/ramp/1178/model.madx` ([gitlab](https://gitlab.cern.ch/acc-models/acc-models-lhc))
  - [x] FCC: via Arianna
- [ ] How do we want to control element models (linear, paraxial, hamiltonian)?
  - [ ] for MAD-X and PALS as a start, we could have a model option in `from_file(...)`
- [x] Tests: MAD-X input language
- [ ] Tests: commit more MAD-X physics files to our repo?
- [x] Close #104: MAD-X input reader: constants, elements, lines, segments
- [x] Close #512: MAD-X: Call, FILENAME Support
- [x] Close #709: MAD-X Reader: RBEND

PR #1467, opened 2026-05-15
MAD-X Line Continuation

ax3l:
Support `&` line continution in MAD-X files.

### DIFF
--- a/src/python/impactx/MADXParser.py
+++ b/src/python/impactx/MADXParser.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 #
-# Copyright 2022-2023 ImpactX contributors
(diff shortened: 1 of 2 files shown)


src/python/impactx/MADXParser.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 #
-# Copyright 2022-2023 ImpactX contributors
-# Authors: Matthias Frey, Andreas Adelmann, Marco Garten, Axel Huebl
+# Copyright 2022-2026 ImpactX contributors
+# Authors: Axel Huebl
```
ax3l, line 4: Forgot to update authors in #1295. This file was fully rewritten in the last PR, so the old credits are now misleading / can cause confusion.

egstern (approved):
Looks good to me

PR #1469, opened 2026-05-15
Implement More Hooks

ax3l:
Allow users to interact with slices and specific collective effects in more detail.


ax3l:
Ah wait, it is not that easy because we store the hooks inside the ImpactX sim and those are free standing functions.

PR #1472, opened 2026-05-27
Add beta gamma to rbc

egstern:
We need a measure of beam energy in the reduced beam characteristics. If the beam is accelerating, the rbc record will have the corresponding beam energy. Store beta*gamma as a column of reduced_beam_characteristics.


egstern:
never mind, it is already in the ref_particle

PR #1473, opened 2026-05-28
Add pipe radius to 2.5D Gaussian SC

qianglbl:
added pipe radius for 2.5D Gaussian SC longitudinal kick

### DIFF
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -1068,6 +1068,10 @@ See there ``nslice`` option on lattice elements for slicing.
 
       Initial integral region to avoid divergence of integrand at 0.
 
+    * ``algo.space_charge.gauss_pipe_radius`` (``float``, default: ``1.0`` m)
+
+      Pipe radius parameter for the Gauss2p5D space charge model.
+
     * ``algo.space_charge.gauss_charge_z_bins`` (``int``, default: ``129``)
 
       Number of bins for longitudinal line density deposition.
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -101,6 +101,10 @@ Collective Effects & Overall Simulation Parameters
 
       Number of bins for longitudinal charge density deposition (default: ``129``).  Used by the Gauss2p5D space charge model.
 
+   .. py:property:: space_charge_gauss_pipe_radius
+      
+      Pipe radius parameter for the Gauss2p5D space charge model (default: ``1.0`` m).
+
    .. py:property:: space_charge_num_longitudinal_bins
 
       Number of bins for longitudinal charge density deposition (default: ``100``).  Used by the 2p5D space charge model.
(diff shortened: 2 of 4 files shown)


cemitch99:
@qianglbl These tests fail because of an "unused variable" error, which I'm working on now.  I can push a few changes in a moment.

cemitch99 (approved):

src/particles/spacecharge/Gauss2p5dPush.cpp
```diff
@@ -255,6 +257,7 @@ namespace impactx::particles::spacecharge
 
                     // field integrals from a 2D Gaussian bunch
                     amrex::ParticleReal eintx, einty, eintz;
+                    //potInt(delta,nint,pipe_radius,x,y,sigx,sigy,eintx,einty,eintz);
```
cemitch99:
```suggestion
```

ax3l (approved):

docs/source/usage/parameters.rst
```diff
@@ -1068,6 +1068,10 @@ See there ``nslice`` option on lattice elements for slicing.
 
       Initial integral region to avoid divergence of integrand at 0.
 
+    * ``algo.space_charge.gauss_pipe_radius`` (``float``, default: ``1.0`` m)
+
+      Pipe radius parameter for the Gauss2p5D space charge model.
```
ax3l, line 1073: I wonder if we should just explicitly say that the pipe geometry is cylindrical?

```suggestion
      Pipe cylinder radius for the Gauss2p5D space charge model.
```

docs/source/usage/python.rst
```diff
@@ -101,6 +101,10 @@ Collective Effects & Overall Simulation Parameters
 
       Number of bins for longitudinal charge density deposition (default: ``129``).  Used by the Gauss2p5D space charge model.
 
+   .. py:property:: space_charge_gauss_pipe_radius
+
+      Pipe radius parameter for the Gauss2p5D space charge model (default: ``1.0`` m).
```
ax3l, line 106:
```suggestion
      Pipe cylinder radius for the Gauss2p5D space charge model (default: ``1.0`` m).
```

ax3l:
LGTM, but have a little follow-up question we can improve separately.

docs/source/usage/parameters.rst
ax3l: @qianglbl : And it is an approximation that only influences the long. component of the beam (space charge kick on pt). A typical value when comparion to 3D would be $6 \times \gamma \times \sigma_z$

@egstern : maybe call it longitudinal space charge scale or so.

ax3l: `algo.space_charge.gauss_long_scale`

docs/source/usage/python.rst
```diff
@@ -101,6 +101,10 @@ Collective Effects & Overall Simulation Parameters
 
       Number of bins for longitudinal charge density deposition (default: ``129``).  Used by the Gauss2p5D space charge model.
 
+   .. py:property:: space_charge_gauss_pipe_radius
```
ax3l, line 104:
```suggestion
   .. py:property:: space_charge_gauss_long_scale
```

PR #1474, opened 2026-05-29
TaperedPL: symmetric split for spin

ax3l:
The spin push rotated the spin vector with a Thomas-BMT generator built from the entry momentum, then applied the full transverse kick. `reverse()` only negates the strength and reuses the same "rotate-then-kick" order, so the in/out rotations were evaluated at different momenta (`p0` vs `p1 = p0 - dp`) and did not cancel, leaving an `O(Omega * dp) ~ 7e-8` spin residual that broke reversibility. The orbit, depending only on `(x, y)`, stayed exact.

Evaluate the BMT generator at the transverse momentum halfway through the kick (average of entry and exit), reusing the shared phase-space `operator()` for the kick itself. Because `(x, y, pt)` are invariant under this thin lens and the kick is linear in the reversible strength, the midpoint momentum is identical forward and reversed while the generator is odd in strength, so negating the strength is the exact inverse rotation.

Spin roundtrip residual drops from ~7e-8 to ~3e-16 (machine precision for DP).

Found and fixed w/ Claude while adding tests in #1455

### DIFF
--- a/src/elements/TaperedPL.H
+++ b/src/elements/TaperedPL.H
@@ -125,7 +125,7 @@ namespace impactx::elements
             T_Real & AMREX_RESTRICT py,
             T_Real & AMREX_RESTRICT pt,
             [[maybe_unused]] T_IdCpu & AMREX_RESTRICT idcpu,
-            RefPart const & AMREX_RESTRICT refpart
+            [[maybe_unused]] RefPart const & AMREX_RESTRICT refpart
         ) const
         {
             using namespace amrex::literals; // for _rt and _prt
@@ -134,12 +134,6 @@ namespace impactx::elements
 
             // shift due to alignment errors of the element
             shift_in(x, y, px, py);
-
-            // normalize focusing strength units to MAD-X convention if needed
-            amrex::ParticleReal g = m_k;
-            if (m_unit == 1) {
-                g = m_k / refpart.rigidity_Tm();
-            }
 
             // initialize output values
             T_Real xout = x;
@@ -151,10 +145,10 @@ namespace impactx::elements
 
             // advance position and momentum
             xout = x;
-            pxout = px - g * ( x + m_taper*0.5_prt * (powi<2>(x) + powi<2>(y)) );
+            pxout = px - m_g * ( x + m_taper*0.5_prt * (powi<2>(x) + powi<2>(y)) );
 
             yout = y;
-            pyout = py - g * ( y + m_taper * x * y );
+            pyout = py - m_g * ( y + m_taper * x * y );
 
             tout = t;
             ptout = pt;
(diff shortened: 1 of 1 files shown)


src/elements/TaperedPL.H
```diff
@@ -209,6 +209,30 @@ namespace impactx::elements
             using namespace std; // for cmath(float)
             using amrex::Math::powi;
 
+            // normalize focusing strength units to MAD-X convention if needed
+            amrex::ParticleReal g = m_k;
+            if (m_unit == 1) {
+                g = m_k / refpart.rigidity_Tm();
+            }
+
```
ax3l: This should not be needed, because we guarantee to call `compute_constants` before pushes

```suggestion
```

src/elements/TaperedPL.H
```diff
@@ -209,6 +209,24 @@ namespace impactx::elements
...
 
+            // Symmetric (Strang) split: half momentum kick, then the spin
+            // rotation evaluated at the half-kicked momentum, then the second
+            // half momentum kick. The transverse kicks depend only on (x, y),
+            // which are invariant for this thin lens, so the kick is time-reversal
+            // symmetric and the BMT generator is evaluated at the same physical
+            // momentum in both directions. Reversing the element (negating the
+            // strength) therefore yields the exact inverse of this map; an
+            // asymmetric "rotate then full kick" ordering would leave an
+            // O(Omega * dp) spin residual that breaks reversibility.
```
ax3l: Even if we leave this block in, then this part can still be removed:
```suggestion
            // strength) therefore yields the exact inverse of this map
```

src/elements/TaperedPL.H
```diff
@@ -244,8 +263,9 @@ namespace impactx::elements
             // push the spin vector using the generator just determined
             rotate_spin(lambdax, lambday, lambdaz, sx, sy, sz);
 
-            // phase space push
-            (*this)(x, y, t, px, py, pt, idcpu, refpart);
```
ax3l, line 248: I kinda do not like that this patch does not reuse the phase space push, but since it is so simple it would be fine I guess...

ax3l:
Ah, just realized this patch makes no sense outside of #1455 because the TaperedPL passes the precision for the existing _perfectly_ aligned test even with spin...

src/elements/TaperedPL.H
ax3l: 2nd version committed

src/elements/TaperedPL.H
```diff
@@ -135,12 +135,6 @@ namespace impactx::elements
             // shift due to alignment errors of the element
             shift_in(x, y, px, py);
 
-            // normalize focusing strength units to MAD-X convention if needed
-            amrex::ParticleReal g = m_k;
-            if (m_unit == 1) {
-                g = m_k / refpart.rigidity_Tm();
-            }
```
ax3l, line 142: This is computed in `compute_constants` so we should not need it here.

cemitch99 (approved):
Yes.  This is the right way to make this spin-symmetric.

PR #1476, opened 2026-05-30
CI: CodSpeed Microarch Ubuntu

ax3l:
Pin the microarch for our SIMD [CodSpeed](https://codspeed.io/BLAST-ImpactX/impactx) benchmark to the observed 2021+ CPUs that are currently available on x86-64 Ubuntu 24.04 runners (`march`).
Make more reproducible when newer AMD/Intel CPUs are assigned by pinning to the lower (Zen 3) generation CPUs in the pool.

PR #1478, opened 2026-05-31
Fix SP: `to_py()` Matrix Type

ax3l:
In the lattice Python code serialization, we should not promote `SmallMatrix` blindly to `double` but keep the build-time precision type.

Update is_close tests to use FP-type sensible mini deltas for tests.

Follow-up to #1366 #1436

### DIFF
--- a/src/python/impactx/extensions/KnownElementsList.py
+++ b/src/python/impactx/extensions/KnownElementsList.py
@@ -10,7 +10,7 @@ import os
 import re
 import weakref
 
-from impactx import elements
+from impactx import Config, elements
 
 # All live FilteredElementsList views for a lattice (WeakKeyDictionary: key is KnownElementsList).
 _filtered_views_by_lattice: weakref.WeakKeyDictionary = weakref.WeakKeyDictionary()
@@ -960,6 +960,9 @@ def to_py(self) -> str:
         ]
     )
 
+    # match the SmallMatrix element type to the build's precision
+    real_t = "float" if Config.precision == "SINGLE" else "double"
+
     for d in self.to_dicts():
         element_type = d["type"]
         kwargs = _filter_kwargs(d)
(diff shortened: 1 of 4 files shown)


ax3l:
`ExactDrift[spin]` is a ~80% memory-bound element, and the only thing that moved is CodSpeed's **cache** cost component. The trigger is the runner CPU changing between `development` (`Intel(R) Xeon(R) Platinum 8370C CPU @ 2.80GHz`, Ice Lake) and this PR (`AMD EPYC 7763 64-Core Processor`, Zen 3):

| | Base (Intel Xeon 8370C, Ice Lake) | Head (AMD EPYC 7763, Zen 3) |
|---|---|---|
| L1d | 48 KiB | 32 KiB |
| L2 | 1280 KiB | 512 KiB |
| L3 | 48 MB | 32 MB |

Our `-march=x86-64-v3 -mtune=generic` pin (#1476) works as intended: the retired **instruction count is identical** across both CPUs, and so is the **memory** component. Only the L1-miss-driven **caches** component shifts:

| Cost component | Base (Intel 8370C, AVX-512) | Head (AMD EPYC 7763, no AVX-512) |
|---|---|---|
| instructions | 69.0 µs | 69.1 µs |
| memory | 349.4 µs | 349.4 µs |
| caches | 40.5 µs | 17.5 µs |

Note: CodSpeed's CPU simulation does **not** use the host cache geometry above — it always runs Callgrind with a **fixed** cache model (`--I1=32768,8,64 --D1=32768,8,64 --LL=8388608,16,64`, see [`CodSpeedHQ/runner`](https://github.com/CodSpeedHQ/runner)). The variance instead comes from **glibc reading the host CPU's features/cache sizes at runtime** and tuning `malloc`/`memcpy` (alignment, thresholds): same instructions, slightly different data layout → different L1 hits/misses in the *fixed* simulated cache. This is the effect CodSpeed documents in ["Why glibc is faster on some GitHub Actions runners"](https://codspeed.io/blog/unrelated-benchmark-regression); their planned fix is to spoof uniform CPU features across simulations (not yet released). There is no CodSpeed config to re-pin or disable the cache sim today — only pinning the runner hardware (GitHub Large Runners / CodSpeed Macro Runners) removes it.

**Safe to ignore** for this PR: identical instruction counts, 40/41 benchmarks unchanged, and CodSpeed itself flags this run with an "Environment Differences" (CPU change) banner.

(updated via Claude)

Pushing a commit via 490f2d077504710829d92c49623d33d92df348fd now to stabilize a bit further.

atmyers (approved):

PR #1479, opened 2026-05-31
Tests: SP Tolerances

ax3l:
Add conditional, slightly looser tolerances for single-precision builds in analysis scripts / tests.

### DIFF
--- a/examples/active_plasma_lens/analysis_APL.py
+++ b/examples/active_plasma_lens/analysis_APL.py
@@ -9,6 +9,7 @@
 import glob
 import os
 import re
+from pathlib import Path
 
 import openpmd_api as io
 import pandas as pd
@@ -100,6 +101,6 @@ def read_time_series(file_pattern):
 
 def data_is_double(file_pattern):
     """Detect float precision (single vs double) from a text diagnostic's digits."""
-    text = "".join(open(f).read() for f in glob.glob(file_pattern))
+    text = "".join(Path(f).read_text() for f in glob.glob(file_pattern))
     mantissas = re.findall(r"\d*\.\d+", text)
     return any(len(m.replace(".", "").strip("0")) >= 12 for m in mantissas)
--- a/examples/active_plasma_lens/analysis_APL.py
+++ b/examples/active_plasma_lens/analysis_APL.py
@@ -6,9 +6,12 @@
 #
 # -*- coding: utf-8 -*-
 
+import glob
 import os
+import re
 
 import openpmd_api as io
+import pandas as pd
 from scipy.stats import moment
 
 
@@ -80,11 +83,6 @@ def read_time_series(file_pattern):
     pandas.DataFrame
     """
 
-    import glob
-    import re
-
-    import pandas as pd
-
     def read_file(file_pattern):
         for filename in glob.glob(file_pattern):
             df = pd.read_csv(filename, delimiter=r"\s+")
@@ -98,3 +96,10 @@ def read_time_series(file_pattern):
         axis=0,
         ignore_index=True,
     )  # .set_index('id')
+
+
+def data_is_double(file_pattern):
+    """Detect float precision (single vs double) from a text diagnostic's digits."""
+    text = "".join(open(f).read() for f in glob.glob(file_pattern))
+    mantissas = re.findall(r"\d*\.\d+", text)
+    return any(len(m.replace(".", "").strip("0")) >= 12 for m in mantissas)
--- a/examples/active_plasma_lens/analysis_APL.py
+++ b/examples/active_plasma_lens/analysis_APL.py
@@ -102,5 +102,7 @@ def read_time_series(file_pattern):
 def data_is_double(file_pattern):
     """Detect float precision (single vs double) from a text diagnostic's digits."""
     text = "".join(Path(f).read_text() for f in glob.glob(file_pattern))
-    mantissas = re.findall(r"\d*\.\d+", text)
-    return any(len(m.replace(".", "").strip("0")) >= 12 for m in mantissas)
+    return any(
+        len(m.replace(".", "") if e else m.replace(".", "").strip("0")) >= 12
+        for m, e in re.findall(r"(\d*\.\d+)([eE][+-]?\d+)?", text)
+    )
--- a/examples/charge_sign/analysis_charge_sign.py
+++ b/examples/charge_sign/analysis_charge_sign.py
@@ -81,7 +81,7 @@ print(
 )
 
 atol = 0.0  # ignored
-rtol = 1.0e-11 if is_double else 1.0e-4  # exact to within roundoff tolerance
+rtol = 1.0e-11 if is_double else 1.0e-4
 print(f"  rtol={rtol} (ignored: atol~={atol})")
 
 assert np.allclose(
--- a/examples/charge_sign/analysis_charge_sign.py
+++ b/examples/charge_sign/analysis_charge_sign.py
@@ -38,6 +38,7 @@ def get_moments(beam):
 series = io.Series("diags/openPMD/monitor.h5", io.Access.read_only)
 last_step = list(series.iterations)[-1]
 initial = series.iterations[1].particles["beam"].to_df()
+is_double = initial["position_x"].dtype == np.float64
 final = series.iterations[last_step].particles["beam"].to_df()
 
 # compare number of particles
@@ -80,7 +81,7 @@ print(
 )
 
 atol = 0.0  # ignored
-rtol = 1.0e-11  # exact to within roundoff tolerance
+rtol = 1.0e-11 if is_double else 1.0e-4  # exact to within roundoff tolerance
 print(f"  rtol={rtol} (ignored: atol~={atol})")
 
 assert np.allclose(
--- a/examples/reversibility/analysis_zero_bend_inverse.py
+++ b/examples/reversibility/analysis_zero_bend_inverse.py
@@ -81,7 +81,7 @@ print(
 )
 
 atol = 0.0  # ignored
-rtol = 1.0e-12 if is_double else 1.0e-10  # exact to within roundoff tolerance
+rtol = 1.0e-12 if is_double else 1.0e-10
 print(f"  rtol={rtol} (ignored: atol~={atol})")
 
 assert np.allclose(
--- a/examples/reversibility/analysis_zero_bend_inverse.py
+++ b/examples/reversibility/analysis_zero_bend_inverse.py
@@ -38,6 +38,7 @@ def get_moments(beam):
 series = io.Series("diags/openPMD/monitor.h5", io.Access.read_only)
 last_step = list(series.iterations)[-1]
 initial = series.iterations[1].particles["beam"].to_df()
+is_double = initial["position_x"].dtype == np.float64
 final = series.iterations[last_step].particles["beam"].to_df()
 
 # compare number of particles
@@ -80,7 +81,7 @@ print(
 )
 
 atol = 0.0  # ignored
-rtol = 1.0e-12  # exact to within roundoff tolerance
+rtol = 1.0e-12 if is_double else 1.0e-10  # exact to within roundoff tolerance
 print(f"  rtol={rtol} (ignored: atol~={atol})")
 
 assert np.allclose(
--- a/examples/rfcavity/analysis_rfcavity_ref_part.py
+++ b/examples/rfcavity/analysis_rfcavity_ref_part.py
@@ -41,8 +41,10 @@ def read_time_series(file_pattern):
 def data_is_double(file_pattern):
     """Detect float precision (single vs double) from a text diagnostic's digits."""
     text = "".join(Path(f).read_text() for f in glob.glob(file_pattern))
-    mantissas = re.findall(r"\d*\.\d+", text)
-    return any(len(m.replace(".", "").strip("0")) >= 12 for m in mantissas)
+    return any(
+        len(m.replace(".", "") if e else m.replace(".", "").strip("0")) >= 12
+        for m, e in re.findall(r"(\d*\.\d+)([eE][+-]?\d+)?", text)
+    )
 
 
 rbc = read_time_series("diags/ref_particle.*")
--- a/examples/rfcavity/analysis_rfcavity_ref_part.py
+++ b/examples/rfcavity/analysis_rfcavity_ref_part.py
@@ -7,6 +7,7 @@
 
 import glob
 import re
+from pathlib import Path
 
 import numpy as np
 import pandas as pd
@@ -39,7 +40,7 @@ def read_time_series(file_pattern):
 # read reference particle data
 def data_is_double(file_pattern):
     """Detect float precision (single vs double) from a text diagnostic's digits."""
-    text = "".join(open(f).read() for f in glob.glob(file_pattern))
+    text = "".join(Path(f).read_text() for f in glob.glob(file_pattern))
     mantissas = re.findall(r"\d*\.\d+", text)
     return any(len(m.replace(".", "").strip("0")) >= 12 for m in mantissas)
 
--- a/examples/rfcavity/analysis_rfcavity_ref_part.py
+++ b/examples/rfcavity/analysis_rfcavity_ref_part.py
@@ -77,7 +77,7 @@ print("")
 print("Final Beam:")
 print(f"  s_ref={sf:e} gamma_ref={gammaf:e}")
 
-atol = 1.0e-4 if is_double else 2.0e-2  # ignored
+atol = 1.0e-4 if is_double else 2.0e-2
 print(f"  atol={atol}")
 
 assert np.allclose(
--- a/examples/rfcavity/analysis_rfcavity_ref_part.py
+++ b/examples/rfcavity/analysis_rfcavity_ref_part.py
@@ -37,7 +37,15 @@ def read_time_series(file_pattern):
 
 
 # read reference particle data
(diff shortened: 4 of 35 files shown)


examples/rfcavity/analysis_rfcavity_ref_part.py
```diff
@@ -69,7 +77,7 @@ def read_time_series(file_pattern):
 print("Final Beam:")
 print(f"  s_ref={sf:e} gamma_ref={gammaf:e}")
 
-atol = 1.0e-4  # ignored
+atol = 1.0e-4 if is_double else 2.0e-2  # ignored
```
ax3l:
```suggestion
atol = 1.0e-4 if is_double else 2.0e-2
```

examples/rfcavity/analysis_rfcavity_ref_part_hook.py
```diff
@@ -69,7 +77,7 @@ def read_time_series(file_pattern):
 print("Final Beam:")
 print(f"  s_ref={sf:e} gamma_ref={gammaf:e}")
 
-atol = 1.0e-4  # ignored
+atol = 1.0e-4 if is_double else 1.5e-2  # ignored
```
ax3l:
```suggestion
atol = 1.0e-4 if is_double else 1.5e-2
```

examples/rfcavity/analysis_rfcavity_ref_part_opt.py
```diff
@@ -69,7 +77,7 @@ def read_time_series(file_pattern):
 print("Final Beam:")
 print(f"  s_ref={sf:e} gamma_ref={gammaf:e}")
 
-atol = 1.0e-4  # ignored
+atol = 1.0e-4 if is_double else 2.0e-1  # ignored
```
ax3l:
```suggestion
atol = 1.0e-4 if is_double else 2.0e-1
```

examples/symplectic_integration/analysis_fodo_multipole.py
```diff
@@ -37,7 +37,15 @@
 
 
 # read reduced diagnostics
+def data_is_double(file_pattern):
+    """Detect float precision (single vs double) from a text diagnostic's digits."""
+    text = "".join(open(f).read() for f in glob.glob(file_pattern))
```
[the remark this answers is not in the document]
ax3l: fixed.

examples/spin_tracking/analysis_quad_spin_rbc.py
```diff
@@ -38,7 +37,15 @@
     )  # .set_index('id')
 
 
+def data_is_double(file_pattern):
+    """Detect float precision (single vs double) from a text diagnostic's digits."""
+    text = "".join(open(f).read() for f in glob.glob(file_pattern))
```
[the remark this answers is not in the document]
ax3l: fixed.

examples/rfcavity/analysis_rfcavity_ref_part_opt.py
```diff
@@ -37,7 +37,15 @@
 
 
 # read reference particle data
+def data_is_double(file_pattern):
+    """Detect float precision (single vs double) from a text diagnostic's digits."""
+    text = "".join(open(f).read() for f in glob.glob(file_pattern))
```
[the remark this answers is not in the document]
ax3l: fixed.

examples/rfcavity/analysis_rfcavity_ref_part_hook.py
```diff
@@ -37,7 +37,15 @@
 
 
 # read reference particle data
+def data_is_double(file_pattern):
+    """Detect float precision (single vs double) from a text diagnostic's digits."""
+    text = "".join(open(f).read() for f in glob.glob(file_pattern))
```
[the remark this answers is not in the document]
ax3l: fixed.

examples/rfcavity/analysis_rfcavity_ref_part.py
```diff
@@ -37,7 +37,15 @@
 
 
 # read reference particle data
+def data_is_double(file_pattern):
+    """Detect float precision (single vs double) from a text diagnostic's digits."""
+    text = "".join(open(f).read() for f in glob.glob(file_pattern))
```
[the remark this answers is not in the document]
ax3l: fixed.

examples/active_plasma_lens/analysis_APL.py
```diff
@@ -98,3 +96,10 @@
         axis=0,
         ignore_index=True,
     )  # .set_index('id')
+
+
+def data_is_double(file_pattern):
+    """Detect float precision (single vs double) from a text diagnostic's digits."""
+    text = "".join(open(f).read() for f in glob.glob(file_pattern))
```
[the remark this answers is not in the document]
ax3l: fixed.

examples/active_plasma_lens/analysis_APL.py
```diff
@@ -98,3 +97,10 @@ def read_file(file_pattern):
         axis=0,
         ignore_index=True,
     )  # .set_index('id')
+
+
+def data_is_double(file_pattern):
+    """Detect float precision (single vs double) from a text diagnostic's digits."""
+    text = "".join(Path(f).read_text() for f in glob.glob(file_pattern))
+    mantissas = re.findall(r"\d*\.\d+", text)
+    return any(len(m.replace(".", "").strip("0")) >= 12 for m in mantissas)
```
EZoni: I think the function `data_is_double` is duplicated 6 times:

- examples/active_plasma_lens/analysis_APL.py
- examples/rfcavity/analysis_rfcavity_ref_part.py
- examples/rfcavity/analysis_rfcavity_ref_part_hook.py
- examples/rfcavity/analysis_rfcavity_ref_part_opt.py
- examples/spin_tracking/analysis_quad_spin_rbc.py
- examples/symplectic_integration/analysis_fodo_multipole.py

Do you think it would be possible to put this in one shared utility module and import it from there?

examples/spin_tracking/analysis_quad_spin_rbc.py
```diff
@@ -116,7 +124,7 @@ def read_file(file_pattern):
 print("")
 print(f"Spin moment consistency condition = {condition:e}")
 
-atol = 1.0e-12  # machine precision
+atol = 1.0e-12 if is_double else 1.0e-3  # machine precision
```
EZoni: From NumPy I get the following values:
```
Python 3.12.3 (main, Mar 23 2026, 19:04:32) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
>>> np.finfo(np.float32).eps
1.1920929e-07
>>> np.finfo(np.float64).eps
2.220446049250313e-16
>>> 
```

I'm not convinced it's accurate to describe `1e-12` and `1e-3` as "machine precision" for the double and single precision branches, as the comment `# machine precision` suggests.

examples/charge_sign/analysis_charge_sign.py
```diff
@@ -80,7 +81,7 @@ def get_moments(beam):
 )
 
 atol = 0.0  # ignored
-rtol = 1.0e-11  # exact to within roundoff tolerance
+rtol = 1.0e-11 if is_double else 1.0e-4  # exact to within roundoff tolerance
```
EZoni: I believe https://github.com/BLAST-ImpactX/impactx/pull/1479#discussion_r3350998177 applies here too.

examples/reversibility/analysis_zero_bend_inverse.py
```diff
@@ -80,7 +81,7 @@ def get_moments(beam):
 )
 
 atol = 0.0  # ignored
-rtol = 1.0e-12  # exact to within roundoff tolerance
+rtol = 1.0e-12 if is_double else 1.0e-10  # exact to within roundoff tolerance
```
EZoni: I believe https://github.com/BLAST-ImpactX/impactx/pull/1479#discussion_r3350998177 applies here too.

examples/active_plasma_lens/analysis_APL.py
ax3l: Yes, we cannot share it right now because we do not want that application (exe) based build with `-DImpactX_PYTHON=OFF` depend in an `import impactx.submodule`.

Just setting PYTHONPATH would also be a bit tricky, because analysis scripts are also user-facing, so making them self-contained is good.

So I do not have a great sharing idea right now.
Need to be refactored another time together with other duplications like the existing moment calcs.

examples/spin_tracking/analysis_quad_spin_rbc.py
ax3l: I think the comment is misleading and should be removed. The whole if/else "is about machine precision", but it is not set "to machine precision", because it also depends on numerical order when comparing to theory.

examples/active_plasma_lens/analysis_APL.py
EZoni: Understood, thanks. 

I have one more question about the function itself. 

If the string is printed as `1.2300`, we count `3`, instead of `5`, and compare to `12`. Same if the string is printed as `0.012300`:
```
>>> len(re.findall(r"\d*\.\d+", "1.2300")[0].replace(".", "").strip("0"))
3
>>> len(re.findall(r"\d*\.\d+", "0.012300")[0].replace(".", "").strip("0"))
3
>>>
```
Aren't trailing zeros after a decimal significant, if printed? 

Maybe I'm missing context with respect to what text output this refers to.

ax3l: This is a bit of a hack and uses the ASCII print logic here to identify a single precision run that only created ASCII output (no openPMD data types to check):
- https://github.com/BLAST-ImpactX/impactx/blob/26.05/src/diagnostics/DiagnosticOutput.cpp#L95
- https://en.cppreference.com/cpp/types/numeric_limits/max_digits10

A single precision run will print less digits, intentionally, already before this PR. The threshold (12) for detection is set to have some wiggle room (DP: 15-16, SP: ~7 significant digits)

For pure executable + input file runs without openPMD output, I found no better indicator right now. We could alternatively try to find the output file (involves one more file) and grep it there or do hacky inspections of the binary that was run. Using the output we read itself as the indicator seemed most self-consistent to me.

I avoided this detection mechanism for all tests that have Python bindings available and/or all tests that create an openPMD file.

ax3l: We can make the doc string of the function more detailed, but because it is pure boiler-plate that is duplicated in a few tests, I try to keep the doc part to a one-liner right now. But if two lines help to clarify it then it might be worth it.

EZoni: Okay, so basically the ASCII output won't fall into this potentially problematic case of decimal trailing zeros, because it will print `1.23` and `0.0123` instead of `1.2300` and `0.012300`, respectively. Hope I understood correctly. 

One last question. Is this function guaranteed to be evaluated always on real values? I think the Python code as is crashes for integers, e.g., `12300`, because of the regex expression:
```
>>> len(re.findall(r"\d*\.\d+", "12300")[0].replace(".", "").strip("0"))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
IndexError: list index out of range
>>>
```

ax3l: Good question, let's check.

I made a compiler explorer for the output formatting: https://godbolt.org/z/eerfY8WcT
```C++
#include <iostream>
#include <limits>
#include <iomanip>

int main ()
{
    double d0 = 1. / 3.;
    double d1 = 1.234;
    double d2 = 123.;
    double d3 = 0.0123;
    std::cout << std::setprecision(std::numeric_limits<double>::max_digits10);
    std::cout << "double:\n "
              << d0 << "\n "
              << d1 << "\n "
              << d2 << "\n "
              << d3 << "\n\n";

    float f0 = 1.f / 3.f;
    float f1 = 1.234f;
    float f2 = 123.f;
    float f3 = 0.0123f;
    std::cout << std::setprecision(std::numeric_limits<float>::max_digits10);
    std::cout << "float:\n "
              << f0 << "\n "
              << f1 << "\n "
              << f2 << "\n "
              << f3 << "\n";
}
```
yields:
```
double:
 0.33333333333333331
 1.234
 123
 0.0123

float:
 0.333333343
 1.23399997
 123
 0.0122999996
```

And this is what Python returns:
```py
import re

def f(text):
    mantissas = re.findall(r"\d*\.\d+", text)
    return any(len(m.replace(".", "").strip("0")) >= 12 for m in mantissas)

# all double
print(f("0.33333333333333331"))  # True
print(f("1.234"))  # False
print(f("123"))  # False
print(f("0.33333333333333331, 1.234, 123"))  # True
```

So the current detector is not trying to prove the type from _every value_. It only needs at least one printed decimal value with enough significant digits.
Integer-looking values such as `123` are ignored by the regex, and that _does not crash_ in the actual function (int case: `any([]) == False`). This is how we already skip over the `step` column in the file.

For these diagnostics that should be OK: the files may contain further integer-looking real values, but they also contain many computed `amrex::ParticleReal` columns. If _at least one of those_ (in any row) exposes the `max_digits10` precision, double precision is detected; otherwise we fall back to the single-precision tolerance path (which does not crash, it just will apply looser tolerance).

But let's make this more water tight for the corner case where there is absolutely no computed value in there (admittedly, we always compute some columns) by always using scientific notation and padding the mantissa with zeros for the corner cases, also looks a bit nicer in the CSV files when opened.

ax3l: Updated! :)

EZoni:
Some tests (both runtime and Ruff's static analysis) complain about `Config` being undefined. However, I don't understand the issue, I do see `Config` being imported from `impactx` in tests/python/test_dipedge_ref.py...

```
=================================== FAILURES ===================================
______________________ test_dipedge_modify_ref_part_false ______________________

    def test_dipedge_modify_ref_part_false():
        """
        This tests the application of longitudinal particle boundary conditions.
        """
        edge_angle = math.pi / 8.0
        rc = 10.0
        g = 1.0e-2
    
        K0 = np.pi**2 / 6.0
        c2 = (1.0 / np.cos(edge_angle)) ** 3 * np.sin(edge_angle) / 2.0 * g**2 / rc**2 * K0
        c3 = (1.0 / np.cos(edge_angle)) ** 2 * g**2 / rc * K0
        c5 = np.tan(edge_angle) / (2.0 * rc)
        c13 = 0.0
    
        rbc, ref = _track_particle_dipedge(False, edge_angle, g, rc)
    
        x_shift = -c3
        px_shift = c13 - c2 - c3 * c5
        t_shift = -c3 * px_shift / ref.beta
    
        # access centroid data (= single particle orbit)
        vec_part = np.array(
            [
                rbc["mean_x"],
                rbc["mean_px"],
                rbc["mean_y"],
                rbc["mean_py"],
                rbc["mean_t"],
                rbc["mean_pt"],
            ]
        )
        vec_part_pred = [x_shift, px_shift, 0.0, 0.0, t_shift, 0.0]
    
        # access reference particle
        vec_ref = np.array([ref.x, ref.px, ref.y, ref.py, ref.t])
        vec_ref_pred = [0, 0, 0, 0, 0]
        vec_ref_on_shell = (
            (ref.px) ** 2 + (ref.py) ** 2 + (ref.pz) ** 2 - (ref.pt) ** 2 + 1.0
        )
    
>       atol = 1.0e-12 if Config.precision != "SINGLE" else 1.0e-6
                          ^^^^^^
E       NameError: name 'Config' is not defined

/Users/runner/work/impactx/impactx/tests/python/test_dipedge_ref.py:119: NameError
______________________ test_dipedge_modify_ref_part_true _______________________

    def test_dipedge_modify_ref_part_true():
        """
        This tests the application of longitudinal particle boundary conditions.
        """
        edge_angle = math.pi / 8.0
        rc = 10.0
        g = 1.0e-2
    
        K0 = np.pi**2 / 6.0
        c2 = (1.0 / np.cos(edge_angle)) ** 3 * np.sin(edge_angle) / 2.0 * g**2 / rc**2 * K0
        c3 = (1.0 / np.cos(edge_angle)) ** 2 * g**2 / rc * K0
        c5 = np.tan(edge_angle) / (2.0 * rc)
        c13 = 0.0
    
        rbc, ref = _track_particle_dipedge(True, edge_angle, g, rc)
    
        x_shift = -c3
        px_shift = c13 - c2 - c3 * c5
        t_shift = -c3 * px_shift / ref.beta
    
        # access centroid data (= single particle orbit)
        vec_part = np.array(
            [
                rbc["mean_x"],
                rbc["mean_px"],
                rbc["mean_y"],
                rbc["mean_py"],
                rbc["mean_t"],
                rbc["mean_pt"],
            ]
        )
        vec_part_pred = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
    
        # access reference particle
        vec_ref = np.array([ref.x, ref.px, ref.y, ref.py, ref.t])
        vec_ref_pred = [x_shift, px_shift, 0, 0, t_shift]
        vec_ref_on_shell = (
            (ref.px) ** 2 + (ref.py) ** 2 + (ref.pz) ** 2 - (ref.pt) ** 2 + 1.0
        )
>       atol = 1.0e-12 if Config.precision != "SINGLE" else 1.0e-6
                          ^^^^^^
E       NameError: name 'Config' is not defined

/Users/runner/work/impactx/impactx/tests/python/test_dipedge_ref.py:164: NameError
```

ax3l:
Yup saw! Needed a rebase because the `merge` in tests here removed an import we needed.

EZoni (approved):

PR #1488, opened 2026-06-03
Simplify Adding Particles

ax3l:
Adding particles from Python arrays was quite verbose and boilerplate-heavy.

This simplifies the existing interfaces to directly accept NumPy or CuPy-style arrays.
All examples and tests have been updated.

Canonical example: https://impactx--1488.org.readthedocs.build/en/1488/usage/examples/initialize_from_array/README.html

Tested locally on GPU: all green and the next function signature is backwards compatible with the old one.


ax3l:
Noise, will try to reduce in #1489

EZoni (approved):
Looks good to me.

PR #1489, opened 2026-06-03
CI: Reduce CodSpeed Benchmark Noise

ax3l:
The cachegrind cache model is sensitive to code/data layout, so unrelated edits that relink the binary shift hot-loop alignment and perturb simulated cache misses. On the cheapest elements (e.g., drift) this showed up as ~5-6% phantom regressions in cache modeling with identical instruction counts (e.g., from #1488).

Lower this layout-driven floor on the CodSpeed simulation:

- Double the simulated particle count (10k -> 20k) to amplify the steady-state push relative to the roughly fixed cache-model floor.
- Pin hot loops to fixed cache-line boundaries via `-falign-functions` and `-falign-loops` so unrelated edits no longer move their alignment.

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -115,7 +115,10 @@ jobs:
     if: github.event.pull_request.draft == false
     env:
       CMAKE_GENERATOR: Ninja
-      CXXFLAGS: "-Werror -march=x86-64-v3 -mtune=generic"
+      # -falign-functions/-falign-loops pin hot loops to fixed cache-line
+      # boundaries so unrelated edits do not shift their alignment and perturb
+      # the cachegrind cache model (CodSpeed simulation noise).
+      CXXFLAGS: "-Werror -march=x86-64-v3 -mtune=generic -falign-functions=64 -falign-loops=32"
       OMP_NUM_THREADS: 2
     steps:
     - name: system info
--- a/tests/python/test_benchmark_elements.py
+++ b/tests/python/test_benchmark_elements.py
@@ -23,7 +23,7 @@ from impactx import ImpactX, Map6x6, distribution, elements, twiss
 if os.environ.get("IS_CODESPEED_CPU_SIMULATION") == "1":
     # https://codspeed.io/docs/instruments/cpu/index
     rounds = 1
-    npart = 10_000
+    npart = 20_000
 else:
     rounds = 5
     npart = 1_000_000  # increase this to >10M or even 100M to avoid L1/L2/L3 cache effects on some hardware.


ax3l:
Looking at #1488 again, this fix did not help and made things slightly worse for noise levels.

Claude analysis:

> **Update (2026-06-03) — corrections:** #1489 has been **reverted** on `development` (`dda1aea1b`), so `npart` is back to 10k and the `-falign-*` flags are gone. Two corrections to the original write-up below:
> 1. **Instruction-count gating is not possible** — CodSpeed exposes no metric selection (measurement modes are only `Simulation` / `Memory` / `WallTime`; "instrumentation" is a deprecated alias for `Simulation`, which includes the cache model). pytest-codspeed 4.3.0 has no Ir-only option.
> 2. **The per-benchmark regression threshold is dashboard-only** (benchmark page → Actions → custom threshold, or project Settings) — not a repo config file or pytest marker.
>
> Also, with the revert the noise **floor is back to ~5–6%** (the ~9% figure was only under the reverted 20k config). The durable fix is a CodSpeed **dashboard** per-benchmark threshold (~10%). Sections below are corrected accordingly.

---

# CodSpeed Simulation Noise: Investigation & Verdict

Analysis of the phantom ~5–9% "regressions" on the cheapest drift+spin
micro-benchmarks in `tests/python/test_benchmark_elements.py` under
CodSpeed's `Simulation` (cachegrind) mode.

## TL;DR

- The "regressions" on `test_Drift[spin]` and `test_ExactDrift[spin]` are
  **not real** — they are a deterministic artifact of CodSpeed's simulated
  cache model reacting to **binary/data layout**, not to any change in the
  drift or spin push code.
- The signature is unmistakable: across every comparison the **instruction
  count and memory-access count are byte-identical**, and **100% of the delta
  is in the simulated `caches` component**.
- The mitigation attempted in **PR #1489** (double particle count 10k→20k +
  `-falign-functions=64 -falign-loops=32`) **did not work and partly
  backfired**: the same two benchmarks regressed again, *larger* than before.
- Root cause is **data-address conflict-miss aliasing** (the SoA component
  arrays re-map into different cache sets when a relink moves their heap
  addresses). Code-alignment flags cannot touch this; doubling particles made
  it worse.
- The only quantity that stayed reproducible throughout is the **instruction
  count** — but CodSpeed exposes no way to gate on it (no metric selection).
  **#1489 has been reverted** (`dda1aea1b`); the durable fix is a CodSpeed
  **dashboard** per-benchmark regression threshold. (Both instruction-count
  gating and thresholds are platform-side, not in-repo — see "Verified
  constraints".)

## Background

CodSpeed `Simulation` mode does not time the wall clock. It runs the
benchmark under a cachegrind-style CPU simulator that counts executed
instructions and feeds every memory access through a modeled cache hierarchy,
then converts the result to a synthetic time. The reported self-time of a
function decomposes into three components:

- `instructions` — count of machine instructions executed
- `memory` — modeled cost of memory accesses (bulk streaming)
- `caches` — modeled L1/L2/LL miss penalty

`instructions` and `memory` depend only on *what* code runs and *how many*
accesses it issues. `caches` additionally depends on the **addresses** of
those accesses, because the simulator maps addresses to cache sets and models
eviction. Two accesses whose addresses collide in the same set evict each
other. That makes `caches` sensitive to code/data layout — which changes on
every relink — while `instructions` does not.

The benchmark harness (under `IS_CODESPEED_CPU_SIMULATION=1`) runs a single
`el.push` over `npart` particles, `nslice=1`, `rounds=1`, with
`OMP_NUM_THREADS=1`. The element push is ~95% of the measured self-time.

## Evidence

All numbers below are the self-time breakdown of the hot kernel
`impactx::elements::mixin::detail::push_all_particles<Drift>` for
`test_Drift[spin]`.

### Round 1 — original report (PR #1488, 10k particles, no align flags)

| Component    | BASE (`70d2124`) | HEAD (`da6d609`) | Δ        |
|--------------|-----------------:|-----------------:|---------:|
| instructions | 56.6 µs          | 56.6 µs          | **0**    |
| memory       | 349.6 µs         | 349.4 µs         | **~0**   |
| caches       | 17.5 µs          | 43.8 µs          | **+26.3 µs** |
| total self   | 423.8 µs         | 449.7 µs         | +25.9 µs |

The entire +26 µs (~5.8% of the benchmark) is the `caches` component. PR #1488
only renamed a pybind symbol (`add_n_particles` → `_add_n_particles`) and
added a Python wrapper — nothing on the drift/spin code path. CodSpeed flagged
an "Environment Differences" warning (changed `build_id`s), confirming the
binary was relinked.

### Round 2 — after the #1489 mitigation (20k particles, align flags)

Controlled experiment: **#1488 rebased onto merged #1489** (head) vs the
**post-merge `development`** run (base). Both builds now have 20k particles
and `-falign-functions=64 -falign-loops=32`. The only difference between them
is the same #1488 symbol rename + a relink.

| Component    | BASE (`8b0cb98`) | HEAD (#1488 rerun) | Δ        |
|--------------|-----------------:|-------------------:|---------:|
| instructions | 112.8 µs         | 112.8 µs           | **0**    |
| memory       | 696.6 µs         | 696.5 µs           | **~0**   |
| caches       | 35.7 µs          | 118.6 µs           | **+82.9 µs** |
| total self   | 845.1 µs         | 927.8 µs           | +82.7 µs |

CodSpeed reported `test_Drift[spin]` −8.64% and `test_ExactDrift[spin]`
−6.67% — the same two benchmarks, same phantom signature, **worse** than
before.

### The two rounds side by side

| Config                | cache swing (base→head) | relative |
|-----------------------|------------------------:|---------:|
| 10k, no align flags   | +26 µs                  | ~5.8%    |
| 20k, with align flags | +83 µs                  | ~8.6%    |

Instructions and memory scaled exactly ×2 with the particle doubling (56.6→
112.8, 349.6→696.6). The clean-base cache term also ~doubled (17.5→35.7,
×2.04). But the *swing* grew ×3.2, so relative noise went **up**.

## Why the mitigation failed

The #1489 changes targeted the wrong mechanism:

1. **`-falign-functions` / `-falign-loops` only align code.** The instruction
   count was already byte-identical in every comparison — the hot loop's code
   was never the variable. These flags move function entry points and loop
   tops; they cannot change where heap-allocated data lands.

2. **The noise is data-side conflict-miss aliasing.** The Drift+spin push
   streams ~11 separate SoA component arrays (x, y, t, px, py, pt, spin
   components, …). Whether those arrays evict each other in the modeled cache
   depends on their **heap base addresses modulo the cache geometry**. A relink
   (the #1488 symbol rename shifting the symbol table and downstream
   allocations) re-maps the arrays into different cache sets, changing the
   modeled miss count for an otherwise identical access stream.

3. **Doubling particles amplified it.** A larger working set has more
   conflict-missing cache lines whose set-mapping is address-dependent, so the
   absolute swing grew (×3.2) — and CI time and the misleading "41 regressions"
   PR comment came along with it.

## What actually works

The diagnosis points at one robust signal (instruction count) and one
pragmatic knob (per-benchmark threshold). **Both turned out to be CodSpeed
platform features, not in-repo configuration** — see "Verified constraints"
below.

1. **Gate on instruction count (Ir).** The instruction count was byte-identical
   in every comparison at both 10k and 20k, so it is fully immune to the
   data-layout aliasing. It would be the ideal gate — **but CodSpeed does not
   expose metric selection.** The measurement modes are `Simulation` / `Memory`
   / `WallTime` (and "instrumentation" is only a deprecated alias for
   `Simulation`, which *includes* the cache/memory model). pytest-codspeed
   4.3.0 has no option to report or gate on Ir alone. **Not implementable.**

2. **Raise the regression threshold** on the sub-millisecond drift benches
   above the layout floor. **Post-revert (back to 10k) the floor is ~5–6%**
   again, not the ~9% seen under the reverted 20k config. This *is* available,
   but **only via the CodSpeed dashboard** — per benchmark (benchmark page →
   Actions menu → custom threshold) or project-wide (project Settings). It is
   not a repo config file or a pytest marker, so it cannot be committed. This
   is the recommended durable fix; prefer per-benchmark so the heavy elements
   keep tight detection.

3. **Deterministic SoA allocation** (deeper fix): pad/align each component
   array to a fixed cache-set offset so their relative aliasing does not move
   with the binary. This is an AMReX arena change — heavy, and out of scope for
   a CI-noise fix.

Note: the repeat-loop amplification idea (loop `el.push` K× over the same 10k
working set) would raise the signal without enlarging the working set, but it
does not address the underlying data-address aliasing either.

## Verified constraints (pytest-codspeed 4.3.0 + CodSpeed docs)

- **No in-repo metric selection.** Inspected every plugin CLI option
  (`--codspeed`, `--codspeed-mode`, `--codspeed-warmup-time`,
  `--codspeed-max-time`, `--codspeed-max-rounds`) and the benchmark marker
  options (`group`, `min_time`, `max_time`, `max_rounds` — the last three
  walltime-only). None select the reported metric. Instruction-count-only
  gating is not a feature.
- **No in-repo threshold.** Regression thresholds are a dashboard setting
  (project Settings, or per-benchmark via the benchmark page Actions menu).
  No `codspeed.yml`, env var, or marker controls it.

## Recommendation / status

- **PR #1489 reverted** on `development` (commit `dda1aea1b`) — restores
  `npart = 10_000` and removes the `-falign-*` flags. The two levers it added
  did not help and made the noise worse.
- **Durable fix is a CodSpeed dashboard action** (owner-only): set a custom
  per-benchmark regression threshold (~10%, comfortably above the post-revert
  ~5–6% 10k floor) on `test_Drift[spin]` and `test_ExactDrift[spin]` (and
  `test_ChrDrift[spin]` if it trips). There is no in-repo code change that
  fixes this.

## Run references (BLAST-ImpactX/impactx)

| Run ID                     | Commit    | What                                  |
|----------------------------|-----------|---------------------------------------|
| `6a1f5609c7e4f5538fa2a08e` | `70d2124` | development base, 10k, no align       |
| `6a1f7ca75acf38ac1388910e` | `da6d609` | #1488 original head, 10k, no align    |
| `6a1f90f361cd3010271b3da7` | `d24c7b7` | #1489, 20k + align                    |
| `6a1f9ec520b8691f762e1150` | `8b0cb98` | development post-merge #1489, 20k+align|
| `6a1f9f76d67ed1398585be81` | `f37af46` | #1488 rebased rerun, 20k + align      |

(#1489 subsequently reverted on `development` as `dda1aea1b`.)

ax3l:
I pushed the global reporting threshold now up to 7% (from 5%)

PR #1491, opened 2026-06-04
2.5D Gaussian SC: Long Scale Default & Name

ax3l:
Rename options in #1473 after discussion in our weekly meeting ([summary](https://github.com/BLAST-ImpactX/impactx/pull/1473#discussion_r3358751371)).
Avoid confusion with transverse SC dynamics (which is open BC).

Default to $$6 \cdot \gamma \cdot \sigma_z$$, which is a good empirical value by @qianglbl for comparisons to 3D.

### DIFF
--- a/src/python/ImpactX.cpp
+++ b/src/python/ImpactX.cpp
@@ -358,18 +358,20 @@ void init_ImpactX (py::module& m)
             },
             "Initial region for computing the integrals (default: ``0.01``)."
         )
-        .def_property("space_charge_gauss_pipe_radius",
+        .def_property("space_charge_gauss_long_scale",
             [](ImpactX & /* ix */) {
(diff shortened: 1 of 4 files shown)


src/python/ImpactX.cpp
```diff
@@ -358,18 +358,21 @@ void init_ImpactX (py::module& m)
             },
             "Initial region for computing the integrals (default: ``0.01``)."
         )
-        .def_property("space_charge_gauss_pipe_radius",
+        .def_property("space_charge_gauss_long_scale",
             [](ImpactX & /* ix */) {
-                return detail::get_or_throw<int>("algo.space_charge", "gauss_pipe_radius");
```
ax3l, line 363: The `int` here was a bug. Fixed now, too :)

qianglbl (approved):

PR #1492, opened 2026-06-04
Fix charge accounting in Poisson solver

### LINKED ISSUE #1447: `PoissonSolve` descales `rho` after every space-charge solve
`<claude-ish-text>`

## Summary

`src/particles/spacecharge/PoissonSolve.cpp:153-157` unconditionally multiplies `rho` by `-ε₀` on every level after `ablastr::fields::computePhi` returns. In the current ablastr, `computePhi` already restores `rho` to its original value on **both** the MLMG and IGF paths, so this post-loop overwrite is wrong. 

## Code

`src/particles/spacecharge/PoissonSolve.cpp:153-157`:

```cpp
// fix side effect on rho from previous call
for (int lev=0; lev<=finest_level; lev++) {
    using namespace ablastr::constant::SI;
    rho[lev].mult(-1._rt * epsilon_0);
}
```

## Details
In the current ablastr `computePhi` (`warpx/Source/ablastr/fields/PoissonSolver.H`), `rho` is symmetrically scaled and un-scaled on the MLMG path:

- **Line 288** (entering MLMG): `rho[lev]->mult(-1._rt / ablastr::constant::SI::epsilon_0);`
- **Line 446** (after MLMG solve, still inside the per-level loop): `rho[lev]->mult(-ablastr::constant::SI::epsilon_0);  // Multiply rho by epsilon again`

The IGF path (`PoissonSolver.H:277-284`) `continue`s before reaching either of those lines, so it never touches `rho` at all.

Net effect of `computePhi` on `rho`:
- IGF path: **unchanged**
- MLMG path: **unchanged** (scale and un-scale cancel exactly)

Then ImpactX runs the post-loop at `PoissonSolve.cpp:153-157`, which multiplies `rho` by `-ε₀` on every level for both paths.

`</claude-ish-text>`

## Reproducer
(boilerplate removed)

```python
sim.track_particles()                # one SC slice
R_post  = gather(sim.rho(0))         # post-PoissonSolve state
sim.deposit_charge()                 # re-deposit on same geometry
R_redep = gather(sim.rho(0))         # clean reference

# elementwise ratio on cells with appreciable magnitude:
nz = abs(R_redep) > R_redep_max * 1e-6
median_ratio = numpy.median(R_post[nz] / R_redep[nz])
# observed: -8.854188e-12 (== -EPS0) on both fft and multigrid
```

Deleting `PoissonSolve.cpp:153-157` seems ok

nikitakuklev:
Removes the obsolete post-solve rescaling of rho. Fixes #1447.

### DIFF
--- a/src/particles/spacecharge/PoissonSolve.cpp
+++ b/src/particles/spacecharge/PoissonSolve.cpp
@@ -12,7 +12,6 @@
 #include "initialization/Algorithms.H"
 #include "particles/ChargeDeposition.H"
 
-#include <ablastr/constant.H>
 #include <ablastr/fields/PoissonSolver.H>
 
 #include <AMReX_BLProfiler.H>
@@ -155,12 +154,6 @@ namespace impactx::particles::spacecharge
             eb_farray_box_factory
             */
         );
-
-        // fix side effect on rho from previous call
-        for (int lev=0; lev<=finest_level; lev++) {
-            using namespace ablastr::constant::SI;
-            rho[lev].mult(-1._rt * epsilon_0);
-        }
 
         // We may need to copy phi from phi_2d
         if (space_charge == SpaceChargeAlgo::True_2D || space_charge == SpaceChargeAlgo::True_2p5D) {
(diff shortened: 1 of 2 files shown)


ax3l (approved):
Thank you for the fix and test!

PR #1493, opened 2026-06-05
Spin-Tracking for Apertures

ax3l:
Our thin apertures do not affect spin tracking. Expose them to our spin tracking loops by forwarding directly to the phase space push.


carlandreaslindstrom (commented):
Looks good to me!

PR #1494, opened 2026-06-08
Catch Missing Gyromagnetic Anomaly

ax3l:
With the Python APIs, it is possible to forget to set the gyromagnetic anomaly. The default is zero, which no physical particle has.
This now raises an error and documents clearner. All new examples use the `set_species` API on the reference particle, so this should not cause too many surprises (and if triggered, it advises cleanly).

First seen by Carl Lindstroem.

### DIFF
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -1286,6 +1286,8 @@ Currently, the implementation of spin tracking is a work in progress, and this f
 
   Whether to track particle spin.
 
+  Spin tracking uses the gyromagnetic anomaly of the reference particle, which is set together with the particle species (see ``<beam>.particle``).
+
 
 .. _running-cpp-parameters-parser:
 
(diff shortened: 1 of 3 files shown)


cemitch99:
@ax3l Just a correction:  Setting the gyromagnetic anomaly to zero does not _generally_ produce the same result as disabling spin tracking--only in _special cases_.  In the general case, the spin should evolve when the gyromagnetic anomaly value is zero.  Of course, the result obtained will differ from the result obtained with nonzero gyromagnetic anomaly.

cemitch99 (approved):
Thanks!

ax3l:
@cemitch99 thanks for spotting this, I edited now the PR description to read w/o the 2nd half of the sentence:
> With the Python APIs, it is possible to forget to set the gyromagnetic anomaly. The default is zero, which no physical particle has ~and which (sometimes) results in the same physics as disabling spin tracking~.

PR #1498, opened 2026-06-10
Shink RefPart by 456 Bytes

### LINKED ISSUE #1423: Keep the Reference Particle Small
We copy the reference particle often around, e.g., to GPU, under the assumption that it stays small.

With PR #222 for the symplectic integrator in s-dependent elements (like `RFCavity`) and now more with #1264 for the equivalent in spin tracking, we introduce small matrices in the reference particle:
```C++
        amrex::SmallMatrix<amrex::ParticleReal, 3, 6, amrex::Order::F, 1> spin_coupling{}; ///< linearized spin-orbit coupling matrix
        amrex::SmallMatrix<amrex::ParticleReal, 3, 1, amrex::Order::F, 1> spin_rotation_vector{}; ///< reference spin rotation vector
```

We need to find a refactoring that makes sure we only "pay" for the allocation & copy & registers of these members when we truly need them.

ax3l:
Move out the maps to the 3 elements that use them. Accelerate all other elements.

An alternative implemenetation could move this to the particle container and only copy it in on-demand for phase-space or particle push, which would optimize kernel startup time/kernel parameter bank usage for RFCavity, SoftQuad, SoftSol a bit more.

- [x] fix #1423
- [x] rebase after #1497

## GPU footprint change

GPU kernel parameter bank (`CONSTANT[0]`):

  Measured with `cuobjdump --dump-resource-usage` (RTX A2000, sm_86, double precision):

  | push kernel              | param bank before | after | Δ        |
  |--------------------------|------------------:|------:|---------:|
  | Drift (nospin)           | 1088 B            | 632 B | −456 B   |
  | ExactQuad (nospin)       | 1144 B            | 688 B | −456 B   |
  | ExactCFbend (nospin)     | 1168 B            | 712 B | −456 B   |
  | ExactCFbend (spin)       | 1192 B            | 736 B | −456 B   |
  | ExactMultipole (spin)    | 1192 B            | 736 B | −456 B   |

Every non-matrix element kernel sheds exactly the 456 bytes of matrices.

Bonus: ExactCFbend spin's local-memory spill dropped 1048 → 592 B (−456), since the bloated RefPart is no longer copied into that register-pressured kernel. The 3 matrix elements (RFCavity/SoftSol/SoftQuad) retain ~1.1 KB.

## Performance Impact vs. Headroom Improvement

For compute-bound kernels this makes no difference on GPU.
The payoff is for short, launch/overhead-dominated elements.

The kernel parameter bank (constant memory), which is read once and cached, does (if unread) not affect occupancy (occupancy is driven by registers/shared memory, not param size).
Also, there is a ~4KB limit on this memory bank before spills, no need to over-use it.

Result: register/parameter-bank footprint reduced (and a spill removed in the heaviest spin kernel). Throughput is performance-neutral on GPU for tested elements (incl. Drift).

### DIFF
--- a/src/elements/RFCavity.H
+++ b/src/elements/RFCavity.H
@@ -220,13 +220,13 @@ namespace impactx::elements
             T_Real & AMREX_RESTRICT py,
             T_Real & AMREX_RESTRICT pt,
             [[maybe_unused]] T_IdCpu const & AMREX_RESTRICT idcpu,
-            RefPart const & AMREX_RESTRICT refpart
+            [[maybe_unused]] RefPart const & AMREX_RESTRICT refpart
         ) const
         {
             using namespace amrex::literals; // for _rt and _prt
 
             // get the linear map
-            amrex::SmallMatrix<amrex::ParticleReal, 6, 6, amrex::Order::F, 1> const R = refpart.map;
+            amrex::SmallMatrix<amrex::ParticleReal, 6, 6, amrex::Order::F, 1> const R = m_map;
 
             // symplectic linear map for the RF cavity is computed using the
             // Hamiltonian formalism as described in:
@@ -269,10 +269,10 @@ namespace impactx::elements
             amrex::ParticleReal const sedge = refpart.sedge;
 
             // initialize linear map (deviation) values
-            refpart.map = decltype(refpart.map)::Identity();
+            m_map = decltype(m_map)::Identity();
 
             // initialize the spin-orbit coupling matrix
-            refpart.spin_coupling = {};
+            m_spin_coupling = {};
 
             // length of the current slice
             amrex::ParticleReal const slice_ds = m_ds / nslice();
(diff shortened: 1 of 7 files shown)


src/elements/RFCavity.H
```diff
@@ -644,6 +644,15 @@ namespace impactx::elements
         amrex::ParticleReal const * m_sin_h_data = nullptr; //! non-owning pointer to host sine coefficients
         amrex::ParticleReal const * m_cos_d_data = nullptr; //! non-owning pointer to device cosine coefficients
         amrex::ParticleReal const * m_sin_d_data = nullptr; //! non-owning pointer to device sine coefficients
+
+        // Reference-trajectory linearization around the reference particle.
+        // Computed during the reference-particle push and consumed during the
+        // particle push. Kept on the element (not on RefPart) so only the few
+        // elements that need it pay the storage/copy/register cost (issue #1423).
+        // mutable: written by the const reference push before the element is copied
+        // into the particle-push functor.
```
ax3l, line 652: This comment (3x) can be simplified

ax3l: simplified

cemitch99 (approved):

PR #1499, opened 2026-06-10
Alignment: Short-Cut

ax3l:
For our (mis)alignment mixing classes and in/out translation and rotation into element frames, we can skip the logic for the (common) case that elements are perfectly aligned.

This also improves comparisons to other codes that do not even support misalignment.

Local GPU benchmarks (NVIDIA RTX A2000 8GB laptop GPU) on perfectly aligned elements show 0-38% speedup for DP and are neutral for SP (tested on: Sol, Sbend, CFbend).

<details>
<summary>GPU measurements</summary>

  GPU wall-clock gain from #1499, measured as **ON (#1499)** vs **OFF 
  (pre-#1499, always shift/rotate)**.
  NVIDIA RTX A2000 (sm_86), standalone app, each element isolated (fresh beam, 
  20 pushes/element),
  200k particles, min-of-4, TinyProfiler kernel time, aligned elements. 
  Negative = #1499 faster.
  (CodSpeed reports CPU instruction counts; this is GPU wall-clock.)

  | Benchmark | GPU DP | GPU SP |
  | --- | ---: | ---: |
  | Aperture[spin] | −8.3% | +16.9% ⚠ |
  | Aperture[nospin] | −2.5% | +2.0% |
  | Buncher | −16.1% | −11.0% |
  | CFbend[spin] | −13.1% | −2.9% |
  | CFbend[nospin] | −6.3% | −0.1% |
  | ChrDrift[spin] | −31.2% | −23.2% |
  | ChrDrift[nospin] | −23.1% | −9.1% |
  | ChrPlasmaLens[spin] | −6.3% | −0.8% |
  | ChrPlasmaLens[nospin] | −7.8% | +1.0% |
  | ChrQuad[spin] | −4.2% | −1.3% |
  | ChrQuad[nospin] | −4.4% | +1.2% |
  | ChrAcc | −8.5% | −0.7% |
  | ConstF | −2.8% | +0.7% |
  | DipEdge | −19.3% | −10.7% |
  | Drift[spin] | −31.3% | −21.2% |
  | Drift[nospin] | −9.0% | −3.5% |
  | ExactCFbend[spin] | −0.3% | +17.7% ⚠ |
  | ExactCFbend[nospin] | −0.2% | +25.8% ⚠ |
  | ExactDrift[spin] | −38.5% | −24.6% |
  | ExactDrift[nospin] | −31.2% | −8.1% |
  | ExactMultipole | −3.4% | +6.6% ⚠ |
  | ExactQuad | −0.3% | +1.0% |
  | ExactSbend | −12.5% | −1.8% |
  | Kicker | −31.5% | −21.8% |
  | LinearMap | −27.0% | +2.3% |
  | Multipole | −34.9% | −9.8% |
  | NonlinearLens | −9.9% | −11.2% |
  | PlaneXYRot | −15.3% | −2.8% |
  | PolygonAperture[spin] | −1.5% | −15.9% ⚠ |
  | PolygonAperture[nospin] | +2.2% | +7.4% ⚠ |
  | PRot † | +0.1% | +0.0% |
  | Quad[spin] | −16.2% | −1.2% |
  | Quad[nospin] | −27.1% | −0.3% |
  | RFCavity | −23.4% | +0.4% |
  | Sbend[spin] | −13.0% | −5.0% |
  | Sbend[nospin] | −12.1% | −3.8% |
  | ShortRF | −23.8% | −16.9% |
  | SoftQuadrupole | −24.8% | −0.9% |
  | SoftSolenoid | −23.6% | +0.0% |
  | Sol[spin] | −12.3% | −2.3% |
  | Sol[nospin] | −9.8% | +1.2% |
  | TaperedPL[spin] | −13.0% | −9.4% |
  | TaperedPL[nospin] | −27.2% | −10.4% |
  | ThinDipole | −27.5% | −23.7% |

  **Notes**
  - Gain ≈ (alignment FLOPs) ÷ (total kernel cost): largest for cheap pushes, 
  ~0 for the heavy spill-bound exact integrators (ExactQuad, ExactCFbend).
  - **DP**: broad win across the element set. **SP**: smaller (SP is more 
  memory-bound and the heavy exact kernels no longer spill), but still gains on
  cheap elements.
  - † **PRot** is the only element that does not inherit `mixin::Alignment`, so
  ON ≡ OFF — its ~0% is the noise baseline.
  - ⚠ rows are measurement/codegen noise, not real regressions: heavy exact 
  kernels in SP have *identical* register usage for the aligned vs. 
  always-shift paths (pure scheduling jitter on sub-0.1 ms kernels), and the 
  aperture elements have particle-loss-dependent timing. For aligned elements 
  #1499 only ever removes instructions.
  - RTX A2000 laptop GPU with unlocked clocks; min-of-4 tames but doesn't 
  eliminate noise on the smallest SP kernels.
</details>


src/elements/mixin/beamoptic.H
```diff
@@ -241,10 +241,18 @@ namespace detail
             constexpr bool has_alignment = std::is_base_of_v<mixin::Alignment, T_Element>;
             constexpr bool has_aperture = std::is_base_of_v<mixin::PipeAperture, T_Element>;
 
+            // runtime short-cut: only this element instance actually has a
+            // translation/rotation error. The flag is uniform across the warp/SIMD
+            // batch (it is an element property), so the branch is divergence-free.
```
ax3l:
```suggestion
            // skip alignment shifts & rotations if the element is perfectly aligned
```

src/elements/mixin/beamoptic.H
```diff
@@ -386,9 +396,17 @@ namespace detail
...
```
ax3l:
```suggestion
            // skip alignment shifts & rotations if the element is perfectly aligned
```

src/elements/mixin/beamoptic.H
```diff
@@ -241,10 +241,18 @@ namespace detail
...
+            bool misaligned = false;
+            if constexpr (has_alignment) { misaligned = m_element.misaligned(); }
+
             // shift into the alignment-error frame (compile-time gated)
             if constexpr (has_alignment) {
-                m_element.shift_in(x, y, px, py);
-                m_element.rotate_in(sx, sy);
+                if (misaligned) {
```
ax3l: It would be slightly cleaner if we checked this inside shift_in and rotate_in, but then we have to check twice. So I pulled the `if` out.

ax3l:
I did a test on real CPU hardware (local Intel CPU) and w/o SIMD the performance on CPU is the same.
With SIMD (which is what CodSpeed measures), we loose 9-10% for Sbend and CFbend. If we want to re-armortize this later on, we can move the whole alignment branch outside of the push operator and specialize it outside of the particle loop.
The improvements in all other elements are real when tested locally.

src/elements/mixin/beamoptic.H
ax3l: Even better would be to specialize the whole `operator()` for aligned/misaligned.

src/elements/mixin/beamoptic.H
```diff
@@ -386,9 +394,15 @@
             constexpr bool has_alignment = std::is_base_of_v<mixin::Alignment, T_Element>;
             constexpr bool has_aperture = std::is_base_of_v<mixin::PipeAperture, T_Element>;
 
+            // skip alignment shifts & rotations if the element is perfectly aligned
+            bool misaligned = false;
```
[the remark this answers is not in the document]
ax3l: We have to define it here, even if `if constexpr (!has_alignment)` never touches it.

src/elements/mixin/beamoptic.H
ax3l: implemented.

ax3l:
Implemented now a branch-free variant. This adds 2x compile time for elements for the above performance gains.  For fast compiles at the cost of runtime, the `-DImpactX_OPTIMIZE_ALIGNMENT=OFF` option can be set, which will compile and perform as before.

cemitch99 (approved):

PR #1501, opened 2026-06-12
Elements: Const Correctness

ax3l:
We rely on proper marking of `const` to avoid unnecessary write-back/store in SIMD code. Ensure all phase space and spin pushes are properly declared. Simplify code that never truly modifies a passed argument.


ax3l:
(The commented out lines are intentional, as we do in other places.)

cemitch99 (approved):

cemitch99 (approved):

PR #1503, opened 2026-06-15
CI: FFTW w/ Newer CMake & VS Studio 2026

ax3l:
Migrate to VS 2026: https://github.com/actions/runner-images/issues/14017

Fix FFTW:
```
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.

-- Building for: Visual Studio 18 2026
-- Configuring incomplete, errors occurred!
```

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -30,9 +30,9 @@ jobs:
              ccache-${{ github.workflow }}-${{ github.job }}-git-
     - name: Download Dependencies
       run: |
-        Invoke-WebRequest http://fftw.org/fftw-3.3.10.tar.gz -OutFile fftw-3.3.10.tar.gz
-        7z.exe x -r fftw-3.3.10.tar.gz
-        7z.exe x -r fftw-3.3.10.tar
+        Invoke-WebRequest http://fftw.org/fftw-3.3.11.tar.gz -OutFile fftw-3.3.11.tar.gz
+        7z.exe x -r fftw-3.3.11.tar.gz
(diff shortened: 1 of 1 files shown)


ax3l:
`test_Drift[spin]` performance changes: noise, not real. Cache grind layout noise from the simulated cache component (the simulated cache-miss cost dropped). Instructions and memory volume are identical.

##  Why the cache number moved without any code change (Claude)

CodSpeed's Simulation runner uses cachegrind: instruction count and bytes-touched are layout-independent and deterministic, but the cache cost depends on the addresses where code and data land, relative to a fixed simulated cache geometry. Bumping FFTW to 3.3.11 (and the build-config churn) relinks libimpactx.*.so and its deps, which shifts symbol/buffer addresses — even though Drift never calls FFT. Different alignment → fewer simulated cache-line conflicts on this run → lower "caches" cost. It's pure layout sensitivity.

This is exactly the recorded CodSpeed layout-noise floor for these cheap, memory-bound element benchmarks (~5–6% phantom swings; this is 7.15%). The diagnosis recipe — split instructions/caches/memory — confirms it cleanly: instructions + memory flat, caches alone moved.

EZoni (approved):

cemitch99 (approved):

PR #1506, opened 2026-06-15
Envelope Tracking with Rotation

### LINKED ISSUE #1505: 90 Degree Rotation for Vertical Bends is Off
For MAD-X bends like:
```
vsei2.5: rbend, at=56679.3727187015, l=4.0, angle=-0.00038835, tilt=-1.570796326795, k1=0.0, e1=0.0, e2=0.0, k2=0.0, hgap=0.0, fint=0.0, fintx=0.0, knl={0.0,0.0,0.0,0.0,0.0,0.0}, ksl={0.0,0.0,0.0,0.0,0.0,0.0};
```
-> vertical bend (tilt = -90 degrees)

## Claude

The translated transfer map puts the dispersion in x/px (R(0,5)=8.2e-4, R(1,5)=4.1e-4) — i.e. a horizontal bend — even though tilt=-π/2 should make this a vertical bend (dispersion in y/py). The rotation (tilt) is not affecting the bend plane in the transfer map. The question is whether this is a translation bug or whether ImpactX's linear-map path ignores the alignment rotation. Let me check how transport_map handles the Alignment rotation versus the particle-push path.

The LinearTransport mixin (lines 62–84, 121–132) applies element.transport_map(ref) directly with no alignment wrapping, and walk_lattice likewise just composes transport_map results. The particle-push path, by contrast, uses the Alignment mixin's transform/undo_transform
  (rotating coordinates in/out). So the rotation is applied to tracking but dropped in the linear-map/transfer_map path. Let me confirm the particle push applies alignment, which would make this an ImpactX transfer-map bug — not a translation bug.

This is now conclusive from the code: the particle-push path applies the Alignment rotation (beamoptic.H:242/388/459–464), but safe_transport_map (LinearMap.cpp:61) returns the element's intrinsic transport_map(ref) with no alignment wrapping, and walk_lattice composes those raw maps. So transfer_map ignores element rotation/tilt. Let me confirm empirically with a clean test: the transfer map of a tilted bend vs. an untilted one should be identical if rotation is being dropped.

ax3l:
Support `rotation` (transverse alignment) in envelope tracking. Enables to model, among others, vertical bends to fix #1505.

Co-authored-by: @cemitch99 & @aeriforme 

- [x] rebase after #1504 was merged
- [x] add a new user-facing example with a vertical dipole: the [dogleg](https://impactx.readthedocs.io/en/latest/usage/examples/dogleg/README.html) example, because it has nonzero dispersion at the end

### DIFF
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -30,9 +30,10 @@ Tracking Modes
 
   .. note::
 
-     Our current ``envelope`` tracking implements ideal transfer maps, assuming always zero misalignments (translation or rotations).
-     Support for misalignments and feed-down effects in envelope tracking is in development.
-     Until then, misalignment options set on elements are silently ignored.
+     Our current ``envelope`` tracking implements ideal transfer maps, assuming always zero misalignments (translations).
+     Element rotations are handled.
+     Support for translations errors, non-zero envelope means, and feed-down effects in envelope tracking is in development.
+     Until then, translations errors set on elements are silently ignored.
 
 .. _running-cpp-parameters-particle:
 
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -437,9 +437,10 @@ Collective Effects & Overall Simulation Parameters
 
       .. note::
 
-         Our current envelope tracking implements ideal transfer maps, assuming always zero misalignments (translation or rotations).
-         Support for misalignments and feed-down effects in envelope tracking is in development.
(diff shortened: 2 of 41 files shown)


cemitch99 (commented):
Implementation looks correct.  I think we need one test--for example, a comparison of particle tracking and envelope tracking in a lattice that includes a misalignment rotation (e.g., a vertical dipole).

cemitch99 (approved):

PR #1507, opened 2026-06-15
Diagnostics: Envelope Mean is Zero

ax3l:
Until we implement centroid tracking #1409 in envelope tracking mode, our mean is always zero (i.e., well-defined). No need to print NaN there, we can be explicit in the diagnostics.


cemitch99 (approved):
Yes.

PR #1513, opened 2026-06-16
MAD-X: `RFCAVITY` is Thin

ax3l:
The `l` in [MAD-X RFCAVITY](https://madx.web.cern.ch/webguide/manual.html#Ch11.S13) is ignored and modeled as a [drift-kick-drift model](https://github.com/MethodicalAcceleratorDesign/MAD-X/blob/7998e738eca575c82a424138517c2a17115f3cdb/src/trrun.f90#L1642-L1666), with `l/2` drifts and a `ShortRF` equivalent kick model in ImpactX.

Before, we translated this to a thick ImpactX RFCavity, which is more detailed than what MAD-X models.
(The MAD-X RF Cavity also supports a fringe field model, which we ignore/warn when encountered.)

### DIFF
--- a/src/python/impactx/madx_to_impactx.py
+++ b/src/python/impactx/madx_to_impactx.py
@@ -396,15 +396,6 @@ def lattice(
             return volt_MV
         return volt_MV / ref_mass_MeV
 
-    def normalized_rf_escale(volt_MV, ds):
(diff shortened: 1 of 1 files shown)


cemitch99 (approved):

ax3l:
Benchmark above is a glitch (parallel merge of #1501 )

PR #1514, opened 2026-06-16
Add QuadEdge spin push.

cemitch99:
Add the spin push through a quadrupole fringe field (`QuadEdge`).  The model used is taken from eq. (41) of: 

D. T. Abell et al, "Accurate and efficient spin integration for particle accelerators,"
Phys. Rev. Accel. Beams 18, 024001 (2015).

- [x] add spin push

Follow-up:  Add benchmark tests involving spin for elements with fringe fields.
The issue with reversibility has been added as Issue #1562

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -2170,8 +2170,8 @@ add_impactx_test(sextupole-kick.py
 
 # Reversibility Test for Spin in a Quadrupole with Fringe Fields ###########
 add_impactx_test(reversibility-quad-fringe-spin.py
-    examples/spin_tracking/run_reversibility-quad-fringe-spin.py
+    examples/spin_tracking/run_reversibility_quad_fringe_spin.py
     ON    # ImpactX MPI-parallel
-    examples/spin_tracking/analysis_reversibility-quad-fringe-spin.py
+    examples/spin_tracking/analysis_reversibility_quad_fringe_spin.py
     OFF  # no plot script yet
 )
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -2167,3 +2167,11 @@ add_impactx_test(sextupole-kick.py
     examples/multipole/analysis_sextupole_kick.py
     OFF  # no plot script yet
 )
+
+# Reversibility Test for Spin in a Quadrupole with Fringe Fields ###########
+add_impactx_test(reversibility-quad-fringe-spin.py
+    examples/spin_tracking/run_reversibility-quad-fringe-spin.py
+    ON    # ImpactX MPI-parallel
+    examples/spin_tracking/analysis_reversibility-quad-fringe-spin.py
+    OFF  # no plot script yet
+)
--- a/examples/spin_tracking/README.rst
+++ b/examples/spin_tracking/README.rst
@@ -595,5 +595,3 @@ The analysis can be run using:
    .. literalinclude:: analysis_reversibility_quad_fringe_spin.py
       :language: python3
       :caption: You can copy this file from ``examples/spin_tracking/analysis_reversibility_quad_fringe_spin.py``.
-
-
--- a/examples/spin_tracking/README.rst
+++ b/examples/spin_tracking/README.rst
@@ -548,3 +548,52 @@ The analysis can be run using the following script:
    .. literalinclude:: analysis_cfbend_spin_scaling.py
       :language: python3
       :caption: You can copy this file from ``examples/spin_tracking/analysis_cfbend_spin_scaling.py``.
+
+
+.. _examples-reversibility-quad-fringe-spin:
+
+Reversibility Test for Spin in a Quadrupole with Fringe Fields
+================================================================
+
+This example tests the reversibility of particle spin and orbit tracking through a quadrupole with nonlinear entry and exit fringe fields.
+
+Particle are tracked forward through a quadrupole with fringe fields, and then backward through the same physical fields.
+
+In this test, the initial and final values of all six phase space variables, as well as the three spin variables, must coincide.
+
+Note:  Because the exact inverse of the quadrupole fringe field map is not yet implemented, the test will pass only if a large tolerance value is specified.
+This value will be tightened once the exact inverse is implemented, and the test input modified accordingly.
+
+
+Run
+---
+
+This example can be run as:
+
+* **Python** script: ``python3 run_reversibility_quad_fringe_spin.py``
+
+For `MPI-parallel <https://www.mpi-forum.org>`__ runs, prefix these lines with ``mpiexec -n 4 ...`` or ``srun -n 4 ...``, depending on the system.
+
+.. tab-set::
+
+   .. tab-item:: Python: Script
+
+       .. literalinclude:: run_reversibility_quad_fringe_spin.py
+          :language: python3
+          :caption: You can copy this file from ``examples/spin_tracking/run_reversibility_quad_fringe_spin.py``.
+
+
+
+Analyze
+-------
+
+The analysis can be run using:
+
+
+.. dropdown:: Script ``analysis_reversibility_quad_fringe_spin.py``
+
+   .. literalinclude:: analysis_reversibility_quad_fringe_spin.py
+      :language: python3
+      :caption: You can copy this file from ``examples/spin_tracking/analysis_reversibility_quad_fringe_spin.py``.
+
+
--- a/examples/spin_tracking/README.rst
+++ b/examples/spin_tracking/README.rst
@@ -553,7 +553,7 @@ The analysis can be run using the following script:
 .. _examples-reversibility-quad-fringe-spin:
 
 Reversibility Test for Spin in a Quadrupole with Fringe Fields
-================================================================
+==============================================================
 
 This example tests the reversibility of particle spin and orbit tracking through a quadrupole with nonlinear entry and exit fringe fields.
 
@@ -561,8 +561,10 @@ Particle are tracked forward through a quadrupole with fringe fields, and then b
 
 In this test, the initial and final values of all six phase space variables, as well as the three spin variables, must coincide.
 
-Note:  Because the exact inverse of the quadrupole fringe field map is not yet implemented, the test will pass only if a large tolerance value is specified.
-This value will be tightened once the exact inverse is implemented, and the test input modified accordingly.
+.. note::
+
+   Because the exact inverse of the quadrupole fringe field map is not yet implemented, the test will pass only if a large tolerance value is specified.
+   This value will be tightened once the exact inverse is implemented, and the test input modified accordingly.
 
 
 Run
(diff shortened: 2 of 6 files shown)


examples/spin_tracking/run_reversibility_quad_fringe_spin.py
```diff
@@ -0,0 +1,97 @@
...
+
+# sequence 2 (edge only):
+sim.lattice.append(quad_edge_out)
+sim.lattice.append(quad_edge_in)
+
+# sequence 3 (inner sequence):
+sim.lattice.append(quad1)
+sim.lattice.append(quad_edge_out)
+sim.lattice.append(quad_edge_in)
+sim.lattice.append(inv_quad1)
```
cemitch99, line 81: Note:  Sequence 1 and Sequence 2 each work to high precision (that is, yield the identity).  Yet, somehow their composite in Sequence 3 fails in a spectacular way.  This is under investigation now.

examples/spin_tracking/run_reversibility_quad_fringe_spin.py
```diff
@@ -0,0 +1,97 @@
...
+inv_quad1 = elements.Quad(name="inv_quad1", ds=-L, k=kquad, nslice=ns)
+quad_edge_in = elements.QuadEdge(name="quad_edge1", k=kquad, flag="entry")
+quad_edge_out = elements.QuadEdge(name="quad_edge1", k=kquad, flag="exit")
+
+# set the lattice
+sim.lattice.append(monitor)
+
+# sequence 1 (quad only):
+sim.lattice.append(quad1)
+sim.lattice.append(inv_quad1)
```
cemitch99, line 71: It appears that orbit reversibility is broken here, at the level ~1% in py.  Why?  Is this related to [PR#1521](https://github.com/BLAST-ImpactX/impactx/pull/1521)?

cemitch99: This appears to be due to extreme problem parameters (e.g., k and resulting matrix elements in the y-py block are very large, while the initial conditions are very small, resulting in loss of precision.  This can be relaxed by making the beam large (on the scale of 0.1 mm for example), and making the focusing weaker (of order 1-10 for example).

examples/spin_tracking/run_reversibility_quad_fringe_spin.py
cemitch99: This appears to be due to the fact that the _orbit_ in QuadEdge is not reversed simply by taking entry->exit.  The correct inverse map needs to be applied for this element.

cemitch99:
@ax3l I think I understand the broken reversibility.  There are two distinct issues.  For the Quad body, the issue is simply a loss of numerical precision due to the extreme problem parameters in this test.  For the Quad edge, the issue is that the inverse of the orbit map is not exact--the true inverse is quite complicated, although we can obtain an expression for the inverse that is correct up to third order.  (This is a feature of maps obtained from mixed-variable generating functions in this form.)  We will need to deal with this at some point, but the issue is separate from this PR.

ax3l:
Oh that's good to know! I would recommend these things to track it:
- an issue with the comment details
- inline comments in `examples/spin_tracking/run_reversibility_quad_fringe_spin.py` and `tests/python/test_reversibility_elements.py` (can be brief an link to the issue)


That way we can set for now tolerances accordingly and won't hunt it down again.

examples/spin_tracking/run_reversibility_quad_fringe_spin.py
cemitch99, line 81:
```suggestion
sim.lattice.append(inv_quad1)
# Note: Sequence 1 and Sequence 2 each work to high precision (that is, yield the identity). Yet, # their composite in Sequence 3 fails (unless the tolerance is very loose).  This is due to the 
# fact that the orbit in QuadEdge is not reversed simply by taking entry->exit. See Issue #1562 .
```

examples/spin_tracking/run_reversibility_quad_fringe_spin.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python3
```
ax3l, line 1: This test needs to be added in `examples/CMakeLists.txt` or it will not be run by `ctest` and thus CI.

examples/spin_tracking/run_reversibility_quad_fringe_spin.py
```diff
@@ -0,0 +1,99 @@
...
+# sim.lattice.append(quad1)
+# sim.lattice.append(quad_edge_out)
+# sim.lattice.append(quad_edge_in)
+# sim.lattice.append(inv_quad1)
+# sim.lattice.append(quad_edge_out)
+
+sim.lattice.append(monitor)
+
+# run simulation
+sim.track_particles()
```
ax3l, line 97: Did you mean to commit an analysis script (or a block of validation here) for the result?

tests/python/test_reversibility_elements.py
```diff
@@ -667,6 +667,7 @@ def test_RFCavity(sim):
     ],
     ids=["linear-zero-gap", "linear-finite-gap", "nonlinear-finite-gap"],
 )
+# WARNING:  The orbit in QuadEdge is not reversed simply by taking entry->exit. The inverse map needs to be fixed.  See Issue #1562.
```
ax3l:
```suggestion
# WARNING:  The orbit in DipEdge is not reversed simply by taking entry->exit. The inverse map needs to be fixed.  See Issue #1562.
```

src/elements/QuadEdge.H
```diff
@@ -120,6 +122,7 @@ namespace impactx::elements
 
             // the global scale constant; sign depends on entry or exit
             m_a = static_cast<amrex::ParticleReal>(m_flag) * (-1_prt * m_g / 12.0_prt);
+
```
ax3l:
```suggestion
```

tests/python/test_reversibility_elements.py
cemitch99: I think you meant to add this comment at line 690 below.

examples/spin_tracking/run_reversibility_quad_fringe_spin.py
cemitch99: No.  Because the reverse of the `QuadEdge` map is not yet implemented correctly, this test fails as-is.  I had intended to keep this for a follow-up PR.  I will add an analysis script with a very large tolerance for now.

examples/CMakeLists.txt
```diff
@@ -2167,3 +2167,11 @@ add_impactx_test(sextupole-kick.py
     examples/multipole/analysis_sextupole_kick.py
     OFF  # no plot script yet
 )
+
+# Reversibility Test for Spin in a Quadrupole with Fringe Fields ###########
+add_impactx_test(reversibility-quad-fringe-spin.py
+    examples/spin_tracking/run_reversibility-quad-fringe-spin.py
+    ON    # ImpactX MPI-parallel
+    examples/spin_tracking/analysis_reversibility-quad-fringe-spin.py
```
```suggestion
add_impactx_test(reversibility-quad-fringe-spin.py
    examples/spin_tracking/run_reversibility_quad_fringe_spin.py
    ON    # ImpactX MPI-parallel
    examples/spin_tracking/analysis_reversibility_quad_fringe_spin.py
```

examples/spin_tracking/README.rst
```diff
@@ -548,3 +548,50 @@ The analysis can be run using the following script:
    .. literalinclude:: analysis_cfbend_spin_scaling.py
       :language: python3
       :caption: You can copy this file from ``examples/spin_tracking/analysis_cfbend_spin_scaling.py``.
+
+
+.. _examples-reversibility-quad-fringe-spin:
+
+Reversibility Test for Spin in a Quadrupole with Fringe Fields
+================================================================
```
```suggestion
Reversibility Test for Spin in a Quadrupole with Fringe Fields
==============================================================
```

examples/spin_tracking/README.rst
```diff
@@ -548,3 +548,50 @@ The analysis can be run using the following script:
...
+
+This example tests the reversibility of particle spin and orbit tracking through a quadrupole with nonlinear entry and exit fringe fields.
+
+Particle are tracked forward through a quadrupole with fringe fields, and then backward through the same physical fields.
+
+In this test, the initial and final values of all six phase space variables, as well as the three spin variables, must coincide.
+
+Note:  Because the exact inverse of the quadrupole fringe field map is not yet implemented, the test will pass only if a large tolerance value is specified.
+This value will be tightened once the exact inverse is implemented, and the test input modified accordingly.
```
```suggestion
.. note::

   Because the exact inverse of the quadrupole fringe field map is not yet implemented, the test will pass only if a large tolerance value is specified.
   This value will be tightened once the exact inverse is implemented, and the test input modified accordingly.
```

examples/spin_tracking/run_reversibility_quad_fringe_spin.py
```diff
@@ -0,0 +1,99 @@
...
+# sequence 2 (edge only):
+sim.lattice.append(quad_edge_out)
+sim.lattice.append(quad_edge_in)
+
+# sequence 3 (inner sequence):
+sim.lattice.append(quad1)
+sim.lattice.append(quad_edge_out)
+sim.lattice.append(quad_edge_in)
+sim.lattice.append(inv_quad1)
+# Note: Sequence 1 and Sequence 2 each work to high precision (that is, yield the identity). Yet, # their composite in Sequence 3 fails (unless the tolerance is very loose).  This is due to the
```
ax3l:
```suggestion
# Note: Sequence 1 and Sequence 2 each work to high precision (that is, yield the identity). Yet,
# their composite in Sequence 3 fails (unless the tolerance is very loose).  This is due to the
```

examples/spin_tracking/analysis_reversibility_quad_fringe_spin.py
```diff
@@ -0,0 +1,109 @@
...
+    [dx_max, dpx_max, dy_max, dpy_max, dt_max, dpt_max],
+    [0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+    atol=atol,
+)
+
+print("Change in the spin:")
+print("||delta s||_max", dspinmax)
+
+atol = 2.0  # large tolerance here, because orbit reversiblity is not yet implemented
+print(f"  atol={atol}")
```
ax3l, line 101: Understood on the reversibility.

Just a codex note that we also would not detect this simpler forward bug:
>   - [P1] The new test cannot detect a missing or incorrect spin push. In
    analysis_reversibility_quad_fringe_spin.py:100
>     atol = 2.0. Since spin rotations preserve unit length, (|s_f-s_i|\le2) for
    any rotation. A no-op QuadEdge spin implementation also produces zero error
    and passes. The existing test_QuadEdge
>
> We could add a deterministic single-edge test against the analytic spin map:
> Choose a particle with nonzero x*y
> and transverse spin, push one entry/exit edge, and compare against the
> expected z rotation with lambda_z = -flag * g * (1 + G) * x * y at normal
> numerical precision. Ideally cover both entry/exit signs and unit
> conversion.

ax3l (approved):
LGTM from the impl, ok to improve validation and reversibility analysis in follow-ups more.

Issue #1483, opened 2026-05-31
Exact symplectic maps (ExactCFbend, +multipole) do not close in single precision


ax3l:
## Summary
Forward + exact-inverse (identity-section) lattices built from the **exact CF-bend** maps do not return the beam to its input in single precision (values finite, no NaN — loss of significance in the forward/inverse map cancellation, growing with the number of slices):

- `exact-cfbend` (10 slices/element): final-vs-initial roundtrip drifts ~**20%** (the `.py` run with `ns=1` is only ~0.7%).
- `exact-cfbend-multipole`: final beam ~**75%** off the initial beam.
- pytest `test_exact_cfbend_multipole_spin`: `position_x` roundtrip off **1.4e-4** (atol 1e-7).

## Reproduce

Single-precision build (from the repo root):
```bash
cmake -S . -B build_sp -DImpactX_PRECISION=SINGLE -DImpactX_PYTHON=ON -DImpactX_COMPUTE=OMP -DImpactX_FFT=ON
cmake --build build_sp -j 8 --target pip_install
```
Then:
```bash
# examples (exact-cfbend.analysis and exact-cfbend-multipole.analysis fail)
ctest --test-dir build_sp -R "exact-cfbend" --output-on-failure
# focused pytest
python -m pytest tests/python/test_exact_cfbend_multipole_spin.py -q
```

## Where to look
- `src/elements/ExactCFbend.H` and the multipole kick: the forward/inverse map composition over `mapsteps`/`nslice` loses float32 significance.

## Not a fix
Investigate the map cancellation rather than loosening tolerances. (The example reversibility analyses are intentionally left strict.)

---
*(found and opened w/ Claude)*

ax3l:
**Claude analysis running Clang/LLVM NSan:** sharpened localization — in double precision the `ExactCFbend` particle map is sanitizer-quiet; the surviving cancellation hotspots are the **versine terms** (`cos(theta) - 1` class) in the reference-particle pushes and the map's position update, which lose ~56 of 52/24 bits per slice and accumulate with `nslice`, matching the slice-count growth reported here.

[...]

PR #1520, opened 2026-06-18
SP Tests: Tolerances

ax3l:
Avoid that our tests fail in SP, which we use to mark a build in package managers like CF as valid:
- skip known bugs #1483
- relax tolerance where scalar/SIMD need a bit more slack

### DIFF
--- a/tests/python/test_plasma_lens_spin.py
+++ b/tests/python/test_plasma_lens_spin.py
@@ -155,14 +155,14 @@ def test_tapered_pl_spin():
     np.testing.assert_allclose(
         [meansxi, meansyi, meanszi],
         [0.6, 0.5, 0.4],
-        atol=2.0e-3,
+        atol=2.0e-3 if Config.precision != "SINGLE" else 5.0e-3,
         rtol=0,
     )
     # test final polarization
     np.testing.assert_allclose(
         [meansxf, meansyf, meanszf],
         [0.6, 0.5, 0.4],
-        atol=2.0e-3,
+        atol=2.0e-3 if Config.precision != "SINGLE" else 5.0e-3,
         rtol=0,
     )
     # test spin evolution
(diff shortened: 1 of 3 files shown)


tests/python/test_plasma_lens_spin.py
```diff
@@ -155,14 +155,14 @@ def test_tapered_pl_spin():
     np.testing.assert_allclose(
         [meansxi, meansyi, meanszi],
         [0.6, 0.5, 0.4],
-        atol=2.0e-3,
+        atol=2.0e-3 if Config.precision != "SINGLE" else 5.0e-3,
```
EZoni, line 158: The tolerance difference between double and single precision is surprisingly small here.

tests/python/test_plasma_lens_spin.py
```diff
@@ -155,14 +155,14 @@ def test_tapered_pl_spin():
...
         rtol=0,
     )
     # test final polarization
     np.testing.assert_allclose(
         [meansxf, meansyf, meanszf],
         [0.6, 0.5, 0.4],
-        atol=2.0e-3,
+        atol=2.0e-3 if Config.precision != "SINGLE" else 5.0e-3,
```
EZoni, line 165: The tolerance difference between double and single precision is surprisingly small here.

EZoni (approved):

tests/python/test_plasma_lens_spin.py
ax3l: yes, particle resolution is dominating the tolerance for this case.

tests/python/test_plasma_lens_spin.py
ax3l: yes, particle resolution is dominating the tolerance for this case.

PR #1526, opened 2026-06-22
Sphinx: `:pp:param:` role

### LINKED ISSUE #1361: Reformat `parameters.rst`
Using Sphinx variables as in https://github.com/BLAST-WarpX/warpx/pull/6639, we can make our parameter inputs file docs more readable and cross-referencable.

ax3l:
Port from WarpX, based on @eebasso's work there :)

Preview of new "Parameters: Inputs File" (`parameters.rst`) section: https://impactx--1526.org.readthedocs.build/en/1526/usage/parameters.html

Close #1361

### DIFF
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -20,294 +20,524 @@ This documents how to use ImpactX with an input file (``impactx input_file.in``)
 Tracking Modes
 --------------
 
-* ``algo.track`` (``string``)
+.. pp:param:: algo.track
+    :type: ``string``
+    :default: ``particles``
 
-  Mode that specifies how the beam is tracked:
+    Mode that specifies how the beam is tracked:
 
-  * ``particles`` (default): symplectic particle tracking
-  * ``envelope``: beam envelope (covariance matrix) tracking, through linearized transport maps
-  * ``reference_orbit``: only tracking of the reference particle orbit
+    * ``particles`` (default): symplectic particle tracking
+    * ``envelope``: beam envelope (covariance matrix) tracking, through linearized transport maps
+    * ``reference_orbit``: only tracking of the reference particle orbit
 
-  .. note::
+    .. note::
 
-     Our current ``envelope`` tracking implements ideal transfer maps, assuming always zero misalignments (translations).
-     Element rotations are handled.
-     Support for translations errors, non-zero envelope means, and feed-down effects in envelope tracking is in development.
-     Until then, translations errors set on elements are silently ignored.
+       Our current ``envelope`` tracking implements ideal transfer maps, assuming always zero misalignments (translations).
+       Element rotations are handled.
+       Support for translations errors, non-zero envelope means, and feed-down effects in envelope tracking is in development.
+       Until then, translations errors set on elements are silently ignored.
 
 .. _running-cpp-parameters-particle:
 
 Initial Beam Distributions
 --------------------------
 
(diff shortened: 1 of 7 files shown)


docs/source/usage/parameters.rst
```diff
@@ -20,294 +20,524 @@ This documents how to use ImpactX with an input file (``impactx input_file.in``)
...
-  Indicates the element type for this lattice element. This should be one of:
+    Indicates the element type for this lattice element. This should be one of the following.
 
 
 ``aperture``
 ^^^^^^^^^^^^
 
-``aperture`` for a thin collimator element applying a transverse aperture boundary.
+``aperture`` for a thin collimator element applying a transverse aperture boundary,
+e.g. for an element ``my_aperture.type = aperture``.
```
EZoni: We now document element parameters with example names like `my_aperture`, as in, e.g., `my_aperture.type = aperture`. 

Before we used the placeholder syntax `<element_name>` with the `<` and `>`, which I think is widely understood by users as a hint to substitute their own element name. Was this a deliberate choice or more of a side effect of the PR?

I personally think that the placeholders are clearer, so I would vote for something like `<aperture_element_name>.type` or `<aperture_name>.type` or, at least, `<my_aperture>.type` rather than `my_aperture.type`.

Just a suggestion up for discussion, not sure what others think.

ax3l: Yes, that was a deliberate choice because `<element_name>.k` and many other parameters are common to many element kinds, so it is not a unique placeholder we can link to / creates naming clashes in the TOC tree.

A unique placeholder like you suggested comes out to the same solution, so the main question is do we want to add `<...>` to keep it concrete. I personally like to have concrete examples one can copy-paste, so I skipped the chevrons this time, but open for ideas.
I definitely met users before that did not understand the universally understood concept of `[...]` as optional and `<...>` as required placeholders... and then answered "your code does not work" questions about failing inputs that had those placeholders in them :sweat: 

Let's brainstorm more and discuss in our weekly meeting on Thu what to pick! :)

EZoni: Yes, that was basically the question.

Is it better `my_aperture.type = aperture` or `<my_aperture>.type = aperture`? Is there a risk that `my_aperture` is mistaken for an ImpactX reserved name rather than a placeholder? 

Looking at other instances, e.g., lines 390-392 below,
```rst
.. pp:param:: my_monitor.name
    :type: ``string``
    :default: ``my_monitor``
```
I'm not even sure myself. Is `my_monitor` a placeholder or an ImpactX reserved name? The `:default: ''my_monitor''` seems to suggest the latter or else why would a parameter default to a placeholder name?

ax3l: Yep, I mean we use `<>` and `[]` already, so we could keep it and just change what is in between as you suggested. Maybe there is also a technical trick how we can make the existing prior names unique and linkable (it's just a TOC/anchor name)...? But probably needs Sphinx hackery.

ax3l: Discussed: `<monitor_name>` and equivalent is a good way to lint it.

ax3l: Done! :)

EZoni (approved):

EZoni:
@ax3l 

I reviewed the commits after my last review and they look good to me.

cemitch99 (approved):

PR #1527, opened 2026-06-22
Restore import of matplotlib for example plotting

cemitch99:
To run the plotting feature in this example, matplotlib needs to be imported.  Although plotting is commented, this PR restores the appropriate line.

### DIFF
--- a/examples/expanding_beam/run_expanding_fft_2D_test_particles.py
+++ b/examples/expanding_beam/run_expanding_fft_2D_test_particles.py
@@ -8,6 +8,7 @@
 
 import numpy as np
 import pandas as pd
+#import matplotlib.pyplot as plt
 
 import amrex.space3d as amr
 from impactx import ImpactX, distribution, elements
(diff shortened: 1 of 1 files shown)


qianglbl (approved):

examples/expanding_beam/run_expanding_fft_2D_test_particles.py
```diff
@@ -9,6 +9,7 @@
 import numpy as np
 import pandas as pd
 
+# import matplotlib.pyplot as plt
```
ax3l:
```suggestion
```

examples/expanding_beam/run_expanding_fft_2D_test_particles.py
```diff
@@ -118,7 +119,7 @@ def hook_before_slice(sim):
 # for i in range(0, len(df), n):
 #    subset = sorted_df.iloc[i : i + n]
 #    plt.scatter(subset["s"], subset["x"], s=5)
-#
+
 # plt.xlabel("s [m]", fontsize=12)
 # plt.ylabel("x [mm]", fontsize=12)
 # plt.title("Test Particles: Horizontal Coordinates")
```
ax3l: How about this? :)
```suggestion
do_plot = False  # True to generate a plot of the test particle orbits
if do_plot:
    import matplotlib.pyplot as plt

    n = len(sarr)
    for i in range(0, len(df), n):
       subset = sorted_df.iloc[i : i + n]
       plt.scatter(subset["s"], subset["x"], s=5)

    plt.xlabel("s [m]", fontsize=12)
    plt.ylabel("x [mm]", fontsize=12)
    plt.title("Test Particles: Horizontal Coordinates")
```

examples/expanding_beam/run_expanding_fft_2D_test_particles.py
cemitch99: This is left here intentionally.  The point is that the user can uncomment those lines that are needed to generate plots of the test particle orbits, while avoiding loading an unnecessary package for the automated test.

cemitch99: Ah, I see this was moved below.

ax3l (approved):

PR #1530, opened 2026-06-25
Strang Split for Collective Effects

### LINKED ISSUE #846: Element Slicing: Strang Splitting

ax3l:
Implements a half-map + collective-kick + half-map strang split. This strongly reduces the number of slices one needs through elements when modeling collective effects.

Soft edge element note: it is important that `s_edge` is only updated once and then the map/2-kick-map/2 are applied, so that the s-dependent integration in the 2nd half-map starts where s ended in the first half-map.

Close #846
Close #920

- [x] depends on #1529
- [x] depends on #1558
- [x] depends on #1591
- [x] change strategy to: MKM split for all tick elements, nothing for thin.
- [x] cleanup
- [x] reduce nslice in the tests that allow it? usually from ~8x to 2x-5x
- [ ] follow-up performance opt: open issue to optimize that inner M(ds/2) maps can be merged if no per-slice diagnostics is used
- [ ] follow-up: `elements::ScopedInvertMap(T_Element && element, bool invert)` class

### DIFF
--- a/src/tracking/particles.H
+++ b/src/tracking/particles.H
@@ -12,8 +12,6 @@
 
 #include "elements/All.H"
 #include "elements/mixin/accessors.H"
-#include "elements/mixin/thick.H"
-#include "elements/mixin/traits.H"
 #include "particles/ImpactXParticleContainer.H"
 #include "tracking/TrackingDirection.H"
 #include "tracking/TrackingState.H"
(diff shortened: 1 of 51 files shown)


ax3l:
@cemitch99 small detail about the expanding beam test:
image


Improved in #1558

src/tracking/particles.H
```diff
@@ -102,14 +129,21 @@ namespace impactx
...
             }
 
             // number of slices used for the application of space charge
             int const nslice = elements::nslice(element_variant);
             amrex::ParticleReal const slice_ds = elements::slice_ds(element_variant); // in meters
 
+            // a kick over zero length leaves the momenta unchanged: skip its field solve
+            bool const kick = collective_effects && slice_ds != amrex::ParticleReal(0);
+
+            // Strang-split the transport ``M`` around the kick ``K`` (#846 / #920) wherever
```
ax3l: Not helpful refs:
```suggestion
            // Strang-split the transport ``M`` around the kick ``K`` wherever
```

src/tracking/particles.cpp
```diff
@@ -127,16 +132,25 @@ namespace impactx
             particles::spacecharge::HandleSpacecharge(amr_data, [this](){ this->ResizeMesh(); }, slice_ds);
         };
 
-        // the per-slice external-field transport map and per-slice housekeeping
-        auto element_push = [this, &pc, verbose, &pp_diag, diag_enable, &early_params_checked] (
+        // the external-field transport map ``M`` of one element slice
+        //   The Strang split applies this twice per slice, once per half-map, so it
```
ax3l:
```suggestion
        //   The Strang split around collective effects applies this twice per slice, once per half-map, so it
```

src/tracking/particles.H
```diff
@@ -27,49 +27,75 @@
...
+     *  element slice) and the @p slice_diagnostics book-keeping (boundary conditions,
+     *  lost particles, slice diagnostics), which runs once per slice.
+     *
+     *  With collective effects active (@p collective_effects), each slice of an element
+     *  that can be subdivided (@see elements::can_slice) is composed as the second-order,
+     *  time-symmetric Strang split
+     *      ``M(ds/2) K(ds) M(ds/2)``
+     *  half of the slice transport, the collective kick at the slice midpoint, then the
+     *  other half. The half transports are realized by temporarily doubling the element's
+     *  ``nslice``. The expensive collective solve still runs only once per slice.
```
ax3l:
```suggestion
     *  ``nslice``, which simply halves the ds per slice. The expensive collective solve still runs only once per slice.
```

src/tracking/particles.H
```diff
@@ -27,49 +27,75 @@
...
+     *  Finite-length elements that cannot be subdivided are composed to first order
+     *  instead: ``K(ds) M(ds)`` forward and ``M(ds) K(ds)`` backward.
+     *  Zero-length slices and runs without collective effects apply ``M`` alone, since a
```
ax3l:
```suggestion
     *  Zero-length elements run without collective effects apply ``M`` alone, since a
```

src/tracking/particles.H
```diff
@@ -27,49 +27,75 @@
...
+     *  kick over zero length leaves the momenta unchanged.
```
ax3l:
```suggestion
     *  kick over zero length leaves the momenta unchanged. Typically, thin elements are
     *  embedded in drift elements, which experience the collective effect kicks.
```

src/tracking/particles.H
```diff
@@ -102,14 +129,21 @@ namespace impactx
             else
             {
                 // back-tracking: invert this element's map in place so the element
-                // transport below applies the inverse map.
+                // transport below applies the inverse map M^-1.
                 elements::reverse(element_variant);
```
ax3l, line 144: Similar to `elements::ScopedNslice` we can probably define an `elements::ScopedInvertMap` that ensures we do not forget to undo this.

ax3l: Good follow-up PR to keep this PR small.

src/tracking/particles.H
```diff
@@ -128,17 +171,37 @@ namespace impactx
...
+                if (split)
                 {
+                    // M(ds/2) K(ds) M(ds/2): the doubled slice count halves each transport
+                    elements::ScopedNslice const half_slices(element_variant, 2 * nslice);
+                    element_transport(element_variant, step, period);
                     collective_kicks(element_variant, slice_ds);
-                    element_push(element_variant, step, period);
+                    element_transport(element_variant, step, period);
+                }
+                else if (kick)
```
cemitch99, line 197: A symmetric way of doing this case is 1/2 space charge kick + element kick + 1/2 space charge kick.

More expensive, but it should preserve reversibility and is second-order in ds.

src/tracking/particles.H
```diff
@@ -128,17 +171,37 @@ namespace impactx
...
-                // TODO: Strang-split here (#846 / #920)
-                if (forward)
+                if (split)
                 {
+                    // M(ds/2) K(ds) M(ds/2): the doubled slice count halves each transport
+                    elements::ScopedNslice const half_slices(element_variant, 2 * nslice);
+                    element_transport(element_variant, step, period);
                     collective_kicks(element_variant, slice_ds);
-                    element_push(element_variant, step, period);
+                    element_transport(element_variant, step, period);
```
cemitch99: I find the logic here confusing, but I think it is largely a matter of naming.  For example, the difference between `element_push` and `element_transport` is not apparent.  Since the split is M(ds/2)K(ds)M(ds/2), but there are four function calls, I think there should be some in-line comments clarifying the role of each function call.

cemitch99 (commented):
Thanks!  I left a few comments.  I also did some testing.  For the most part, the behavior in those tests looked reasonable.  Most importantly, I verified that the previous behavior can be reproduced by setting `strang_split = False` (a fallback in case any bugs creep in).

src/tracking/particles.H
ax3l: Discussed: 

- The extreme case here would likely be a user building their whole lattice from `LinearMap`, (`SpinMap` or `Programmable`) which cannot be automatically split.
- It would be worth updating the user-facing elements docs to indicate which elements cannot be automatically split for application of space charge.
- Having a 1/2 collective + element + 1/2 collective is likely the most consistent if strang-split is on (meaning: user would like 2nd-order precision), and falling back to the current collective + element (fw) / element + collective (bw) for when it is off.

src/tracking/particles.H
cemitch99: Sorry, I missed that `element_push` is deleted here...

ax3l: Oh, both are the same. I maybe should rename `element_transport` to `element_push` again, it really just calls `push(...)` now.

tests/python/test_strang_split.py
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022-2026 The ImpactX Community
+#
+# Authors: Axel Huebl
```
ax3l:
```suggestion
# Authors: Axel Huebl, Chad Mitchell
```

tests/python/test_programmable_collective_kick.py
```diff
@@ -0,0 +1,190 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022-2026 The ImpactX Community
+#
+# Authors: Axel Huebl
```
ax3l:
```suggestion
# Authors: Axel Huebl, Chad Mitchell
```

src/tracking/particles.H
ax3l: Updated accordingly :)

cemitch99 (approved):

PR #1532, opened 2026-06-25
Fix: GPU runs w/o GPU-Aware MPI

ax3l:
A few of our reductions in collective effects where assuming we have GPU-aware MPI support. In some cases, eg., conda-forge desktop MPICH this is not the case and crashes. This adds the conditional d2h copies before speaking to MPI in those cases.

- [x] quick LLM draft
- [x] detailed self-review
- [x] needs https://github.com/AMReX-Codes/amrex/pull/5516
- [x] cleanup
- [x] local GPU tests w/ GPU-unaware MPI from Conda-Forge are good now

### DIFF
--- a/src/particles/wakefields/ChargeBinning.H
+++ b/src/particles/wakefields/ChargeBinning.H
@@ -65,6 +65,44 @@ namespace impactx::particles::wakefields
         bool is_unity_particle_weight = false
     );
 
+    /** All-reduce (sum) a 1D histogram across MPI ranks, in place.
+     *
+     * The histogram is stored in (non-managed) device memory. A device pointer
+     * is only handed to MPI when AMReX is configured for GPU-aware MPI;
+     * otherwise the data is staged on the host for the collective. No-op for a
+     * single rank.
+     *
+     * @param[in,out] data device-resident 1D array, summed across all ranks
+     */
+    void AllReduceSum1D (
+        amrex::Gpu::DeviceVector<amrex::Real> & data
+    );
+
+    /** Reduce (sum) a 1D histogram onto a single MPI rank, in place.
+     *
+     * Same GPU-aware-MPI staging behavior as @ref AllReduceSum1D, but the
+     * reduced result is only valid on rank @p root.
+     *
+     * @param[in,out] data device-resident 1D array, summed onto @p root
+     * @param[in] root the MPI rank that receives the reduced result
+     */
+    void ReduceSum1D (
+        amrex::Gpu::DeviceVector<amrex::Real> & data,
+        int root
+    );
+
+    /** Broadcast a 1D histogram from the IO processor to every MPI rank, in place.
+     *
+     * On entry @p data is only valid on the IO processor; on exit every rank holds
+     * a correctly-sized copy (receivers are resized to match). Same GPU-aware-MPI
+     * staging behavior as @ref AllReduceSum1D. No-op for a single rank.
+     *
+     * @param[in,out] data device-resident 1D array, broadcast from the IO processor
+     */
+    void BcastFromIOProc1D (
+        amrex::Gpu::DeviceVector<amrex::Real> & data
+    );
+
 } // namespace impactx::particles::wakefields
 
 #endif // CHARGE_BINNING_H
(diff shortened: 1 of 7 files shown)


src/particles/wakefields/ChargeBinning.cpp
```diff
@@ -189,4 +194,96 @@ namespace impactx::particles::wakefields
...
+        amrex::Gpu::DeviceVector<amrex::Real> & data
+    )
+    {
+        // Nothing to communicate for a single rank: the local histogram is
+        // already the global one.
+        if (amrex::ParallelDescriptor::NProcs() == 1) { return; }
+
+        if (amrex::ParallelDescriptor::UseGpuAwareMpi()) {
+            // GPU-aware MPI can read the (non-managed) device buffer directly.
+            amrex::ParallelAllReduce::Sum(
```
ax3l: 3X here: Maybe if we add DeviceVector overloads to them we could make that automatic in AMReX... @WeiqunZhang ? :)

src/particles/wakefields/ChargeBinning.H
```diff
@@ -65,6 +65,44 @@ namespace impactx::particles::wakefields
...
+     * @param[in,out] data device-resident 1D array, summed across all ranks
+     */
+    void AllReduceSum1D (
+        amrex::Gpu::DeviceVector<amrex::Real> & data
+    );
+
+    /** Reduce (sum) a 1D histogram onto a single MPI rank, in place.
+     *
+     * Same GPU-aware-MPI staging behavior as @ref AllReduceSum1D, but the
+     * reduced result is only valid on rank @p root.
```
ax3l: Odd location, to move into a more general helper...

src/particles/wakefields/ChargeBinning.cpp
```diff
@@ -189,4 +194,96 @@ namespace impactx::particles::wakefields
...
+
+        int const root = amrex::ParallelDescriptor::IOProcessorNumber();
+
+        // Broadcast the length first: only the root computed data; the other
+        // ranks arrive with an empty vector and must size their buffer to match.
+        amrex::Long n = data.size();
+        amrex::ParallelDescriptor::Bcast(&n, 1, root);
+        if (!amrex::ParallelDescriptor::IOProcessor()) {
+            data.resize(n);
+        }
```
ax3l: Another real bug that this fixes, too.

But, even if we do not calculate as IO processor, we should know this size from user input. Avoid the extra MPI comm step that way.

src/particles/wakefields/ChargeBinning.cpp
ax3l: Confirmed on slack. I'll send a PR

ax3l: https://github.com/AMReX-Codes/amrex/pull/5516

src/particles/wakefields/HandleWakefield.H
```diff
@@ -137,21 +143,27 @@ namespace impactx::particles::wakefields
...
+                amrex::Gpu::dtoh_memcpy(h_wakefield.data(), convolved_wakefield.data(),
+                                        convolved_wakefield.size() * sizeof(amrex::Real));
                 std::cout << "Convolved wakefield: ";
                 std::ofstream outfile("convolved_wakefield.txt");
-                for (double const i : convolved_wakefield) {
-                    std::cout << i << " ";
-                    outfile << i << std::endl;
+                for (amrex::Real const val : h_wakefield) {
+                    std::cout << val << " ";
+                    outfile << val << std::endl;
```
WeiqunZhang: Minor suggestion: use `'\n'` instead of `std::endl` to avoid flushing the buffer unnecessarily.

ax3l: Thanks, good point!

ax3l:
```suggestion
                    outfile << val << "\n";
```

src/particles/wakefields/HandleWakefield.H
```diff
@@ -137,21 +143,27 @@ namespace impactx::particles::wakefields
...
                 }
                 std::cout << std::endl;
```
ax3l:
```suggestion
                std::cout << "\n";
```

ax3l:
@WeiqunZhang ready :)

cemitch99 (approved):

WeiqunZhang (approved):

PR #1538, opened 2026-06-26
pyImpactX: Only export `_PyInit_impactx_pybind`

ax3l:
Hide all other symbols to avoid clashes when loading transient dependencies in workflows that combine ImpactX, WarpX, and openPMD-api.

Relevant for pip/wheels that do static builds (i.e. Python wheels):
Currently, openPMD-api and ImpactX could not be `import`-ed together because HDF5 would clash.

Used in https://github.com/BLAST-ImpactX/impactx-wheels/pull/26

Not a problem when installing with shared dependencies (of HDF5, openPMD, ABLASTR, etc.), i.e., from conda-forge or Spack.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -241,6 +241,25 @@ if(ImpactX_PYTHON)
         PDB_OUTPUT_DIRECTORY ${CMAKE_PYTHON_OUTPUT_DIRECTORY}/impactx
         COMPILE_PDB_OUTPUT_DIRECTORY ${CMAKE_PYTHON_OUTPUT_DIRECTORY}/impactx
     )
+
+    # The CXX_VISIBILITY_PRESET above only hides pyImpactX's own objects, but
+    # statically-embedded dependencies (HDF5, openPMD, ADIOS2, and the ImpactX
+    # core archive) are compiled without hidden visibility and would otherwise
+    # be re-exported from this extension. If another extension loaded in the
+    # same process also statically embeds them (openpmd_api, WarpX, a second
+    # openPMD/HDF5), the duplicate exported symbols cross-bind, one has two
+    # HDF5/openPMD instances in one process, UBs and crashes.
+    # Export only the module init symbol so each extension keeps its
+    # dependencies private.
+    # Windows does not auto-export and Emscripten links a single namespace, so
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
codex:

• Findings

  - [P1] CMakeLists.txt:263: the Emscripten fix can be silently disabled when Python IPO/LTO is on. ImpactX_PYTHON_IPO defaults to ON, and the associated wheel work shows that with LTO enabled,
    Pyodide’s later -s SIDE_MODULE=1 can override this PR’s -sSIDE_MODULE=2. That leaves the module whole-archiving/re-exporting bundled HDF5/openPMD, which is the crash this PR is meant to
    prevent. I’d disable Python IPO for EMSCRIPTEN in ImpactX CMake itself, or otherwise guarantee -sSIDE_MODULE=2 is the final effective link setting.

  Checked
  I inspected PR 1538’s diff, local CMakeLists.txt context, PR checks/comments, openPMD-api PR 1903, and the related impactx-wheels PR 26. Normal CI is green; I did not run a local Emscripten/
  Pyodide build.

  Reference: Emscripten documents SIDE_MODULE=2 as the DCE side-module mode and says required exports must be kept via EXPORTED_FUNCTIONS:
  https://emscripten.org/docs/compiling/Dynamic-Linking.html#code-size

ax3l:
Codex is right and I actually had to actively disable LTO in https://github.com/BLAST-ImpactX/impactx-wheels/pull/26 to get it to work.
Let me see if I can disable this foot-gun here.

Update: Actually, let me confirm one last time with LTO in that, there was another issue and maybe `-sSide_MODULE=2` from the 2nd commit is not needed at all - because it gets overwritten with `=1` even when LTO is off (because of other logic from `pyodide-build`): https://github.com/BLAST-ImpactX/impactx-wheels/pull/26/changes/27e6b2d52ceafb78f5f4a2769dda26152893e00c

ax3l:
yes, this was indeed not needed :tada: I reverted the 2nd commit.

WeiqunZhang (approved):

ax3l:
Oh wait, now I reverted too much. I need at least `=1`

cemitch99 (approved):

PR #1540, opened 2026-06-27
`setup.py`: Windows `...openPMD_CMAKE_...` Env Vars

ax3l:
Same as https://github.com/openPMD/openPMD-api/pull/1896

We use this for pip-based installes, e.g., on cibuildwheel, to ensure we see missing deps at configure time instead of having to wait for runtime tests (and builds).


ax3l:
Hm, technically this is temporary and will be fixed in an openPMD >0.17.1 release via https://github.com/openPMD/openPMD-api/pull/1896 - so maybe no need to patch here.

PR #1550, opened 2026-07-06
Gauss2p5D solver for low relativistic beta

cemitch99:
This PR fixes #1549 .

- [x] address transverse kick
- [x] address longitudinal kick
- [x] add examples

PR #1556, opened 2026-07-08
Use linear interpolation for longitudinal position binning

cemitch99:
For solvers that access longitudinally binned data (along the `t` or `z` variable), improved accuracy is obtained by using linear interpolation (grid points -> particle position).

Testing will rely on PR[#1550](https://github.com/BLAST-ImpactX/impactx/pull/1550).


cemitch99:
The interpolated results are continuous, as expected, but some error is introduced--likely due to +/-1 indexing error--to be verified.

src/particles/spacecharge/GatherAndPush.cpp
```diff
@@ -198,8 +198,16 @@ namespace impactx::particles::spacecharge
...
                        }
                        #endif
-                       amrex::ParticleReal const Fxy = (Qb_abs==0.0) ? 0.0_prt : beam_profile[idx] / Qb_abs;
-                       amrex::ParticleReal const Fz = (Qb_abs==0.0) ? 0.0_prt : beam_profile_slope[idx] * charge_abs / Qb_abs;
+
+                       // Interpolation
+                       amrex::ParticleReal const idxreal = static_cast<amrex::ParticleReal>(idx);
+                       amrex::ParticleReal const lambda = (z - bin_min) / bin_size - idxreal;
+                       amrex::ParticleReal const beam_profile_interp = (1_prt-lambda)*beam_profile[idx] + lambda*beam_profile[idx+1];
+                       amrex::ParticleReal const beam_profile_slope_interp = (1_prt-lambda)*beam_profile_slope[idx] + lambda*beam_profile_slope[idx+1];
```
ax3l: The `beam_profile` has `num_bins + 1` size, so line 205 is fine.

But the line here on `beam_profile_slope[idx+1]` is out-of-bounds, because the `beam_profile_slope` only has size of `num_bins`.

src/particles/spacecharge/Gauss2p5dPush.cpp
```diff
@@ -289,8 +289,15 @@ namespace impactx::particles::spacecharge
                         std::cerr << "Warning: Index out of range for 2.5D Gaussian SC: " << idx << std::endl;
                     }
 #endif
-                    amrex::ParticleReal const Fxy = beam_profile[idx] * chargesign;
-                    amrex::ParticleReal const Fz = (apply_longitudinal_kick)? beam_profile_slope[idx] * charge : 0_prt;
+                    // Interpolation
+                    amrex::ParticleReal const idxreal = static_cast<amrex::ParticleReal>(idx);
+                    amrex::ParticleReal const lambda = (z - bin_min) / bin_size - idxreal;
+                    amrex::ParticleReal const beam_profile_interp = (1_prt-lambda)*beam_profile[idx] + lambda*beam_profile[idx+1];
+                    amrex::ParticleReal const beam_profile_slope_interp = (1_prt-lambda)*beam_profile_slope[idx] + lambda*beam_profile_slope[idx+1];
```
ax3l: As above: `beam_profile_slope[idx+1]` is out-of-bounds by one.

qianglbl (approved):

src/particles/spacecharge/GatherAndPush.cpp
ax3l: Simple modification: we clamp the index for the last value in the slope, when `idx` reaches num_bins` we do not interpolate anymore.

Codex idea:
Technically, `beam_profile_slope` is already the finite-difference slope over interval `i`.
```C++
  slope[i] = (beam_profile[i + 1] - beam_profile[i]) / bin_size;
```
So if you linearly interpolate `beam_profile` between `idx` and `idx + 1` the derivative over that same segment could be just:
```suggestion
                       amrex::ParticleReal const beam_profile_slope_interp = beam_profile_slope[idx];
```

src/particles/spacecharge/Gauss2p5dPush.cpp
ax3l:
```suggestion
                    amrex::ParticleReal const beam_profile_slope_interp = beam_profile_slope[idx];
```

src/particles/spacecharge/GatherAndPush.cpp
cemitch99: @ax3l Codex is taking the derivative of the linearly-interpolated fit to `beam_profile`, to get a fit that is piecewise-constant to `beam_profile_slope`.  This is equivalent what we have already (for the longitudinal kick), so it won't help address the underlying issue.

Strictly speaking, for consistency we should use a quadratic interpolation to fit `beam_profile` and a linear interpolation to fit `beam_profile_slope`.  Of course, the quadratic interpolation is more costly and requires careful handling of the endpoint values.

PR #1557, opened 2026-07-08
 BeamMonitor: Options `particles` and `beam_moments`

ax3l:
Add two more options to `BeamMonitor`:

- `particles=True`
      Output the individual particles of the beam (default: ``True``).
      When set to ``False``, the beam monitor becomes a fast, small diagnostic that stores only per-output attributes (reference particle and, if enabled, beam moments).

- `beam_moments=True`
      Output the beam moments (reduced beam characteristics) of the whole bunch as openPMD attributes on the ``beam`` particle species (default: ``True``).

This enables that the `BeamMonitor` can be used as a fast, binary monitor for moments, particles, nonlinear invariants of the insertion element, or any subset or combination of those. In particular, one can now use it instead of the slow ASCII files that our reduced beam diagnostics creates.

### DIFF
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -1762,7 +1762,7 @@ For an element with ``nslice`` > 1, the pushes and maps refer to a single ``ds/n
    :param rotation: rotation error in the transverse plane [degrees]
    :param name: an optional name for the element
 
-.. py:class:: impactx.elements.BeamMonitor(name, backend="default", encoding="g", period_sample_intervals=1)
+.. py:class:: impactx.elements.BeamMonitor(name, backend="default", encoding="g", period_sample_intervals=1, particles=True, beam_moments=True)
 
    A beam monitor, writing all beam particles at fixed ``s`` to openPMD files.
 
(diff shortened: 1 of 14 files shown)


docs/source/usage/python.rst
```diff
@@ -1773,21 +1773,33 @@ For an element with ``nslice`` > 1, the pushes and maps refer to a single ``ds/n
    ``json`` only works with serial/single-rank jobs.
    By default, the first available backend in the order given above is taken.
 
-   openPMD `iteration encoding <https://openpmd-api.readthedocs.io/en/0.14.0/usage/concepts.html#iteration-and-series>`__ determines if multiple files are created for individual output steps or not.
-   Variable based is an `experimental feature with ADIOS2 <https://openpmd-api.readthedocs.io/en/0.14.0/backends/adios2.html#experimental-new-adios2-schema>`__.
+   openPMD `iteration encoding <https://openpmd-api.readthedocs.io/en/latest/usage/concepts.html#iteration-and-series>`__ determines if multiple files are created for individual output steps or not.
+   Variable based encoding requires linear read access (``Access.read_linear``) in downstream readers.
```
```suggestion
   Variable based is an `experimental feature with ADIOS2 <https://openpmd-api.readthedocs.io/en/0.14.0/backends/adios2.html#experimental-new-adios2-schema>`__ and currently requires linear read access (``Access.read_linear``) in downstream readers.
```

cemitch99 (commented):
Ok, I tested this on the GSI benchmark example, and it appears to work.  Two notes:
1) in addition to/in lieu of the iteration number, the turn (period) number is useful output; of course, it can be reconstructed simply by counting!
2) if I turn particle output off, and then collect the particle data using `to_csv`, then I get an output file containing only the following headings:
`row positionOffset_t positionOffset_x positionOffset_y`
This would be a bit confusing to the user.  Perhaps the result should be entirely empty?  It's not clear why these headings in particular are retained.

src/elements/diagnostics/BeamMonitor.cpp
```diff
@@ -278,31 +274,70 @@ namespace detail {
...
+            // particle species without any component extent by default
+            io::Datatype const dtype_fl = io::determineDatatype<amrex::ParticleReal>();
+            auto d_none = io::Dataset(dtype_fl, {0});
+
+            beam["positionOffset"]["x"].resetDataset(d_none);
+            beam["positionOffset"]["x"].makeConstant(ref_part.x);
+            beam["positionOffset"]["y"].resetDataset(d_none);
+            beam["positionOffset"]["y"].makeConstant(ref_part.y);
+            beam["positionOffset"]["t"].resetDataset(d_none);
+            beam["positionOffset"]["t"].makeConstant(ref_part.t);
```
ax3l: Less confusing when inspected by users:
```suggestion
            beam["empty"][scalar].resetDataset(d_none);
            beam["empty"][scalar].makeConstant(0.0_prt);
```

ax3l:
Hi @cemitch99 , thanks for the review. Both addressed now.

cemitch99 (approved):

Issue #1102, opened 2025-08-18
Beam Diagnostics is too Slow


ax3l:
The diagnostics code in `reduced_beam_characteristics(pc)` is too slow. In 1-MPI-rank simulations like the [HTU beamline](https://impactx.readthedocs.io/en/latest/usage/examples/htu_beamline/README.html), when setting [`sim.particle_container().store_beam_moments = True`](https://impactx.readthedocs.io/en/latest/usage/python.html#impactx.ParticleContainer.store_beam_moments), it is dominating the runtime by ~1.5x compared to the next costly element of the actual simulation.

```
TinyProfiler total time across processes [min...avg...max]: 0.02604 ... 0.02604 ... 0.02604

-------------------------------------------------------------------------------------------------------
Name                                                    NCalls  Excl. Min  Excl. Avg  Excl. Max   Max %
-------------------------------------------------------------------------------------------------------
impactx::diagnostics::reduced_beam_characteristics(pc)      91    0.01197    0.01197    0.01197  45.96%
impactx::Push::ChrQuad                                      34   0.007997   0.007997   0.007997  30.71%
impactx::Push::ExactDrift                                   33   0.001654   0.001654   0.001654   6.35%
impactx::Push::ExactSbend                                    5  0.0004234  0.0004234  0.0004234   1.63%
impactX::collect_lost_particles                             91  0.0003877  0.0003877  0.0003877   1.49%
ImpactX::evolve::slice_step                                 91  0.0003815  0.0003815  0.0003815   1.47%
ImpactX::add_particles                                       1  0.0003395  0.0003395  0.0003395   1.30%
impactx::Push::Kicker                                        8  0.0002024  0.0002024  0.0002024   0.78%
ImpactXParticleContainer::record_beam_moments               91  0.0001794  0.0001794  0.0001794   0.69%
DistributionMapping::LeastUsedCPUs()                         1  0.0001495  0.0001495  0.0001495   0.57%
ImpactX::track_particles                                     1   3.08e-05   3.08e-05   3.08e-05   0.12%
impactx::Push                                               91  1.807e-05  1.807e-05  1.807e-05   0.07%
AmrMesh::MakeDistributionMap()                               1  7.808e-06  7.808e-06  7.808e-06   0.03%
DistributionMapping::SFCProcessorMapDoIt()                   1  2.937e-06  2.937e-06  2.937e-06   0.01%
Other                                                      357  0.0001655  0.0001655  0.0001655   0.64%
[...]

PR #1561, opened 2026-07-10
Optimize Performance of Beam Moments Calculation: Single Pass

ax3l:
Calculate beam moments in a single pass instead of two.

For beams with phase space components far off the reference particle orbit, we run the risk of floating point cancellation errors when we compute `Var = <x^2> − <x>^2` directly, whose single-precision error grows as `eps·(offset/sigma)^2` instead of the two-pass's `eps·(offset/sigma)`. Sampling and using the first (random) particle, which is in the order of the beam rms (pos, mom, spin), to shift brings this back to the linear accuracy scaling.

This now has 1 instead of 2 loops over all particles, and 18% less instructions in our benchmark.

First part of https://github.com/BLAST-ImpactX/impactx/issues/1102#issuecomment-4938086729

### DIFF
--- a/src/diagnostics/ReducedBeamCharacteristics.cpp
+++ b/src/diagnostics/ReducedBeamCharacteristics.cpp
@@ -16,6 +16,7 @@
 #include "EmittanceInvariants.H"
 
 #include <AMReX_BLProfiler.H>           // for TinyProfiler
+#include <AMReX_GpuDevice.H>            // for dtoh_memcpy
 #include <AMReX_GpuQualifiers.H>        // for AMREX_GPU_DEVICE
 #include <AMReX_ParallelDescriptor.H>   // for ParallelDescriptor
 #include <AMReX_ParticleReduce.H>       // for ParticleReduce
@@ -24,7 +25,9 @@
 #include <AMReX_SmallMatrix.H>          // for SmallMatrix
 #include <AMReX_TypeList.H>             // for TypeMultiplier
 
+#include <array>
 #include <limits>
+#include <vector>
 
 
 namespace impactx::diagnostics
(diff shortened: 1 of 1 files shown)


src/diagnostics/ReducedBeamCharacteristics.cpp
```diff
@@ -47,263 +52,226 @@ namespace impactx::diagnostics
...
         using PType = typename ImpactXParticleContainer::SuperParticleType;
 
+        /* The reduced beam characteristics are computed in a single pass over the
+         * particles from raw (weighted) power sums. To keep the second moments
+         * well-conditioned for beams that are off-center from the reference orbit
+         * in single precision, the sums are accumulated relative
+         * to a shift close to the actual beam centroid: any global constant is exact under
+         * the parallel-axis theorem, and sampling the first particle keeps
+         * (coordinate - shift) of order of the rms beam size. The subtraction is then
+         * Sterbenz-exact and the fused single pass matches the scaling accuracy of a
```
ax3l: This is a stretch.

More like:
```
// Accumulating the (weighted) power sums relative to a shift near the beam
// centroid keeps them at the O(rms^2) scale instead of O(offset^2). The
// recovered central moments then carry relative error ~eps*(offset/rms):
// linear, as in a mean-subtracted two-pass -- rather than the ~eps*(offset/rms)^2
// of a raw <u^2> - <u>^2. (Any global constant is exact under the parallel-axis
// theorem; sampling the first particle needs only offset/rms, not the true mean.
// In the offset >> rms limit the leading digits cancel exactly per Sterbenz's
// lemma, but the essential point is squaring the deviation, not the coordinate.)
```

ax3l:
CodSpeed measures on a single thread in DP, where the performance is neutral to slightly negative, because it is not a realistic benchmark on modern multi-core CPUs where we use multiple threads to saturate compute and memory BW.

ax3l:
### CPU: Local wall-clock benchmark of the single-pass fusion

Complementing the CodSpeed report (which is a cachegrind instruction/cache **simulation** at `OMP_NUM_THREADS=1`), here are **wall-clock** timings of `beam.beam_moments()` before (two-pass, `f3df88d51^`) and after (fused, `f3df88d51`).

**Environment**
- CPU: Intel Core i9-12900H (Alder Lake, ~24 MB L3), **P-cores only, no hyperthreads** — pinned one thread per physical P-core (6 P-cores available); E-cores unused.
- Build: `-DCMAKE_BUILD_TYPE=Release -DImpactX_SIMD=ON`, GCC 13.3, single MPI rank.
- Metric: min wall-clock over rounds. Quiet machine (no overlapping build/tests).

#### Single thread — complete particle sweep (ns/particle)
speedup = two-pass ÷ fused

| Np | DP two-pass | DP fused | DP speedup | SP two-pass | SP fused | SP speedup |
|---:|---:|---:|:---:|---:|---:|:---:|
| 1k | 13.3 | 13.8 | 0.96× | 13.8 | 12.9 | 1.07× |
| 10k | 7.90 | 8.62 | 0.92× | 8.47 | 7.52 | 1.13× |
| 100k | 7.43 | 7.94 | 0.94× | 8.01 | 7.03 | 1.14× |
| 1M | 8.12 | 8.45 | 0.96× | 8.22 | 7.27 | 1.13× |
| 10M | 8.61 | 9.09 | 0.95× | 9.01 | 8.02 | 1.12× |

At **1 thread**, DP is ~4–8% *slower* and SP ~1.1× faster: one core is not memory-bandwidth-bound, so halving the passes over the SoA does not pay off for DP, while the wider single-pass reduction costs slightly more compute. **This is the regime CodSpeed measures**, so its DP flag is expected.

#### Thread scaling on P-cores (no HT) — complete sweep, speedup = two-pass ÷ fused

**Double precision**
| Np | 1T | 2T | 4T | 6T |
|---:|:---:|:---:|:---:|:---:|
| 1k* | 0.93× | 1.24× | 1.26× | 1.54× |
| 10k* | 0.91× | 1.09× | 1.11× | 1.26× |
| 100k | 0.90× | 0.91× | 0.95× | 0.99× |
| 1M | 0.99× | 0.96× | 1.15× | **1.30×** |
| 10M | 0.99× | 0.97× | 1.21× | **1.55×** |

**Single precision**
| Np | 1T | 2T | 4T | 6T |
|---:|:---:|:---:|:---:|:---:|
| 1k* | 1.06× | 1.04× | 1.27× | 1.33× |
| 10k* | 1.05× | 1.14× | 1.23× | 1.44× |
| 100k | 1.12× | 1.15× | 1.53×\* | 0.88×\* |
| 1M | 1.12× | 1.13× | 1.13× | 1.16× |
| 10M | 1.14× | 1.09× | 1.10× | **1.22×** |

Headline raw times at 6 threads: DP 10M `26.6 → 17.2 ms`, DP 1M `2.22 → 1.71 ms`, SP 10M `18.0 → 14.8 ms`.

\*The 1k / 10k rows and the 100k-SP 4T/6T points are ≤ 0.2 ms and **overhead-/noise-dominated** (the ~70-entry result-dict marshalling + reduction launch + thread spawn, and — for two-pass — *two* `ParticleReduce` launches vs the fused's one). They are not a reliable reduction signal.

#### Interpretation — a memory-bandwidth optimization, so it is cache-size-gated

The fused pass reads the particle SoA **once** instead of twice, which only helps once the beam **spills out of L3** (~24 MB here) and the reduction becomes **DRAM-bandwidth-bound**:

- **≤ 100k particles (≈ 9 MB, L3-resident):** not bandwidth-bound → fused ≈ two-pass (DP even slightly slower, as at 1 thread; SP a touch faster). Halving the passes buys nothing when the data already sits in cache.
- **≳ 1M particles (≫ L3):** DRAM-bandwidth-bound → the fused wins, and the win **grows with thread count** as DDR5 saturates — DP **1.30× @1M / 1.55× @10M** at 6 P-cores. (Note two-pass DP 10M barely improves from 4→6T, `31.4 → 26.6 ms`, while fused keeps scaling, `25.9 → 17.2 ms`.) SP is ~1.1–1.22×.
- **1 thread, any size:** not bandwidth-bound → DP ~4–8% slower, SP ~1.1× faster — exactly the CodSpeed regime.

**Net:** the optimization targets **large beams (≳ 1M) tracked on multiple cores** — the regime where a diagnostics-dominated run (#1102) actually hurts. cachegrind runs single-threaded with fixed cache-miss penalties, so it structurally cannot see this multi-core bandwidth win; hence the single-thread DP flag despite the real speedup at scale.

ax3l:
### GPU single-precision — the same fusion, on the local GPU

Companion to the CPU wall-clock benchmark (https://github.com/BLAST-ImpactX/impactx/pull/1561#issuecomment-4963809369). Same before/after (two-pass `f3df88d51^` → fused `f3df88d51`), now on GPU.

**Environment**
- GPU: NVIDIA RTX A2000 8 GB Laptop (sm_86), single GPU, single MPI rank.
- Build: `-DImpactX_COMPUTE=CUDA -DImpactX_PRECISION=SINGLE` (RelWithDebInfo), AMReX 26.06.
- Metric: min wall-clock over rounds (the reduction's `value()` synchronizes the device).

| Np | two-pass (ns/particle) | fused (ns/particle) | speedup |
|---:|---:|---:|:---:|
| 100k | 1.16 | 1.47 | 0.79× |
| 1M | 0.53 | 0.38 | **1.41×** |
| 10M | 0.47 | 0.25 | **1.86×** |
| 100M | 0.46 | 0.23 | **1.99×** |

Raw times: 100M `46.4 → 23.3 ms`, 10M `4.69 → 2.53 ms`, 1M `0.53 → 0.38 ms`.

**Takeaway:** on GPU the fusion is a clean win at scale — **~2× at 10M–100M, 1.41× at 1M**. A GPU is bandwidth-rich and massively parallel, so reading the particle SoA **once instead of twice** converts almost directly into the theoretical 2×. It only loses at **100k**, where the kernel is tiny and dominated by fixed overhead (kernel launch + the wider 46-value `ReduceData` + the shift sample's `dtoh_memcpy`/broadcast) rather than the reduction itself. This is exactly the regime that matters for the ≳1M-particle dispersion studies on GPU.

ax3l:
@cemitch99 I think this is ready. I will do a follow-up PR where I will add the ability to sub-select moment groups, which should give a 6-9x speedup, e.g., when only reducing transverse moments.

ax3l:
ping @cemitch99 @atmyers @WeiqunZhang @EZoni or your agents for review :)

src/diagnostics/ReducedBeamCharacteristics.cpp
```diff
@@ -47,263 +50,227 @@ namespace impactx::diagnostics
...
-        amrex::ParticleReal const min_t = values_per_rank_min.at(2);
-        amrex::ParticleReal const min_px = values_per_rank_min.at(3);
-        amrex::ParticleReal const min_py = values_per_rank_min.at(4);
-        amrex::ParticleReal const min_pt = values_per_rank_min.at(5);
+        amrex::ParticleReal const min_x = values_min.at(0);
+        amrex::ParticleReal const min_y = values_min.at(1);
+        amrex::ParticleReal const min_t = values_min.at(2);
+        amrex::ParticleReal const min_px = values_min.at(3);
+        amrex::ParticleReal const min_py = values_min.at(4);
+        amrex::ParticleReal const min_pt = values_min.at(5);
```
cemitch99, line 239: Note:  I believe none of our existing tests access rbc min and max values, so if there is any change that results from the shift implemented here, it could slip through.

cemitch99 (commented):
I ran a test on my Mac using `examples/iota_lattice/input_iotalattice_sdep.in`, and all of the columns in reduced_beam_characteristics.. are unaffected by this PR to within 1e-16 (absolute change).    The Twiss functions show differences at the level 1e-12.

cemitch99 (approved):

PR #1568, opened 2026-07-22
Update Xopt Examples & CI

ax3l:
The xopt examples are outdated with new APIs in recent Xopt 3+ versions.
Also, CI did not exercise them.


cemitch99 (approved):
Thanks!

PR #1570, opened 2026-07-24
`BeamMonitor`: gyromagnetic anomaly

ax3l:
Write the anomalous magnetic dipole moment of the reference particle to openPMD beam monitor data, too.

Needed to restore for the reference particle in a follow-up PR.


cemitch99 (approved):
Thanks!  Looks good once the issue with Windows test is resolved.

PR #1572, opened 2026-07-24
Spin push for ChrAcc element

cemitch99:
This PR adds the spin push through the ChrAcc element (uniform Ez + Bz).

- [x] develop analytical spin map
- [x] implement spin push
- [x] add Bz=0 depolarization benchmark
- [x] add Ez=0 solenoid comparison
- [x] add nonzero Ez & Bz comparison (isotropic beam)

Benchmark ideas:

- when Ez=0 and energy spread is small, the result must coincide with a Sol push
- when Bz=0, the spin transport becomes simple and the polarization can be predicted
- when the initial beam is X-Y isotropic, the polarization can be predicted

The first test above relies on PR #1577


src/elements/ChrUniformAcc.H
```diff
@@ -274,6 +276,99 @@ namespace impactx::elements
...
+            T_Real & AMREX_RESTRICT y,
+            T_Real & AMREX_RESTRICT t,
+            T_Real & AMREX_RESTRICT px,
+            T_Real & AMREX_RESTRICT py,
+            T_Real & AMREX_RESTRICT pt,
+            T_Real & AMREX_RESTRICT sx,
+            T_Real & AMREX_RESTRICT sy,
+            T_Real & AMREX_RESTRICT sz,
+            T_IdCpu & AMREX_RESTRICT idcpu,
+            [[maybe_unused]] RefPart const & AMREX_RESTRICT refpart
```
ax3l:
```suggestion
            RefPart const & AMREX_RESTRICT refpart
```

qianglbl (approved):

examples/spin_tracking/analysis_chracc_sol.py
```diff
@@ -0,0 +1,43 @@
...
+szf = final["spin_z"]
+
+dspin2 = (sxf - sxi) ** 2 + (syf - syi) ** 2 + (szf - szi) ** 2
+dspin = np.sqrt(dspin2)
+dspinmax = dspin.max()
+
+print("Change in the spin:")
+print("||delta s||_max", dspinmax)
+
+atol = 2.0e-7
```
cemitch99:
```suggestion
atol = 2.1e-7
```

examples/spin_tracking/README.rst
```diff
@@ -597,3 +597,86 @@ The analysis can be run using:
...
+       .. literalinclude:: run_chracc_depolarization.py
+          :language: python3
+          :caption: You can copy this file from ``examples/spin_tracking/run_chracc_depolarization.py``.
+
+Analyze
+-------
+
+The analysis can be run using:
+
+.. dropdown:: Script ``analysis_cfbend_spin.py``
```
cemitch99:
```suggestion
.. dropdown:: Script ``analysis_chracc_depolarization.py``
```

examples/spin_tracking/README.rst
```diff
@@ -597,3 +597,86 @@ The analysis can be run using:
...
+
+   .. literalinclude:: analysis_cfbend_spin.py
```
cemitch99:
```suggestion
   .. literalinclude:: analysis_chracc_depolarization.py
```

examples/spin_tracking/README.rst
```diff
@@ -597,3 +597,86 @@ The analysis can be run using:
    .. literalinclude:: analysis_reversibility_quad_fringe_spin.py
       :language: python3
       :caption: You can copy this file from ``examples/spin_tracking/analysis_reversibility_quad_fringe_spin.py``.
+
+
+.. _examples-chracc-depol:
+
+Spin Depolarization in a Uniform Accelerating Field
+====================================================
```
```suggestion
Spin Depolarization in a Uniform Accelerating Field
===================================================
```

src/elements/ChrUniformAcc.H
```diff
@@ -288,6 +290,100 @@ namespace impactx::elements
...
+
+                // compute focusing constant (1/m) and rotation angle (in rad)
+                T_Real const theta_over_alpha = (m_ez==0_prt)? m_slice_ds/pzi_tot : log(numer / denom)/m_ez;
+                T_Real const theta = m_alpha * theta_over_alpha;
+                auto const [sin2Gth, cos2Gth] = amrex::Math::sincos(2_prt * refpart.gyromagnetic_anomaly * theta);
+
+                T_Real const mu = 0.5_prt * (pzf_tot / (1_prt - ptf_tot) - pzi_tot / (1_prt - pti_tot));
+
+                T_Real const Ffunc = (m_bz==0_prt)? 0_prt : pzi_tot/(1_prt-pti_tot)*sin2Gth + m_ez*(1_prt-cos2Gth)/m_bz + 2_prt*mu*sin2Gth;
+                T_Real const Gfunc = (m_bz==0_prt)? -2_prt*mu - m_ez*refpart.gyromagnetic_anomaly*theta_over_alpha :
```
ax3l, line 365: We discussed this today, just highlighting that the general case where Ez and Bz are both non-zero is not tested at all.
(Even if no analytic solution exist, we could cover it in some test at least to ensure an expected converged result.)

examples/spin_tracking/run_chracc_depolarization.py
```diff
@@ -0,0 +1,88 @@
...
+sim.spin = True
+sim.slice_step_diagnostics = True
+
+# domain decomposition & space charge mesh
+sim.init_grids()
+
+# basic beam parameters
+kin_energy_MeV = 20.00  # reference energy (kinetic)
+bunch_charge_C = 25.0e-12  # used with space charge
+npart = 1000000  # number of macro particles
```
ax3l: Codex comment:

  2. P2 — Particle-count mismatch weakens the statistical test and makes it unnecessarily slow.
     run_chracc_depolarization.py (https://github.com/BLAST-ImpactX/impactx/blob/94933a07be6be5d72b1446d9c0714091c99fa4c5/examples/spin_tracking/run_chracc_depolarization.py#L22-L25) generates
     1,000,000 particles, but analysis_chracc_depolarization.py
     (https://github.com/BLAST-ImpactX/impactx/blob/94933a07be6be5d72b1446d9c0714091c99fa4c5/examples/spin_tracking/analysis_chracc_depolarization.py#L83-L86) computes tolerance using 100,000. That
     makes the tolerance √10 too loose. CI also shows the run taking 41.79 seconds, above the repository’s ideal 30-second target. Setting npart=100000 looks like the likely intended fix;
     alternatively derive and assert the actual particle count in the analysis.

ax3l:
```suggestion
npart = 100000  # number of macro particles
```
or for people like me that cannot count :D
```suggestion
npart = int(1e5)  # number of macro particles
```
or

```suggestion
npart = 100_000  # number of macro particles
```

examples/spin_tracking/analysis_chracc_depolarization.py
```diff
@@ -0,0 +1,155 @@
...
+atol = 1.0e-12 if is_double else 1.0e-3
+print(f"  atol={atol}")
+
+assert np.allclose(
+    [condition],
+    [
+        1.0,
+    ],
+    atol=atol,
+)
```
ax3l, line 156: Codex:

  3. P3 — The claimed 1e-12 norm tolerance is not enforced.
     The final np.allclose (https://github.com/BLAST-ImpactX/impactx/blob/94933a07be6be5d72b1446d9c0714091c99fa4c5/examples/spin_tracking/analysis_chracc_depolarization.py#L146-L155) leaves the
     default rtol=1e-5, so the double-precision check effectively accepts errors around 1e-5, not 1e-12. Add rtol=0.0.

examples/spin_tracking/analysis_chracc_depolarization.py
```diff
@@ -0,0 +1,155 @@
...
+sigma_sx = rbc["sigma_sx"].iloc[-1]
+sigma_sy = rbc["sigma_sy"].iloc[-1]
+sigma_sz = rbc["sigma_sz"].iloc[-1]
+polarization = np.sqrt(polarization_x**2 + polarization_y**2 + polarization_z**2)
+condition = sigma_sx**2 + sigma_sy**2 + sigma_sz**2 + polarization**2
+
+print("")
+print(f"Spin moment consistency condition = {condition:e}")
+
+atol = 1.0e-12 if is_double else 1.0e-3
```
cemitch99, line 146:
```suggestion
atol = 1.0e-12 if is_double else 1.0e-3
rtol = 0.0
```

src/elements/ChrUniformAcc.H
cemitch99: This case can be done (when the initial beam is X-Y isotropic).  The analytical result is more complicated, but after some consideration I think it's worth the effort to include it with this PR.  I will add it.

examples/spin_tracking/analysis_chracc_sol.py
```diff
@@ -0,0 +1,43 @@
...
+
+sxi = initial["spin_x"]
+syi = initial["spin_y"]
+szi = initial["spin_z"]
+
+sxf = final["spin_x"]
+syf = final["spin_y"]
+szf = final["spin_z"]
+
+dspin2 = (sxf - sxi) ** 2 + (syf - syi) ** 2 + (szf - szi) ** 2
```
ax3l: Codex:

  - In analysis_chracc_sol.py (https://github.com/BLAST-ImpactX/impactx/blob/94933a07be6be5d72b1446d9c0714091c99fa4c5/examples/spin_tracking/analysis_chracc_sol.py#L15-L27), match initial/final
    particles by id before subtracting, as the newer reversibility test does. This avoids relying on openPMD row ordering.

examples/spin_tracking/run_chracc_sol.py
```diff
@@ -0,0 +1,85 @@
...
+monitor = elements.BeamMonitor("monitor", backend="h5")
+
+# bend radius (> 0)
+ks_value = 1.0
+print(f"ks_value={ks_value}")
+
+# length for this test should be one period
+ds_value = 2.0 * np.pi / (ref.gyromagnetic_anomaly * ks_value)
+
+sol1 = elements.Sol(name="sol1", ds=ds_value, ks=ks_value, nslice=ns)
```
ax3l: Unused:

```suggestion
# sol1 = elements.Sol(name="sol1", ds=ds_value, ks=ks_value, nslice=ns)
```

src/elements/ChrUniformAcc.H
```diff
@@ -288,6 +290,100 @@ namespace impactx::elements
...
+            amrex::ParticleIDWrapper<T_IdCpu>{idcpu}.make_invalid(mask);
+            {
+                T_Real const pzi_tot = sqrt(powi<2>(pti_tot) - 1_prt);
+                T_Real const pzf_tot = sqrt(powi<2>(ptf_tot) - 1_prt);
+
+                T_Real const numer = -ptf_tot + pzf_tot;
+                T_Real const denom = -pti_tot + pzi_tot;
+
+                // compute focusing constant (1/m) and rotation angle (in rad)
+                T_Real const theta_over_alpha = (m_ez==0_prt)? m_slice_ds/pzi_tot : log(numer / denom)/m_ez;
```
ax3l: Unusually long line:
```suggestion
                T_Real const theta_over_alpha = (m_ez == 0_prt) ?
                    m_slice_ds / pzi_tot :
                    log(numer / denom) / m_ez;
```

examples/spin_tracking/analysis_chracc_depolarization.py
```diff
@@ -0,0 +1,156 @@
...
+gamma_i = beam_initial.get_attribute("gamma_ref")
+beta_gamma_i = beam_initial.get_attribute("beta_gamma_ref")
+beam_final = series.iterations[last_step].particles["beam"]
+gamma_f = beam_final.get_attribute("gamma_ref")
+beta_gamma_f = beam_final.get_attribute("beta_gamma_ref")
+
+# numerical parameters based on input file
+gyro_anomaly = 1.7928473446  # for protons
+ez_value = 10.0  # accelerating field in 1/m (dgamma/dz) from input
+ds_value = 50.0  # accelerating distance in m from input
```
ax3l: Unused:
```suggestion
```

examples/spin_tracking/analysis_chracc_depolarization.py
ax3l: We just realized `np.allclose` has some fun defaults:
> numpy.allclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False)
>
> ...
> If the following equation is element-wise True, then allclose returns True.:
>
> absolute(a - b) <= (atol + rtol * absolute(b))

https://numpy.org/doc/stable/reference/generated/numpy.allclose.html

cemitch99: Fixed.

examples/spin_tracking/analysis_chracc_sol.py
cemitch99: What is the "newer reversibility test" that Codex points to here?

cemitch99: Just added a sort on id to address this issue.

ax3l: > What is the "newer reversibility test" that Codex points to here?

That is the spin addition for `ChrAcc` in `tests/python/test_reversibility_elements.py` line 294ff:
```py
@pytest.mark.parametrize("sim", [True, False], indirect=True, ids=["spin", "nospin"])
def test_ChrAcc(sim):
    # ...
```

Ah wait, it probably means an other analysis script.

ax3l (approved):
Thank you, this looks great!

Issue #1573, opened 2026-07-26
Unexpected t-update in ChrAcc when accelerating field is small


cemitch99:
Comparison between the elements `Sol `and `ChrAcc` (for negligible but nonzero values of Ez) shows good agreement for all phase space coordinates except t.  The t-update shows surprising behavior.

<img width="2666" height="2000" alt="Image" src="https://github.com/user-attachments/assets/2929dbd6-e5b8-4628-88af-561d913ac8f9" />

cemitch99:
This is due to a small-denominator issue that occurs in the t-update when Ez is sufficiently small (e.g. in this example, qEz/mc^2 = 1.0e-12 results in this behavior, while qEz/mc^2 = 1.0e-9 does not).  A separate branch needs to be added for the special case Ez = 0, and the case of extremely small Ez may either collapse to this case or require a third branch using Taylor expansion.

PR #1577, opened 2026-07-27
Treat special case Ez=0 for ChrAcc element.

cemitch99:
The ChrAcc element map (uniform Ez + Bz field) requires special handling for the edge case that Ez=0.  This PR adds the necessary changes.

- [x] modify push
- [x] add a benchmark test

This partially addresses Issue #1573

### DIFF
--- a/src/elements/ChrUniformAcc.H
+++ b/src/elements/ChrUniformAcc.H
@@ -111,7 +111,6 @@ namespace impactx::elements
 
             // compute focusing constant (1/m) and rotation angle (in rad)
             m_alpha = m_bz * 0.5_prt;
-            m_alpha_iez = m_ez == 0_prt ? 0_prt : m_alpha / m_ez;
 
             // dimensionless acceleration parameter (change in gamma)
             m_acc = m_ez * m_slice_ds;
@@ -172,11 +171,13 @@ namespace impactx::elements
                 T_Real const denom = -pti_tot + pzi_tot;
 
                 // compute focusing constant (1/m) and rotation angle (in rad)
-                T_Real const theta = (m_acc == 0_prt)? m_alpha * m_slice_ds / pzi_tot : m_alpha_iez * log(numer / denom);
+                T_Real const theta_ialpha = m_acc == 0_prt
+                    ? m_slice_ds / pzi_tot
+                    : log(numer / denom) / m_ez;
+                T_Real const theta = m_alpha * theta_ialpha;
                 auto const [sin_theta, cos_theta] = amrex::Math::sincos(theta);
                 T_Real const sin_theta_ialpha = m_alpha == 0_prt
-                    ? (m_acc == 0_prt ? m_slice_ds / pzi_tot
-                                      : log(numer / denom) / m_ez)
+                    ? theta_ialpha
                     : sin_theta / m_alpha;
 
                 // initialize output values
@@ -423,7 +424,7 @@ namespace impactx::elements
         // see: compute_constants() to refresh
         amrex::ParticleReal m_slice_ds;  //! m_ds / nslice();
         amrex::ParticleReal m_ptf_ref, m_pti_ref, m_bgf, m_bgi;
-        amrex::ParticleReal m_alpha, m_alpha_iez, m_acc;
+        amrex::ParticleReal m_alpha, m_acc;
     };
 
 } // namespace impactx
--- a/src/elements/ChrUniformAcc.H
+++ b/src/elements/ChrUniformAcc.H
@@ -194,7 +194,7 @@ namespace impactx::elements
                 T_Real Pxy_mech2 = powi<2>(py - m_alpha * x) + powi<2>(px + m_alpha * y);
                 if (m_acc == 0_prt) {
                    // In this case, t-update coincides with that of a ChrDrift:
-                   tout = t - (pti_tot/pzi_tot + 0.5_prt 
+                   tout = t - (pti_tot/pzi_tot + 0.5_prt
                           * pti_tot/powi<3>(pzi_tot) * Pxy_mech2) * m_slice_ds;
                    tout += m_pti_ref/m_bgi * m_slice_ds;
                 } else {
--- a/src/elements/ChrUniformAcc.H
+++ b/src/elements/ChrUniformAcc.H
@@ -174,6 +174,10 @@ namespace impactx::elements
                 // compute focusing constant (1/m) and rotation angle (in rad)
                 T_Real const theta = (m_acc == 0_prt)? m_alpha * m_slice_ds / pzi_tot : m_alpha_iez * log(numer / denom);
                 auto const [sin_theta, cos_theta] = amrex::Math::sincos(theta);
+                T_Real const sin_theta_ialpha = m_alpha == 0_prt
+                    ? (m_acc == 0_prt ? m_slice_ds / pzi_tot
+                                      : log(numer / denom) / m_ez)
+                    : sin_theta / m_alpha;
 
                 // initialize output values
                 T_Real xout = x;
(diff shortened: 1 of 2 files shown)


tests/python/test_chracc_Ez0.py
```diff
@@ -0,0 +1,115 @@
...
+            dpt,
+            qm_eev,
+            bunch_charge=0.0,
+        )
+
+    # store initial particles
+    df_initial = pc.to_df()
+
+    # design the accelerator lattice
+    ns = 50  # number of slices per ds in the element
```
cemitch99:
```suggestion
    ns = 1  # number of slices per ds in the element
```

tests/python/test_chracc_Ez0.py
```diff
@@ -0,0 +1,115 @@
...
+    # set numerical parameters and IO control
+    sim.space_charge = False
+    sim.slice_step_diagnostics = True
+
+    # domain decomposition & space charge mesh
+    sim.init_grids()
+
+    # basic beam parameters
+    kin_energy_MeV = 2000.0  # reference energy (kinetic)
+    bunch_charge_C = 25.0e-12  # used with space charge
```
cemitch99:
```suggestion
```

src/elements/ChrUniformAcc.H
```diff
@@ -112,6 +112,9 @@ namespace impactx::elements
             // compute focusing constant (1/m) and rotation angle (in rad)
             m_alpha = m_bz * 0.5_prt;
             m_alpha_iez = m_alpha / m_ez;
+
+            // dimensionless acceleration parameter (change in gamma)
+            m_acc = m_ez * m_slice_ds;
```
ax3l, line 116: Minor note (ok to ignore): `m_alpha` and `m_acc` are probably not worth caching and would be ok to recompute for every particle, but also does not hurt as we benchmarked recently.

src/elements/ChrUniformAcc.H
ax3l: Probably a bug: In the reference particle, there is still a division by zero here and the zero-path is not treated.

tests/python/test_chracc_Ez0.py
```diff
@@ -0,0 +1,114 @@
...
+        "momentum_y",
+        "momentum_t",
+    ]
+
+    for c in PHASE_COLS:
+        np.testing.assert_allclose(
+            df_final[c].to_numpy(),
+            df_initial[c].to_numpy(),
+            atol=1.0e-12,
+            rtol=0,
```
ax3l, line 167: Here, can you also assert the reference particle state for correctness?

src/elements/ChrUniformAcc.H
ax3l: Probably same bug: The transport map still has a division by zero / no treatment of the zero case.

cemitch99: Yes, good catch.  Will fix.

tests/python/test_chracc_Ez0.py
cemitch99: Will do.

src/elements/ChrUniformAcc.H
cemitch99: Fixed.

cemitch99:
@ax3l Thanks for the comments.  Suggestions addressed; ready for re-review.

src/elements/ChrUniformAcc.H
ax3l: Did you mean to comment in line 270 and remove line 271? :)

ax3l: fixed

cemitch99: Fixed.

src/elements/ChrUniformAcc.H
```diff
@@ -169,7 +172,7 @@
                 T_Real const denom = -pti_tot + pzi_tot;
 
                 // compute focusing constant (1/m) and rotation angle (in rad)
-                T_Real const theta = m_alpha_iez * log(numer / denom);
+                T_Real const theta = (m_acc == 0_prt)? m_alpha * m_slice_ds / pzi_tot : m_alpha_iez * log(numer / denom);
```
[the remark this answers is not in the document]
ax3l: Ok for now. We could use a machine epsilon as in `abs(m_acc) < machine eps` to make sure simple preparatory calculations that end up close to zero are also hitting here. But for now this is ok, for the special case that is directly set.

src/elements/ChrUniformAcc.H
```diff
@@ -169,7 +172,7 @@ namespace impactx::elements
...
```
ax3l: Codex:

  - `Ez=0, Bz=0` still produces NaNs on this line (175-190)
    Here theta=0 and m_alpha=0, so the subsequent sin_theta / m_alpha evaluates
    as 0/0. An `alpha==0` branch or stable `sinc` formulation is needed.

src/elements/ChrUniformAcc.H
```diff
@@ -112,6 +112,9 @@ namespace impactx::elements
             // compute focusing constant (1/m) and rotation angle (in rad)
             m_alpha = m_bz * 0.5_prt;
             m_alpha_iez = m_alpha / m_ez;
```
ax3l: Codex (is right):

  - [P2] The zero-field path still evaluates a division by zero.
    `m_alpha_iez = m_alpha / m_ez` executes _bef_ore the new branch. It leaves an
    infinity in normal execution and aborts when the documented
    `amrex.fpe_trap_zero` debugging option is enabled. Compute it conditionally
    or restructure around the finite integrated-drift factor.

tests/python/test_chracc_Ez0.py
```diff
@@ -0,0 +1,129 @@
...
+    sim.init_grids()
+
+    # basic beam parameters
+    kin_energy_MeV = 2000.0  # reference energy (kinetic)
+
+    # set reference particle
+    ref = sim.beam.ref
+    ref.set_species("proton").set_kin_energy_MeV(kin_energy_MeV)
+    qm_eev = ref.charge_qe / (ref.mass_MeV * 1.0e6)  # electron charge/mass in e / eV
+    initial_ref = ref
```
ax3l: Codex:

  - [P2] The reference-particle regression assertion is tautological —
    line 31-35

    `beam.ref` is a live mutable object, so `initial_ref = ref` aliases it. The
    final loop compares the reference particle with itself. Use `initial_ref = ref.copy()`, as the existing reversibility tests do.

ax3l (approved):
I was about to approve this but Codex found a few more bugs.

>  The exact Ez → 0 formulas for particle t, reference t, drift_log, and R56
  otherwise look correct. I’d also consider adding nonzero transverse
  coordinates and a linear-map check; the current test does not exercise the
  new transverse Pxy_mech2 term or transport_map.

ax3l:
@cemitch99 I can add commits for those fixes and you can review them quickly.

src/elements/ChrUniformAcc.H
cemitch99: This quantity is not used at all when `m_ez=0`, so we can assign zero in this case.  Otherwise, it might be easier simply to remove it, and accept the division cost.

ax3l: fixed via 3d0c69c976873b8f04d9219de94a81a662df6abf

tests/python/test_chracc_Ez0.py
ax3l: fixed via b1ba61a5898b65777550b1cda7de474330e35d75

test expanded via 40d62d431b5a44fd871244850e8929e0b53b03d4

src/elements/ChrUniformAcc.H
ax3l: fix via e1a4ff35b14609b28c535dd0ed8ffd19c2b9419f

src/elements/ChrUniformAcc.H
```diff
@@ -169,8 +172,12 @@ namespace impactx::elements
                 T_Real const denom = -pti_tot + pzi_tot;
 
                 // compute focusing constant (1/m) and rotation angle (in rad)
-                T_Real const theta = m_alpha_iez * log(numer / denom);
+                T_Real const theta = (m_acc == 0_prt)? m_alpha * m_slice_ds / pzi_tot : m_alpha_iez * log(numer / denom);
                 auto const [sin_theta, cos_theta] = amrex::Math::sincos(theta);
+                T_Real const sin_theta_ialpha = m_alpha == 0_prt
+                    ? (m_acc == 0_prt ? m_slice_ds / pzi_tot
+                                      : log(numer / denom) / m_ez)
+                    : sin_theta / m_alpha;
```
ax3l, line 181: If this nested branching is too tricky to read, then we can define another intermediate variable.

ax3l: done

ax3l (approved):

PR #1578, opened 2026-07-28
Warn if too Many (or too Few) Boxes Per Process

ax3l:
Creating too many boxes is a common way to tank AMReX performance, especially on single-rank simulations. Warns now about the most common problems that cause costly and unnecessary particle redistribution, extra communication in SC solves, and extra kernels on GPU.

We have a similar warning in WarpX for a while and I just this week fooled myself again in ImpactX by forgetting about this in benchmarks.

Reviewing by commit:
- C++ logic for warning b80f4f506cde212a8857990aff9fd3cf461a11ae
- Documentation 7e5f320280d5fbd184ea6d1b27eb9e791f53ec7b
- Large example change 36d90b1599026b2b6ae1e3e170d18b187c678d81

is the easiest for this PR.

### DIFF
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -2368,8 +2368,8 @@ See there ``nslice`` option on lattice elements for slicing.
     (grid) per MPI process. One box per process is the ideal decomposition for
     every simulation, including space charge calculations. The automatically
     chosen mesh size is only suitable for simulations that need no auxiliary
-    field grids, e.g., without space charge calculations, set this parameter
-    to control the field resolution otherwise.
+    field grids, e.g., without space charge calculations. Otherwise, set this
+    parameter to control the field resolution.
 
     If it is set, the mesh is chopped into boxes of at
     most :pp:param:`amr.max_grid_size` cells per direction, independent of the
@@ -2378,10 +2378,12 @@ See there ``nslice`` option on lattice elements for slicing.
 .. pp:param:: amr.max_grid_size
     :type: ``integer, per MR level``
     :optional:
-    :default: ``32``
+    :default: ``32`` on CPU, ``64`` on GPU
 
     The maximum number of grid points per box (`grid <https://amrex-codes.github.io/amrex/docs_html/GridCreation.html>`__) along each direction.
     The mesh on each mesh-refinement level is decomposed into boxes of at most this size.
+
+    If :pp:param:`amr.n_cell` is not set, ImpactX overwrites this default with the `blocking factor <https://amrex-codes.github.io/amrex/docs_html/GridCreation.html>`__ (``8``), which creates exactly one box per MPI process.
 
     For best performance, aim for **exactly one box per parallel (MPI) process**, which is also one box per GPU on accelerated machines.
     More boxes per process cause significant overhead (particle redistribution, communication in field solves, extra kernel launches).
@@ -2395,6 +2397,8 @@ See there ``nslice`` option on lattice elements for slicing.
 
     The maximum number of grid points per box along ``x``, overwriting :pp:param:`amr.max_grid_size`.
     Analogous parameters exist as ``amr.max_grid_size_y`` and ``amr.max_grid_size_z``.
+
+    The per-direction parameters always win over the isotropic :pp:param:`amr.max_grid_size`, independent of the order in which the two are given.
 
 .. pp:param:: amr.max_level
     :type: ``integer``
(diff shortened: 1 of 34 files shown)


ax3l:
Looking at the examples, porting the `warpx.numprocs` option could be a good follow-up PR.
https://warpx.readthedocs.io/en/latest/usage/parameters.html#distribution-across-mpi-ranks-and-parallelization

ax3l:
ping @WeiqunZhang ready for review :)

WeiqunZhang:
1. The new CMake block is unreachable dead code — examples/CMakeLists.txt

```
  if(is_mpi AND NOT ImpactX_MPI)
      set(THIS_SERIAL_DECOMP amr.max_grid_size=512 ...)
  endif()
```

  add_impactx_test() already returns early on exactly this condition:

```
  # examples/CMakeLists.txt:50-53
  # cannot run MPI tests w/o MPI support
  if(is_mpi AND NOT ImpactX_MPI)
      return()
  endif()
```

  So the condition can never be true by the time the else() branch is reached. The block's comment — "In non-MPI builds, MPI-parallel tests run on a single process" — is also factually wrong for
  this repo: those tests aren't registered at all. This explains why "GCC w/ SIMD w/o MPI w/ Python" passes with or without the block.

  Either drop the block, or — if the intent is to actually start running MPI-marked tests serially in non-MPI builds — remove the early return() and say so in the PR. Note the second option has a
  trap (below).

  2. If that block were made live, it would break the one_box_per_rank path

  initialization::one_box_per_rank() (src/initialization/InitAmrCore.cpp:114-125) gates on presence, not value:

```
  bool const has_max_grid_size =
          pp_amr.countname("max_grid_size") > 0 || ... ;
  if (!has_max_grid_size) {
      pp_amr.addarr("max_grid_size", bf_lvl0);   // forces exactly one box per rank
```

  An unconditional amr.max_grid_size=512 on the command line suppresses that, collapsing the blocking_factor × nprocs domain into a single box — which under MPI trips the new "fewer boxes than
  processes" warning and aborts. If the override is resurrected, it must apply only to inputs that actually set amr.n_cell.

  3. Documented default for amr.max_grid_size is incomplete — docs/source/usage/parameters.rst

  :default: 32 holds only for CPU builds. AMReX (AMReX_AmrMesh.H:33-38):

```
  #if defined(AMREX_USE_GPU)
      Vector<IntVect> max_grid_size {{IntVect(AMREX_D_PICK(128,128,64))}};
  #else
      Vector<IntVect> max_grid_size {{IntVect(AMREX_D_PICK(128,128,32))}};
  #endif
```

  And when amr.n_cell is not set, ImpactX overrides it to amr.blocking_factor (8). Since the whole point of this PR is getting users to compute their box count correctly, a GPU user reading "32"
  will get it wrong. Suggest: 32 (CPU) / 64 (GPU); overridden to amr.blocking_factor when amr.n_cell is not set.

  4. sim.max_grid_size cannot undo what blocking_factor_* wrote

  The Python docs correctly warn that blocking_factor_x also sets max_grid_size_x, and say to "assign max_grid_size_x afterwards to overwrite". But assigning the new all-direction
  sim.max_grid_size afterwards will not work — AMReX parses max_grid_size first, then lets max_grid_size_x/_y/_z override it (AMReX_AmrMesh.cpp:237-249), independent of ParmParse insertion order.
  So:

```
  sim.blocking_factor_x = [16]
  sim.max_grid_size = [72]      # x stays at 16
```

  Worth stating explicitly in both python.rst and the max_grid_size docstring, since that property is the new headline API and this is precisely the foot-gun the warning will now flag. (I checked
  all touched examples — none hit this; they either use per-direction values throughout or set no blocking factor.)
  Worth stating explicitly in both python.rst and the max_grid_size docstring, since that property is the new headline API and this is precisely the foot-gun the warning will now flag. (I checked
  all touched examples — none hit this; they either use per-direction values throughout or set no blocking factor.)

  5. Garbled sentence in the amr.n_cell docs

  > The automatically chosen mesh size is only suitable for simulations that need no auxiliary field grids, e.g., without space charge calculations, set this parameter to control the field resolution otherwise.

  Comma splice — two sentences fused. Suggest: "The automatically chosen mesh size is only suitable for simulations that need no auxiliary field grids, e.g., without space charge. Otherwise, set 
  this parameter to control the field resolution."

WeiqunZhang (approved):

cemitch99 (approved):

EZoni:
@WeiqunZhang 

What should we make of the review in https://github.com/BLAST-ImpactX/impactx/pull/1578#issuecomment-5185137350, given that you approved the PR? Do you think the issues raised in the comment are false positives or do you think they should be addressed in a separate PR?

WeiqunZhang:
I think they are non-blocking issues that could be addressed here or in a a separate PR.

ax3l:
Thanks! I agree they are non-blocking but I will address them here and then merge.

PR #1581, opened 2026-07-30
`np.allclose` Defaults

ax3l:
We just realized `np.allclose` has non-zero defaults for both `atol` and `rtol`. Fixing where we assumed they default to zero.

### DIFF
--- a/examples/rfcavity/analysis_rfcavity_ref_part_opt.py
+++ b/examples/rfcavity/analysis_rfcavity_ref_part_opt.py
@@ -63,7 +63,7 @@ print("")
 print("Initial Beam:")
 print(f"  s_ref={si:e} gamma_ref={gammai:e}")
 
-atol = 1.0e-4
+atol = 2.5e-4
 print(f"  atol={atol}")
 
 assert np.allclose(
--- a/examples/rfcavity/analysis_rfcavity_ref_part_opt.py
+++ b/examples/rfcavity/analysis_rfcavity_ref_part_opt.py
@@ -63,7 +63,7 @@ print("")
 print("Initial Beam:")
 print(f"  s_ref={si:e} gamma_ref={gammai:e}")
 
-atol = 2.5e-4
+atol = 1.0e-4
 print(f"  atol={atol}")
 
 assert np.allclose(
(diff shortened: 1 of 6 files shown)


cemitch99 (commented):
The changes look good.  I think it's safe to assume that one or more of these tests will need to have tolerance loosened, or revisited.  Will wait for CI to see what needs adjustment.

cemitch99:
Nice!  Only one test fails:  `rfcavity-ref-part-opt.py`.  Will suggest a change.

examples/rfcavity/analysis_rfcavity_ref_part_opt.py
```diff
@@ -63,7 +63,7 @@ def data_is_double(file_pattern):
 print("Initial Beam:")
 print(f"  s_ref={si:e} gamma_ref={gammai:e}")
 
-atol = 1.0e-4  # ignored
+atol = 1.0e-4
```
cemitch99:
```suggestion
atol = 2.5e-4
```

cemitch99: This should work.

ax3l: Oops, this is the initial line. The tolerance needs to be relaxed on the final beam line. Will update.

ax3l:
Worth a look separately: that 6e-6 relative drift in the final $$\gamma$$ is real. The expected value was recorded in #1369 (RF cavity phase optimization) and the code has produced something different since, the default `rtol` was hiding it:

Maybe relaxing `atol` is one option, re-recording the reference final $$\gamma$$ is the other. What do you think @cemitch99 ?

cemitch99:
Looking into this.  The $\gamma$ value from #1369 was obtained from numerical optimization (it is not an analytically-determined value).  Which of the two options above is more appropriate depends on whether the new value is larger or smaller than the original value.  Will follow-up asap.

cemitch99:
The (new) computed value is ~1e-6 _less_ relative to the value in the analysis script, so I would not suggest changing the reference value.  The example `run_rfcavity_ref_part_opt.py` appears somewhat delicate.  I propose to experiment with some additional fine-tuning in a separate follow-up PR.

ax3l:
Sounds good! CI otherwise came back green now :)

cemitch99 (approved):

PR #1588, opened 2026-08-03
[WIP] CSR Surrogate Hook

ax3l:
Provide a Python hook to register custom CSR surrogate ML models, e.g., as in [A. L. Edelen et al., in Proc. IPAC'22, WEPOMS013, 2022](https://doi.org/10.18429/JACoW-IPAC2022-WEPOMS013).

Writes the bindings generic enough that we can later on also add transverse kicks.

Attn @lee-edelen @ChristopherMayes @jy-tang

- [Example](https://impactx--1588.org.readthedocs.build/en/1588/usage/examples/pytorch_csr_surrogate/README.html)
- [How-To](https://impactx--1588.org.readthedocs.build/en/1588/usage/howto/python_extend.html#custom-csr-wake-models-ml-surrogates)
- [API docs](https://impactx--1588.org.readthedocs.build/en/1588/usage/python.html#impactx.ImpactX.csr_kick_model)

Credits to Claude to help me with typing.

### DIFF
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -183,11 +183,42 @@ Collective Effects & Overall Simulation Parameters
 
          CSR effects are only calculated for lattice elements that include bending, such as ``Sbend``, ``ExactSbend`` and ``CFbend``.
 
-         CSR effects require the compilation flag ``-DImpactX_FFT=ON``.
+         CSR effects require the compilation flag ``-DImpactX_FFT=ON``, unless a user-defined kick model is set via ``csr_kick_model``.
 
    .. py:property:: csr_bins
 
       The number of bins along the longitudinal direction used for the CSR calculations (default: ``150``).
+
+   .. py:property:: csr_kick_model
+
+      A user-defined CSR kick model (default: ``None``, which selects the built-in analytic model).
+
+      Use this to couple custom CSR models into the tracking loop, e.g., machine-learning surrogate models that were pre-trained on data from high-fidelity CSR solvers, as in
+      `A. L. Edelen et al., in Proc. IPAC'22, WEPOMS013, DOI:10.18429/JACoW-IPAC2022-WEPOMS013 <https://doi.org/10.18429/JACoW-IPAC2022-WEPOMS013>`__
+      (`arXiv:2203.07542 <https://arxiv.org/abs/2203.07542>`__).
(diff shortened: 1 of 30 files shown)


ax3l:
As a follow-up PR, distributions for 2D would be needed:

- https://doi.org/10.18429/JACoW-IPAC2023-WEPA105
- https://proceedings.jacow.org/fls2023/talks/th3d4_talk.pdf
- https://github.com/slaclab/pyDFCSR

Will need to expose charge binning in 2D, e.g., binned for z and x (or whatever direction the beam is bent as the transverse bin).

docs/source/usage/python.rst
```diff
@@ -183,12 +183,43 @@ Collective Effects & Overall Simulation Parameters
...
+         CSR effects require the compilation flag ``-DImpactX_FFT=ON``, unless a user-defined kick model is set via ``csr_kick_model``.
 
    .. py:property:: csr_bins
 
       The number of bins along the longitudinal direction used for the CSR calculations (default: ``150``).
 
+   .. py:property:: csr_kick_model
+
+      A user-defined CSR kick model (default: ``None``, which selects the built-in analytic model).
+
```
ax3l, line 195: I am still a bit on edge if that should not simply be assignable to `sim.csr` and to skip this extra option here.

src/python/impactx/extensions/ImpactX.py
```diff
@@ -0,0 +1,158 @@
...
+            ) from e
+        if value.ndim != 1:
+            raise RuntimeError(
+                "The CSR kick model returned an array that is not 1D" + error_context()
+            )
+
+    if isinstance(value, amr.DeviceVector_real):
+        # fresh copy: the CSRKick construction consumes (moves from) its
+        # inputs, which must not empty a vector owned by the user model
+        vec = amr.DeviceVector_real.from_xp(value.to_xp())
```
ax3l, line 58: Codex:
On GPU, this round-trip routes through the pyAMReX CuPy-backed from_xp copy. That copy is asynchronous, but is_podvector prevents used_cuda_interface from triggering the synchronization below, so ImpactX can consume vec on its AMReX stream before the copy completes. It also makes CuPy mandatory for a documented DeviceVector return. Please use value.to_device() instead; it creates the independent vector through the synchronized AMReX copy without requiring CuPy.

PR #1590, opened 2026-08-04
CI: Actions Cache Hygiene

ax3l:
## Problem

The repository sits at the **10 GB GitHub Actions cache quota** — 10.01 GB across 44 caches when measured — so GitHub LRU-evicts entries and jobs silently lose their ccache.

Unlike WarpX and pyAMReX, this repository has no cache cleanup at all, so nothing ever prunes stale generations. When measured, `development` alone held six generations of `ccache-🐧 HIP-build_hip-*` (109 MB each) and three of `ccache-🔄 Update Stub Files-stubs-*` (~180 MB each). The `codeql-trap-*` caches that `github/codeql-action` stores were also accumulating unmanaged at ~900 MB per generation (1.79 GB when measured).

## Changes

Adopt the cache handling WarpX uses:

- **Add `CleanUpCache`, `PostPR` and `CleanUpCachePostPR`.** The first keeps only the last used cache of each job per branch; the latter two drop all caches of a pull request once it is closed. `CleanUpCache` additionally prunes the CodeQL TRAP database caches, which carry neither the `ccache-` prefix nor a `-git-` separator and so need their own pass.

- **Only run on pushes to `development`.** Developers work from forks, whose pushes never trigger this repository; the only branches pushed here directly are mainline and the pre-commit.ci / dependabot update branches, which open PRs anyway. Running on those pushes as well only stored a second, redundant cache set under `refs/heads/<branch>` next to the `refs/pull/<n>/merge` one. A branch filter also excludes tag pushes, which would cache under the never-restorable `refs/heads/refs/tags/<tag>`. `codeql.yml` already did this, so it is unchanged.

## Notes

- The `workflows:` list in `cleanup-cache.yml` must stay in sync with the workflows that set up a cache — a name that does not match an existing workflow silently disables the cleanup for it. All eight names were verified against the `name:` fields in this branch.
- `🔍 CodeQL` is in the list even though it uses no ccache, because it produces the TRAP caches.
- `🪟 Windows` is in the list for its `build_win_msvc` job. The `build_win_clang` caches use `hashFiles()` keys instead of `-git-<sha>` ones and are intentionally left untouched by the pruning loop.

### DIFF
--- a/.github/workflows/cleanup-cache-postpr.yml
+++ b/.github/workflows/cleanup-cache-postpr.yml
@@ -1,0 +1,50 @@
+# Copyright 2022-2026 The ImpactX Community
+#
+# This file is part of ImpactX.
+#
+# License: BSD-3-Clause-LBNL
+# Authors: Axel Huebl
+
+name: CleanUpCachePostPR
+
+on:
+  workflow_run:
+    workflows: [PostPR]
+    types:
+      - completed
+
+jobs:
+  CleanUpCcacheCachePostPR:
+    name: Clean Up Ccache Cache Post PR
+    runs-on: ubuntu-24.04
+    permissions:
+      actions: write
+      contents: read
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v7
+      - name: Clean up ccache
+        run: |
+          gh extension install actions/gh-actions-cache
+
+          REPO=${{ github.repository }}
+
+          # For debugging cat ${GITHUB_EVENT_PATH} to see the payload.
+
+          pr_head_sha=${{ github.event.workflow_run.head_sha }}
+          pr_number=$(gh pr list --state all --search $pr_head_sha --json number --jq '.[0].number')
+          echo "Post-PR cache cleanup for PR ${pr_number}"
+          BRANCH=refs/pull/${pr_number}/merge
+
+          # Setting this to not fail the workflow while deleting cache keys.
+          set +e
+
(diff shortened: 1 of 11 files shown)


.github/workflows/cleanup-cache-postpr.yml
```diff
@@ -0,0 +1,50 @@
+# Copyright 2022-2026 The ImpactX Community
+#
+# This file is part of ImpactX.
+#
+# License: BSD-3-Clause-LBNL
+# Authors: Axel Huebl
```
ax3l:
```suggestion
# Authors: Axel Huebl, Weiqun Zhang
```

.github/workflows/cleanup-cache.yml
```diff
@@ -0,0 +1,96 @@
+# Copyright 2022-2026 The ImpactX Community
+#
+# This file is part of ImpactX.
+#
+# License: BSD-3-Clause-LBNL
+# Authors: Axel Huebl
```
ax3l:
```suggestion
# Authors: Axel Huebl, Weiqun Zhang
```

.github/workflows/post-pr.yml
```diff
@@ -0,0 +1,24 @@
+# Copyright 2022-2026 The ImpactX Community
+#
+# This file is part of ImpactX.
+#
+# License: BSD-3-Clause-LBNL
+# Authors: Axel Huebl
```
ax3l:
```suggestion
# Authors: Axel Huebl, Weiqun Zhang
```

WeiqunZhang (approved):

PR #1592, opened 2026-08-04
Fix `BeamMonitor`: Sample Interval Leak

### LINKED ISSUE #1583: `period_sample_intervals` inappropriately applied to other `BeamMonitor` elements
I modified the run_fodo.py example so there were two BeamMonitor elements saving to different files which is attached to this issue. The first at the beginning of the lattice has a period_sample_interval set to a large number. The second BeamMonitor element at the end of the lattice has the default sample interval which should be 1. I set sim.periods = 4 to run four iterations of the lattice. After running the script, output from the first monitor has one iteration at the beginning of the lattice but the output from the second monitor also has only one iteration saved.

You can see just from the size of the output files that they have the same number of iterations saved:
```
(impactx-test) [egstern@WL-146057 fodo_monitor_wtf]$ ls -l diags/openPMD
total 3100
-rw-r--r-- 1 egstern egstern 1621720 Aug  2 09:37 monitor1.h5
-rw-r--r-- 1 egstern egstern 1621720 Aug  2 09:37 monitor2.h5
-rw-r--r-- 1 egstern egstern      12 Aug  2 09:37 paraview.pmd
```

But I also have written a little utility that lists the iterations in an openPMD file:

```
(impactx-test) [egstern@WL-146057 fodo_monitor_wtf]$ python ~/impactx-apps/booster-simple/list_iterations.py diags/openPMD/monitor1.h5
file:  diags/openPMD/monitor1.h5
iterations in file:  [1]
```

```
(impactx-test) [egstern@WL-146057 fodo_monitor_wtf]$ python ~/impactx-apps/booster-simple/list_iterations.py diags/openPMD/monitor2.h5
file:  diags/openPMD/monitor2.h5
iterations in file:  [127]
```

[run_fodo.py](https://github.com/user-attachments/files/30633592/run_fodo.py)

ax3l:
The sample interval leaked across all monitors. This fixes it.

Fix #1583

### DIFF
--- a/src/elements/diagnostics/BeamMonitor.cpp
+++ b/src/elements/diagnostics/BeamMonitor.cpp
@@ -28,6 +28,7 @@ namespace io = openPMD;
 
 #include <filesystem>
 #include <fstream>  // for std::ofstream
+#include <stdexcept>
 #include <string>
 #include <utility>
 #include <vector>
@@ -141,6 +142,9 @@ namespace detail {
         m_particles(particles),
         m_beam_moments(beam_moments)
     {
+        if (period_sample_intervals < 1) {
+            throw std::runtime_error("BeamMonitor: period_sample_intervals must be >= 1");
+        }
     }
 
     void BeamMonitor::open ()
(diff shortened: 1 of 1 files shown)


cemitch99 (approved):
Thanks for the quick fix.

Issue #1435, opened 2026-04-27
More Lattice Manipulations: Models


ax3l:
Based on our current selection logic, we need need to change the numerical models after loading a PALS or MAD-X file.

This includes the following common lattice manipulations:
- promote model: from linear to paraxial or exact
- exact models: change integration order (`int_order`)
- exact models: change map steps (`mapsteps`)

We need to do this usually by segment ([start/end element](https://github.com/pals-project/pals-python/issues/26#issuecomment-3406998523)) and element type.

PR #1593, opened 2026-08-06
MAD-X/PALS Load File: `min_model=...`

ax3l:
The MAD-X and PALS readers so far pick the cheapest ImpactX element that represents an imported element: `Drift`, `Quad`, `Sbend`, `CFbend`, `DipEdge(model="linear")`, escalating to `Exact*` only when a feature such as a skew or higher-order multipole leaves no alternative. Users who need chromatic or exact-Hamiltonian dynamics from the start had to post-process the imported lattice element by element.

Add a `min_model=` argument to `KnownElementsList.load_file()`, `from_pals()` and the underlying `read_lattice()`/`lattice()` translators. It raises the lower gate of that choice, using the same "linear"/"paraxial"/"exact" vocabulary as `FilteredElementsList.replace_with_drifts(model=...)`. The default "linear" reproduces the previous output exactly.

Docs: https://impactx--1593.org.readthedocs.build/en/1593/usage/python.html#impactx.elements.KnownElementsList.load_file

Related to #1435

## Details

The floor never lowers a tier: an element that already requires a richer model, such as a skew quadrupole that only exists as `ExactMultipole`, keeps it. Not every family implements every tier (there is no `ChrSbend`, `ChrCFbend`, `ChrSol` or `ExactSol`), so a missing tier is skipped upwards. A paraxial floor on a plain SBEND yields `ExactSbend` rather than falling back to the linear model. Where no implemented model reaches the floor, as for SOLENOID, the most faithful one available is used and a warning is emitted once per lattice.

<details>

New `impactx/element_models.py` holds the one place of truth for the tier ordering: `MODEL_TIERS`, `tier_rank`, `tier_of_class`, `validate_model`, `select_model` plus the `DRIFT_MODEL_CLASSES`/`QUAD_MODEL_CLASSES` tables. `KnownElementsList` now imports its tier helpers from there instead of keeping private copies, so both knobs share one vocabulary and one validation message.

In the MAD-X translator, a single `build(family, builders, **kwargs)` dispatcher plus the `make_drift`, `make_quad` and `make_dipedge` closures replace 14 `Drift`, 3 `Quad` and 5 `DipEdge` construction sites, so applying the floor is one table per family rather than a branch per call site. `make_bend_body_element` keeps its documented priority ladder, with each rung now declaring the tiers ImpactX implements for it.

`DipEdge` selects its fringe model by argument rather than by class, so its 6th-order "nonlinear" map is registered as the exact tier. This matches the choice the synmadx importer already makes in `cnv_dipedge`.

</details>

- [x] Design spec by @ax3l
- [x] Typed with help of Claude Opus 5 xHigh.
- [x] Auto-reviewed with help of Codex GPT 5.6-Sol xHigh.
- [x] Self-reviewed manually.

### DIFF
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -1015,7 +1015,7 @@ This module provides elements and methods for the accelerator lattice.
       starts out with at least the requested level of physical fidelity
       (see :ref:`theory-assumptions`):
 
-      * ``"linear"`` (default) reproduces the historic behavior
+      * ``"linear"`` (default) uses the linear model
       * ``"paraxial"`` requires at least the chromatic ``Chr*`` models
       * ``"exact"`` requires at least the exact-Hamiltonian ``Exact*`` models
 
@@ -1066,7 +1066,7 @@ This module provides elements and methods for the accelerator lattice.
       the same at every floor.
       Their *length*, however, is carried by drifts that the reader adds around
       that kick, and those follow ``min_model`` like any other drift.
-      A finite-length ``RFCAVITY``, for example, becomes
+      A finite-length ``RFCAVITY`` is a thin kick model, for example, becomes
       ``ExactDrift`` + ``ShortRF`` + ``ExactDrift`` at ``min_model="exact"``.
       The same applies to the drifts emitted for ``COLLIMATOR``, ``INSTRUMENT``,
       ``PLACEHOLDER`` and for a ``MONITOR`` with ``L > 0``.
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -1005,9 +1005,64 @@ This module provides elements and methods for the accelerator lattice.
 
       Add a single element to the list.
 
-   .. py:method:: load_file(filename, nslice=1)
+   .. py:method:: load_file(filename, nslice=1, min_model="linear")
 
       Load and append a lattice file from MAD-X (.madx) or PALS (e.g., .pals.yaml) formats.
+
+      The reader picks the *cheapest* ImpactX element that represents each imported
+      element.
+      ``min_model`` raises the lower gate of that choice, so that an imported lattice
+      starts out with at least the requested level of physical fidelity
+      (see :ref:`theory-assumptions`):
+
+      * ``"linear"`` (default) reproduces the historic behavior
+      * ``"paraxial"`` requires at least the chromatic ``Chr*`` models
+      * ``"exact"`` requires at least the exact-Hamiltonian ``Exact*`` models
+
+      A floor never *lowers* a model.
+      An element that already requires a richer model, such as a skew quadrupole
+      that only exists as ``ExactMultipole``, keeps it.
+      Where a tier is not implemented for an element family, the next higher one is
+      used.
+      Where no model reaches the requested tier at all, the most faithful model
+      available is used and a warning is emitted:
+
+      .. list-table::
+         :header-rows: 1
+         :widths: 20 25 25 30
+
+         * - MAD-X element
+           - ``"linear"``
+           - ``"paraxial"``
+           - ``"exact"``
+         * - ``DRIFT``
+           - ``Drift``
+           - ``ChrDrift``
+           - ``ExactDrift``
+         * - ``QUADRUPOLE``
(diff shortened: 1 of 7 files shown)


docs/source/usage/python.rst
```diff
@@ -1005,10 +1005,72 @@ This module provides elements and methods for the accelerator lattice.
...
+      Elements that MAD-X itself defines as thin (``MULTIPOLE``, ``KICKER``,
+      ``RFCAVITY``, ``NLLENS``) and structural elements (``MARKER``, apertures,
+      monitors) have no model tiers, so the kick or map they translate into is
+      the same at every floor.
+      Their *length*, however, is carried by drifts that the reader adds around
+      that kick, and those follow ``min_model`` like any other drift.
+      A finite-length ``RFCAVITY``, for example, becomes
+      ``ExactDrift`` + ``ShortRF`` + ``ExactDrift`` at ``min_model="exact"``.
+      The same applies to the drifts emitted for ``COLLIMATOR``, ``INSTRUMENT``,
+      ``PLACEHOLDER`` and for a ``MONITOR`` with ``L > 0``.
```
ax3l: To check with @cemitch99 if that is a sensible element tier promotion path.

docs/source/usage/python.rst
```diff
@@ -1005,10 +1005,72 @@ This module provides elements and methods for the accelerator lattice.
...
 
       Load and append a lattice file from MAD-X (.madx) or PALS (e.g., .pals.yaml) formats.
 
+      The reader picks the *cheapest* ImpactX element that represents each imported
+      element.
+      ``min_model`` raises the lower gate of that choice, so that an imported lattice
+      starts out with at least the requested level of physical fidelity
+      (see :ref:`theory-assumptions`):
+
+      * ``"linear"`` (default) reproduces the historic behavior
```
```suggestion
      * ``"linear"`` (default) uses the linear model
```

docs/source/usage/python.rst
```diff
@@ -1005,10 +1005,72 @@ This module provides elements and methods for the accelerator lattice.
...
+           - ``ExactMultipole``
+           - ``ExactMultipole``
+
+      Elements that MAD-X itself defines as thin (``MULTIPOLE``, ``KICKER``,
+      ``RFCAVITY``, ``NLLENS``) and structural elements (``MARKER``, apertures,
+      monitors) have no model tiers, so the kick or map they translate into is
+      the same at every floor.
+      Their *length*, however, is carried by drifts that the reader adds around
+      that kick, and those follow ``min_model`` like any other drift.
+      A finite-length ``RFCAVITY``, for example, becomes
```
```suggestion
      A finite-length ``RFCAVITY`` is a thin kick model, for example, becomes
```

cemitch99 (commented):
This is great.  I noticed that `Sol` is used for a solenoid at all 3 tiers.  However, we _do_ have the equivalent of a `ChrSol`: by setting `ez=0` in the element `ChrAcc`.  An `ExactAcc` model will be implemented at some point.

cemitch99 (approved):
Will propose a follow-up relating Acc -> Solenoid models once `ExactAcc` element is ready.

docs/source/usage/python.rst
```diff
@@ -1005,10 +1005,72 @@ This module provides elements and methods for the accelerator lattice.
...
+      ``min_model`` raises the lower gate of that choice, so that an imported lattice
+      starts out with at least the requested level of physical fidelity
+      (see :ref:`theory-assumptions`):
+
+      * ``"linear"`` (default) uses the linear model
+      * ``"paraxial"`` requires at least the chromatic ``Chr*`` models
+      * ``"exact"`` requires at least the exact-Hamiltonian ``Exact*`` models
+
+      A floor never *lowers* a model.
+      An element that already requires a richer model, such as a skew quadrupole
```
cemitch99: Oh, this statement is false.  A skew quad can be represented as a linear Quad, paraxial ChrQuad, or ExactQuad with a 45 deg rotation (`tilt`).

```suggestion
      An element that already requires a richer model, such as a thick octupole
```

ax3l: thx! fixed

ax3l: thx! fixed

PR #1599, opened 2026-08-07
Python: Bulk `set()` and `select(has=...)`

ax3l:
Tuning numeric knobs on an imported lattice — `nslice`, `int_order`, `mapsteps` — meant looping over a selection and assigning attributes one at a time. That breaks on any mixed selection, because elements only carry the properties they support.

This is the last PR of the series that started with #1595, #1596, #1597 and #1598.

- [ ] cleaner implementable once #1381 is merged

## Motivation

The ImpactX defaults are measurably too loose for ring tracking. On the Fermilab Booster (1 turn, 10k particles, no space charge), varying only the integrator settings against a converged `(int_order=6, mapsteps=40)` reference:

| setting | rel. err σ_x | rel. err σ_y | rel. err ε_x |
|---|---|---|---|
| `(6, 6)` — what synmadx hardcodes | 4.8e-08 | 4.2e-08 | 1.0e-08 |
| `(2, 10)` — ImpactX defaults | **3.1e-03** | **2.4e-03** | 5.4e-04 |

So ~0.3 % error in beam size after a single turn. The Synergia-derived `syn2_to_impactx` converter hardcodes per-element-type overrides for exactly this reason.

We deliberately did **not** fold this into the importer: `min_model=` (#1593) selects the element *class*, and silently making it ~2.5x slower would be surprising. Integrator accuracy belongs in the post-load manipulation API, where it applies however the lattice was built.

## What was blocking it

```python
for el in lat.select():
    el.nslice = 4
# AttributeError: property of 'DipEdge' object has no setter
```

and capability was not selectable — `select(kind=r"Exact.*")` over-matches `ExactDrift`, which has no `int_order`, so the loop dies there too. Selecting "everything with an integrator" meant hardcoding three class names.

## API

```python
sim.lattice.set(nslice=8, skip=True)
sim.lattice.select(has="int_order").set(int_order=6, mapsteps=6)
```

**`set(**kwargs)`** on both `KnownElementsList` and `FilteredElementsList`. Generic over property names — `ds`, `k`, `rotation`, `aperture_x` all work — so future knobs need no new API. Returns the number of elements written.

By default it **raises** `AttributeError`, naming the property and offending element kinds, if any selected element cannot take a value. `skip=True` sets only where applicable. Raising is the default so that a mistyped name (`set(nslize=8)`) cannot silently do nothing.

Assignment is **all-or-nothing**: values are checked first, then capability is scanned across the whole selection, and only then is anything written. A rejected value or an unsettable element leaves every element untouched rather than half-applying and raising midway.

**`select(has=...)`** matches elements that can *set* a property, not merely report one. This distinction matters: `nslice` is readable on all 38 element classes but settable only on thick ones, so a readable-based filter would match everything and be useless. Matching is exact, without regex, since these are API names rather than user labels. `has=` joins the existing OR group with `kind` and `name`; AND is expressed by chaining, as before.

## Notes for review

`set` is the only mutating method that does **not** invalidate live selections, because it modifies elements in place instead of rebuilding the lattice. The `FilteredElementsList` docs previously claimed *all* mutating operations invalidate references; that has been corrected and `set` called out as the exception.

The module-level function is named `set_properties` and bound as `set`, because this module calls the builtin `set(...)` in `delete`, `replace_each`, `replace_with_drifts` and `get_kinds`.

`_SET_VALIDATORS` mirrors the C++ checks added in #1596/#1598 so a bad value is rejected before any write. The rules are duplicated in Python deliberately — delegating entirely to the setters would mean a mid-loop failure leaves earlier elements already modified.

## Testing

25 new tests. Coverage includes the settable-not-readable semantics of `has=`, OR/AND composition, raise-by-default leaving the lattice unchanged, `skip=True`, typo detection, each invalid value rejected with nothing written, and the contrast that `set()` keeps selections valid while `delete()` invalidates them. Also asserts the element-level validation from #1596/#1598 so a regression in the C++ layer surfaces here.

`pytest.ImpactX`: 355 passed.

- [x] Design spec by @ax3l
- [x] Typed with help of Claude Opus 5 xHigh.
- [ ] Auto-reviewed with help of Codex GPT 5.6-Sol xHigh.
- [x] Self-reviewed manually.

### DIFF
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -1085,17 +1085,19 @@ This module provides elements and methods for the accelerator lattice.
       Assign element properties in bulk, on every element of the lattice.
 
       Works for any settable property (``nslice``, ``int_order``, ``mapsteps``,
-      ``ds``, ``k``, ``rotation``, ``aperture_x``, ...), so no new API is needed
-      as elements gain knobs.
+      ``ds``, ``k``, ``rotation``, ``aperture_x``, ...).
 
       By default this **raises** ``AttributeError`` if any element cannot take one
       of the given properties, naming the property and the offending element
       kinds. Pass ``skip=True`` to set only where applicable, or narrow the
       selection first with ``select(has=...)``.
 
-      Assignment is all-or-nothing: values are checked, then capability is checked
-      across the whole lattice, and only then is anything written. A rejected
-      value or an unsettable element leaves every element untouched.
+      Assignment is all-or-nothing. Values are checked, then capability is checked
+      across the whole lattice, and only then is anything written. Should a
+      property setter still reject a value at that point, every write already made
+      is undone in reverse order. A rejected value or an unsettable element
+      therefore leaves every element exactly as it was, whichever property or
+      element the rejection came from.
 
       .. note::
 
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -1005,6 +1005,22 @@ This module provides elements and methods for the accelerator lattice.
 
       Add a single element to the list.
 
+   .. note::
+
+      ``append`` and ``extend`` store a **copy** of the element. Indexing the
+      lattice afterwards returns that copy, not the object you passed in, and the
+      two are independent from then on: ``lattice[0].ds = 7.0`` does not change
+      the original, nor the reverse.
+
+      This matters for :py:class:`~impactx.elements.Programmable`, whose
+      callbacks usually close over the object you constructed while tracking
+      drives the lattice's copy. Changing ``ds`` or ``nslice`` on one side only
+      is enough to make the two disagree. Set such values before appending, or
+      re-fetch the element from the lattice and change it there.
+
+      Making ``sim.lattice`` hold references instead of copies is tracked in
+      `PR #1381 <https://github.com/BLAST-ImpactX/impactx/pull/1381>`__.
+
    .. py:method:: load_file(filename, nslice=1)
 
       Load and append a lattice file from MAD-X (.madx) or PALS (e.g., .pals.yaml) formats.
@@ -1086,6 +1102,9 @@ This module provides elements and methods for the accelerator lattice.
 
       Works for any settable property (``nslice``, ``int_order``, ``mapsteps``,
       ``ds``, ``k``, ``rotation``, ``aperture_x``, ...).
+
+      This writes to the elements held by the lattice, which are copies of the
+      ones that were appended; see the note on :py:meth:`~append`.
 
       By default this **raises** ``AttributeError`` if any element cannot take one
       of the given properties, naming the property and the offending element
--- a/docs/source/usage/python.rst
+++ b/docs/source/usage/python.rst
@@ -1029,17 +1029,25 @@ This module provides elements and methods for the accelerator lattice.
       :param pals_line: PALS Python Line with beamline elements
       :param nslice: number of slices used for the application of collective effects
 
-   .. py:method:: select(kind=None, name=None)
+   .. py:method:: select(kind=None, name=None, has=None)
 
-      Filter elements by type and/or name.
-      If both are provided, OR-based logic is applied.
+      Filter elements by type, name and/or settable property.
+      If more than one is provided, OR-based logic is applied.
 
       Returns references to original elements, allowing modification and chaining.
       Chained ``.select(...).select(...)`` selections are AND-filtered.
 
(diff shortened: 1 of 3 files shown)


docs/source/usage/python.rst
```diff
@@ -1067,6 +1080,50 @@ This module provides elements and methods for the accelerator lattice.
          # replace all Quads with drift equivalents
          lattice.select(kind=r".*Quad").replace_with_drifts()
 
+   .. py:method:: set(*, skip=False, **kwargs)
+
+      Assign element properties in bulk, on every element of the lattice.
+
+      Works for any settable property (``nslice``, ``int_order``, ``mapsteps``,
+      ``ds``, ``k``, ``rotation``, ``aperture_x``, ...), so no new API is needed
+      as elements gain knobs.
```
```suggestion
      ``ds``, ``k``, ``rotation``, ``aperture_x``, ...).
```

tests/python/test_lattice_select.py
```diff
@@ -1169,3 +1169,219 @@ def test_replace_each_with_thin_element_present():
...
+    ]
+
+
+def test_select_has_integrator_does_not_overmatch_exact():
+    """`Exact.*` over-matches; has="int_order" selects only real integrators."""
+    lattice = _mixed_lattice()
+    assert "ExactDrift" in _kinds(lattice.select(kind=r"Exact.*"))
+    assert _kinds(lattice.select(has="int_order")) == ["ExactQuad"]
+
+
```
ax3l: Not super useful / covered elsewhere
```suggestion
```

src/python/impactx/extensions/KnownElementsList.py
```diff
@@ -498,16 +589,44 @@ def _matches_string(text: str, string_pattern: str) -> bool:
...
+
+    Args:
+        element: The element to check
+        attr: Property name, e.g. "nslice", "int_order", "mapsteps"
+
+    Returns:
+        bool: True if the property exists and can be assigned
+    """
+    descriptor = getattr(type(element), attr, None)
+    return descriptor is not None and getattr(descriptor, "fset", None) is not None
```
ax3l, line 626: Codex:

[P1] Please exclude Programmable.ds/nslice from bulk lattice mutation (or synchronize their state). _is_settable treats both as safe because they have setters, but KnownElementsList.append stores a copy, while Programmable callbacks normally close over the original Python element and read its ds/nslice. Tracking uses the lattice copy value of nslice to choose the callback count. After sim.lattice.set(nslice=8, skip=True), the copy may cause eight calls while each callback still advances using the original slice length, silently producing the wrong total transport. Please cover this case with a regression test.

src/python/impactx/extensions/KnownElementsList.py
```diff
@@ -479,6 +501,75 @@ def __str__(self):
...
+                f"select(has=...), or pass skip=True to set only where applicable."
+            )
+
+    # 3) apply
+    changed = 0
+    for element in element_list:
+        touched = False
+        for attr, value in kwargs.items():
+            if _is_settable(element, attr):
+                setattr(element, attr, value)
```
ax3l: Codex:

[P2] Please make the generic setter genuinely transactional. The preflight validators cover only nslice, int_order, and mapsteps, but other setters can reject values. For example, lat.set(psi=0.2, R=-1.0) raises ValueError from DipEdge.R after psi has already changed from 0.1 to 0.2 (confirmed against this build). Invalid enum values and type-conversion failures can do the same. This contradicts the documented all-or-nothing guarantee; preflight every actual assignment, roll back on failure, or narrow the guarantee/API.

ax3l: I was afraid of that when deciding the design docs.
I think the generic way to make it stable is to do two passes, one for validation and one for setting. For book-keeping like this, the cost of 2N is totally fine.

ax3l: Codex:

On reflection, clone-based preflight is likely over-designed if the intended guarantee is only that no writes occur until every selected element has the requested setters. The existing capability scan already provides that behavior and handles mixed selections and property-name typos.

It does not provide the stronger documented guarantee for all rejected values: DipEdge has an R setter, so capability validation passes, and R=-1 can still fail after an earlier property such as psi was written.

The simplest resolution is therefore to narrow the contract: attribute and capability failures are all-or-nothing; the known numeric knobs are prevalidated; other setter-time errors may leave preceding assignments applied. With the documentation adjusted to say that, no clone machinery is needed and my P2 concern becomes a contract/documentation correction rather than a request for full transactional implementation.

PR #1602, opened 2026-08-10
Python: Advance Reference Particle in `Element.transfer_map`

ax3l:
`Element.transfer_map(ref)` handed the caller's reference particle straight to `transport_map()`. That is not the state our elements expect there, returning invalid maps for some elements (e.g., soft edge elements).

`diagnostics::walk_lattice` (and the envelope tracker) advance the reference particle by one slice *first*, then evaluate the map at the post-advance state -- the convention documented in `src/diagnostics/LinearMap.H`. Two groups of elements depend on it:

- `ShortRF::transport_map()` and `ChrUniformAcc::transport_map()` reconstruct the *initial* `pt` from `refpart.pt`, i.e. they assume it holds the final value. From Python they were evaluated at the wrong reference momentum.
- `RFCavity`, `SoftQuadrupole` and `SoftSolenoid` integrate their linear map *during* the reference push and cache it in `m_map`. Without a preceding push, `transfer_map()` returned the all-zero matrix that member is default-constructed to:

```python
>>> elements.SoftSolenoid(
    ds=0.3,
    bscale=1.0,
    # ...
    cos_coefficients=[2.0],
    sin_coefficients=[0.0]
).transfer_map(ref).to_numpy()

array([[0., 0., 0., 0., 0., 0.],
       ...
       [0., 0., 0., 0., 0., 0.]])
```

This advances a private copy of the reference particle, with `sedge` set to the element entrance, as `walk_lattice` does, so the soft-edge elements integrate over the right window, and evaluates the map there. The caller's reference particle is left untouched, matching what `KnownElementsList.transfer_map` already promises.

### Tests

`tests/python/test_lattice_optics.py` gains a parametrized test that an element's own `transfer_map` equals the map of a one-element lattice containing it (they are both the single-slice map), covering `Drift` as a control, the two post-advance-`pt` elements and the three map-integrating elements; plus a check that the caller's reference particle is not advanced.


tests/python/test_lattice_optics.py
```diff
@@ -70,6 +70,72 @@ def test_lattice_linear_map():
     assert np.allclose(R.to_numpy(), R_expected, rtol=rtol, atol=atol)
 
 
+def _single_element_lattices():
+    """One-slice elements whose linear map depends on the reference push."""
+    return [
+        # plain control: map is independent of the reference push
+        elements.Drift(ds=0.3),
+        # map is evaluated at the post-advance reference momentum
+        elements.ShortRF(V=5.0e6, freq=1.3e9, phase=-90.0),
```
ax3l: Codex:

Please use a phase with a nonzero synchronous energy kick here. `ShortRF` advances `pt` by `-V*cos(phi)`, so at `-90°` the reference momentum is unchanged (to floating-point precision) and this case would also pass with the old pre-advance implementation. Setting the phase to `0°` makes this specifically regression-test the reference push.

```suggestion
        elements.ShortRF(V=5.0e6, freq=1.3e9, phase=0.0),
```

ax3l:
@cemitch99 this one is ready for review now :)

It's a fix for the Python API only when querying the `transfer_map` of individual elements, e.g., for debugging/diagnostics. Lattice-level transfer maps, envelope tracking, and particle tracking already used the correct reference-particle advancement convention and are unchanged.

cemitch99 (approved):

Issue #1322, opened 2026-02-24
Spin capability element list


cemitch99:
This Issue tracks the addition of spin-capable tracking to existing element types, and will be updated regularly.

**Linear+paraxial elements:**

- [x] CFbend
- [ ] ConstF
- [x] Drift
- [x] ChrDrift
- [x] Quad
- [x] ChrQuad
- [ ] ChrPlasmaLens
- [ ] ChrAcc
- [ ] Buncher
- [x] Sbend
- [x] Sol

**Soft edge elements:**

- [ ] RFCavity #1264 
- [ ] SoftQuadrupole
- [ ] SoftSolenoid

**Exact nonlinear elements:**

- [ ] ExactCFbend
- [x] ExactDrift
- [ ] ExactMultipole
- [x] ExactQuad #1323
- [ ] ExactSbend

**Thin elements:**

- [ ] DipEdge
- [ ] LinearMap
- [ ] Multipole #1154
- [ ] Kicker
- [ ] NonlinearLens
- [ ] QuadEdge
- [ ] ShortRF
- [ ] ThinDipole
- [ ] TaperedPL #1154

carlandreaslindstrom:
Hi @ax3l and @cemitch99: are you able to also add `Aperture` to this list? It currently does not work. Presumably it should be a quick fix, given that nothing actually happens.

ax3l:
@carlandreaslindstrom Oh ha, yes they are easy! Thanks for raising, here they are: https://github.com/BLAST-ImpactX/impactx/pull/1493

PR #1606, opened 2026-08-10
Spin push for thin kick elements

cemitch99:
This PR adds the spin push through those `Thin` elements _not_ associated with fringe fields.  The logic for all of these elements uses the same pattern, and only the evaluation of the magnetic or electric fields should differ.

The implementation here is $s$-symmetric.  A reversibility test will be included.

- [x] Kicker
- [x] ShortRF
- [x] NonlinearLens
- [x] ThinDipole
- [x] Buncher
- [x] finalize reversibility tests
- [x] change all but `ShortRF` to midpush scheme? https://github.com/cemitch99/impactx/pull/4

This closes Issue #1322

### DIFF
--- a/src/elements/Buncher.H
+++ b/src/elements/Buncher.H
@@ -181,6 +181,7 @@ namespace impactx::elements
         {
             using namespace amrex::literals; // for _rt and _prt
             using amrex::Math::powi;
+            using namespace std; // for cmath(float)
 
             // reference particle relativistic factors
             amrex::ParticleReal const gamma_ref = refpart.gamma();
@@ -236,7 +237,6 @@ namespace impactx::elements
 
             // push the spin vector using the generator just determined (half spin-kick)
             rotate_spin(lambdax, lambday, lambdaz, sx, sy, sz);
-
         }
 
 
--- a/src/elements/Buncher.H
+++ b/src/elements/Buncher.H
@@ -190,24 +190,42 @@ namespace impactx::elements
             T_Real lambday = 0_prt;
             T_Real lambdaz = 0_prt;
 
-            // Integrated magnetic field normalized by q/mc (half-step):
-            T_Real const Bx = 0.5_prt * m_kV_r2bg2 * y * m_beta * gamma_ref;
-            T_Real const By = -0.5_prt * m_kV_r2bg2 * x * m_beta * gamma_ref;
+            // Integrated magnetic field normalized by q/mc:
+            T_Real const Bx = m_kV_r2bg2 * y * m_beta * gamma_ref;
+            T_Real const By = -m_kV_r2bg2 * x * m_beta * gamma_ref;
             T_Real const Bz = 0_prt;
 
-            // Integrated electric field normalized by q/mc^2 (half-step):
+            // Integrated electric field normalized by q/mc^2:
             T_Real const Ex = 0.0_prt;
             T_Real const Ey = 0.0_prt;
-            T_Real const Ez = -0.5_prt * m_neg_kV * t * m_beta * gamma_ref;
+            T_Real const Ez = -m_neg_kV * t * m_beta * gamma_ref;
 
-            // Quantities required to evaluate the full Thomas-BMT precession vector (before phase space kick).
-            T_Real Pnorm = sqrt(1_prt - 2_prt * pt / m_beta + pt*pt);
-            T_Real iPnorm = 1_prt/Pnorm;
-            T_Real gamma = gamma_ref * (1_prt - pt*m_beta);
-            T_Real Ps = sqrt(Pnorm*Pnorm - px*px - py*py);
-            T_Real ux = px * iPnorm;
-            T_Real uy = py * iPnorm;
-            T_Real uz = Ps * iPnorm;
+            // The Thomas-BMT generator is evaluated at the momentum halfway
+            // through the kick, so that reversing the element (negating V)
+            // produces the exact inverse rotation. The buncher also changes the
+            // energy deviation, so pt is averaged along with the transverse
+            // momenta.
+            T_Real const px_in = px;
+            T_Real const py_in = py;
+            T_Real const pt_in = pt;
+
+            // phase space push
+            (*this)(x, y, t, px, py, pt, idcpu, refpart);
+
+            // momenta at the kick midpoint
+            T_Real const px_mid = 0.5_prt * (px_in + px);
+            T_Real const py_mid = 0.5_prt * (py_in + py);
+            T_Real const pt_mid = 0.5_prt * (pt_in + pt);
+
+            // Quantities required to evaluate the full Thomas-BMT precession
+            // vector, at the midpoint momentum.
+            T_Real const Pnorm = sqrt(1_prt - 2_prt * pt_mid / m_beta + pt_mid*pt_mid);
+            T_Real const iPnorm = 1_prt/Pnorm;
+            T_Real const gamma = gamma_ref * (1_prt - pt_mid*m_beta);
+            T_Real const Ps = sqrt(Pnorm*Pnorm - px_mid*px_mid - py_mid*py_mid);
+            T_Real const ux = px_mid * iPnorm;
+            T_Real const uy = py_mid * iPnorm;
+            T_Real const uz = Ps * iPnorm;
 
             // Zero curvature in a buncher
             amrex::ParticleReal const h = 0.0_prt;
@@ -215,25 +233,7 @@ namespace impactx::elements
             // Evaluation of the full Thomas-BMT precession vector.
             tbmt_precession_vector(x,ux,uy,uz,gamma,h,gyro_anomaly,Bx,By,Bz,Ex,Ey,Ez,lambdax,lambday,lambdaz);
 
-            // push the spin vector using the generator just determined (half spin-kick)
-            rotate_spin(lambdax, lambday, lambdaz, sx, sy, sz);
-
-            // phase space push
-            (*this)(x, y, t, px, py, pt, idcpu, refpart);
-
-            // Quantities required to evaluate the full Thomas-BMT precession vector (after phase space kick).
-            Pnorm = sqrt(1_prt - 2_prt * pt / m_beta + pt*pt);
-            iPnorm = 1_prt/Pnorm;
-            gamma = gamma_ref * (1_prt - pt*m_beta);
-            Ps = sqrt(Pnorm*Pnorm - px*px - py*py);
-            ux = px * iPnorm;
-            uy = py * iPnorm;
-            uz = Ps * iPnorm;
-
-            // Evaluation of the full Thomas-BMT precession vector.
-            tbmt_precession_vector(x,ux,uy,uz,gamma,h,gyro_anomaly,Bx,By,Bz,Ex,Ey,Ez,lambdax,lambday,lambdaz);
-
-            // push the spin vector using the generator just determined (half spin-kick)
+            // push the spin vector using the generator just determined
             rotate_spin(lambdax, lambday, lambdaz, sx, sy, sz);
         }
 
--- a/src/elements/Buncher.H
+++ b/src/elements/Buncher.H
@@ -15,6 +15,7 @@
 #include "mixin/beamoptic.H"
 #include "mixin/thin.H"
 #include "mixin/lineartransport.H"
+#include "mixin/spintransport.H"
 #include "mixin/named.H"
 #include "mixin/nofinalize.H"
 
@@ -34,6 +35,7 @@ namespace impactx::elements
       public mixin::LinearTransport<Buncher>,
       public mixin::Thin,
       public mixin::Alignment,
+      public mixin::SpinTransport,
       public mixin::NoFinalize
       // At least on Intel AVX512, there is a small overhead to vectorize this element, see
       // https://github.com/BLAST-ImpactX/impactx/pull/1002
@@ -90,6 +92,9 @@ namespace impactx::elements
 
             m_neg_kV = -m_k * m_V;
             m_kV_r2bg2 = -m_neg_kV / (2.0_prt * betgam2);
+
+            m_beta = refpart.beta();
+            m_ibeta = 1.0_prt / m_beta;
         }
 
         /** This is a buncher functor, so that a variable of this type can be used like a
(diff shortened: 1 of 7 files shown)


cemitch99:
The error is due to a warning associated with a possible loss of precision (double > float conversion) when running w/ SIMD.

src/elements/Buncher.H
```diff
@@ -143,6 +148,98 @@ namespace impactx::elements
...
+            T_Real & AMREX_RESTRICT pt,
+            T_Real & AMREX_RESTRICT sx,
+            T_Real & AMREX_RESTRICT sy,
+            T_Real & AMREX_RESTRICT sz,
+            [[maybe_unused]] T_IdCpu const & AMREX_RESTRICT idcpu,
+            [[maybe_unused]] RefPart const & AMREX_RESTRICT refpart
+        ) const
+        {
+            using namespace amrex::literals; // for _rt and _prt
+            using amrex::Math::powi;
```
ax3l: CI fails due to an actual bug, this is the fix:
```suggestion
            using amrex::Math::powi;
            using namespace std; // for cmath(float)
```

Otherwise it would use the C math APIs.

After we had this bug pattern a few times, I added a blocking CI test for it :)

src/elements/Buncher.H
```diff
@@ -143,6 +148,98 @@ namespace impactx::elements
...
+            ux = px * iPnorm;
+            uy = py * iPnorm;
+            uz = Ps * iPnorm;
+
+            // Evaluation of the full Thomas-BMT precession vector.
+            tbmt_precession_vector(x,ux,uy,uz,gamma,h,gyro_anomaly,Bx,By,Bz,Ex,Ey,Ez,lambdax,lambday,lambdaz);
+
+            // push the spin vector using the generator just determined (half spin-kick)
+            rotate_spin(lambdax, lambday, lambdaz, sx, sy, sz);
+
```
ax3l: Small style nit:
```suggestion
```

src/elements/Kicker.H
```diff
@@ -173,6 +178,95 @@ namespace impactx::elements
...
+            T_Real const & AMREX_RESTRICT pt,
+            T_Real & AMREX_RESTRICT sx,
+            T_Real & AMREX_RESTRICT sy,
+            T_Real & AMREX_RESTRICT sz,
+            T_IdCpu const & AMREX_RESTRICT idcpu,
+            RefPart const & AMREX_RESTRICT refpart
+        ) const
+        {
+            using namespace amrex::literals; // for _rt and _prt
+            using amrex::Math::powi;
```
ax3l: Same bug:
```suggestion
            using amrex::Math::powi;
            using namespace std; // for cmath(float)
```

src/elements/Kicker.H
```diff
@@ -173,6 +178,95 @@ namespace impactx::elements
...
+            ux = px * iPnorm;
+            uy = py * iPnorm;
+            uz = Ps * iPnorm;
+
+            // Evaluation of the full Thomas-BMT precession vector.
+            tbmt_precession_vector(x,ux,uy,uz,gamma,h,gyro_anomaly,Bx,By,Bz,Ex,Ey,Ez,lambdax,lambday,lambdaz);
+
+            // push the spin vector using the generator just determined (half spin-kick)
+            rotate_spin(lambdax, lambday, lambdaz, sx, sy, sz);
+
```
ax3l: Style nit:
```suggestion
```

src/elements/NonlinearLens.H
```diff
@@ -184,6 +189,117 @@ namespace impactx::elements
...
+            [[maybe_unused]] T_Real const & AMREX_RESTRICT pt,
+            T_Real & AMREX_RESTRICT sx,
+            T_Real & AMREX_RESTRICT sy,
+            T_Real & AMREX_RESTRICT sz,
+            [[maybe_unused]] T_IdCpu const & AMREX_RESTRICT idcpu,
+            [[maybe_unused]] RefPart const & AMREX_RESTRICT refpart
+        ) const
+        {
+            using namespace amrex::literals; // for _rt and _prt
+            using amrex::Math::powi;
```
ax3l:
```suggestion
            using amrex::Math::powi;
            using namespace std; // for cmath(float)
```

src/elements/NonlinearLens.H
```diff
@@ -184,6 +189,117 @@ namespace impactx::elements
...
+            ux = px * iPnorm;
+            uy = py * iPnorm;
+            uz = Ps * iPnorm;
+
+            // Evaluation of the full Thomas-BMT precession vector.
+            tbmt_precession_vector(x,ux,uy,uz,gamma,h,gyro_anomaly,Bx,By,Bz,Ex,Ey,Ez,lambdax,lambday,lambdaz);
+
+            // push the spin vector using the generator just determined (half spin-kick)
+            rotate_spin(lambdax, lambday, lambdaz, sx, sy, sz);
+
```
ax3l:
```suggestion
```

src/elements/ShortRF.H
```diff
@@ -221,6 +225,105 @@ namespace impactx::elements
...
+            T_Real & AMREX_RESTRICT pt,
+            T_Real & AMREX_RESTRICT sx,
+            T_Real & AMREX_RESTRICT sy,
+            T_Real & AMREX_RESTRICT sz,
+            [[maybe_unused]] T_IdCpu const & AMREX_RESTRICT idcpu,
+            [[maybe_unused]] RefPart const & AMREX_RESTRICT refpart
+        ) const
+        {
+            using namespace amrex::literals; // for _rt and _prt
+            using amrex::Math::powi;
```
ax3l:
```suggestion
            using amrex::Math::powi;
            using namespace std; // for cmath(float)
```

src/elements/ShortRF.H
```diff
@@ -221,6 +225,105 @@ namespace impactx::elements
...
+            ux = px * iPnorm;
+            uy = py * iPnorm;
+            uz = Ps * iPnorm;
+
+            // Evaluation of the full Thomas-BMT precession vector.
+            tbmt_precession_vector(x,ux,uy,uz,gamma,h,gyro_anomaly,Bx,By,Bz,Ex,Ey,Ez,lambdax,lambday,lambdaz);
+
+            // push the spin vector using the generator just determined (half spin-kick)
+            rotate_spin(lambdax, lambday, lambdaz, sx, sy, sz);
+
```
ax3l:
```suggestion
```

src/elements/ThinDipole.H
```diff
@@ -179,6 +182,101 @@ namespace impactx::elements
...
+            T_Real const & AMREX_RESTRICT pt,
+            T_Real & AMREX_RESTRICT sx,
+            T_Real & AMREX_RESTRICT sy,
+            T_Real & AMREX_RESTRICT sz,
+            T_IdCpu const & AMREX_RESTRICT idcpu,
+            RefPart const & AMREX_RESTRICT refpart
+        ) const
+        {
+            using namespace amrex::literals; // for _rt and _prt
+            using amrex::Math::powi;
```
ax3l:
```suggestion
            using amrex::Math::powi;
            using namespace std; // for cmath(float)
```

src/elements/ThinDipole.H
```diff
@@ -179,6 +182,101 @@ namespace impactx::elements
...
+
+            // Evaluation of the full Thomas-BMT precession vector.
+            tbmt_precession_vector(x,ux,uy,uz,gamma,h,gyro_anomaly,Bx,By,Bz,Ex,Ey,Ez,lambdax,lambday,lambdaz);
+            lambdax *= m_ds;
+            lambday *= m_ds;
+            lambdaz *= m_ds;
+
+            // push the spin vector using the generator just determined (half spin-kick)
+            rotate_spin(lambdax, lambday, lambdaz, sx, sy, sz);
+
```
ax3l:
```suggestion
```

ax3l:
Thanks for the PR! :tada:

I fixed the bug that CI caught.
Will do a detailed review a bit later :)

src/elements/ThinDipole.H
```diff
@@ -179,6 +182,101 @@ namespace impactx::elements
...
+            Ps = sqrt(Pnorm*Pnorm - px*px - py*py);
+            ux = px * iPnorm;
+            uy = py * iPnorm;
+            uz = Ps * iPnorm;
+
+            // Evaluation of the full Thomas-BMT precession vector.
+            tbmt_precession_vector(x,ux,uy,uz,gamma,h,gyro_anomaly,Bx,By,Bz,Ex,Ey,Ez,lambdax,lambday,lambdaz);
+            lambdax *= m_ds;
+            lambday *= m_ds;
+            lambdaz *= m_ds;
```
ax3l: Claude:

The curvature `h` is passed at half strength (`0.5 * m_kx`), together with a half-strength `By`, and the resulting generator is then scaled by the *full* `m_ds`. That is self-consistent for everything that enters the T-BMT vector **linearly** — the `+ h` frame-rotation term and the `B` terms both come out at the correct half-step value — but `h` also enters **multiplicatively** through the geometric factor in `mixin/spintransport.H`:

```cpp
T_Real const leading_factor = -(1_prt + h * x) / beta_gamma_z;
```

That `(1 + h*x)` is the Frenet-Serret Jacobian, not something that scales with the step length, so it needs the true design curvature `1/rc`. As written it is evaluated with `0.5/rc`, i.e. the correction is half of what it should be.

Reversibility cannot catch this: the half-kick/kick/half-kick construction is exactly self-inverting for *any* value of `h`, so `test_ThinDipole[spin]` stays green either way.

Magnitude: the field part of the generator is off by the relative factor `0.5 * x/rc`, independent of the bend angle. (Correcting the absolute figures I first posted here — I had read `theta` as radians, but the constructor takes degrees.) For the reversibility-test parameters (`theta=0.45` deg = 7.85e-3 rad, `rc=1 m`, 1 GeV electrons, `sigma_x ~ 3.2e-5 m`) that is a relative error of `~1.6e-5` on the field term, i.e. `~4e-7 rad` of absolute spin angle at 1 sigma, growing to `~5e-4` relative / `~1.3e-5 rad` absolute at a 1 mm offset. Small for this configuration, but it is a systematic error that scales with `x/rc`, not a truncation error that vanishes with finer slicing.

The suggestion below moves the factor of 1/2 out of `h`/`By` and onto the arc length, which is mathematically identical except for the `(1 + h*x)` factor and matches the convention already used in `ExactCFbend.H` (`h = 1_prt/m_rc`, unscaled fields, `lambda *= tau`). It also drops the `// Integrated curvature = bend angle in rad (half-step)` comment, which does not describe `0.5*m_kx` (a curvature in 1/m, not an angle).

```suggestion
            // Magnetic field normalized by q/mc [1/m]:
            T_Real const Bx = 0_prt;
            T_Real const By = m_kx * m_beta * gamma_ref;
            T_Real const Bz = 0_prt;

            // Electric field normalized by q/mc^2 [1/m]:
            T_Real const Ex = 0.0_prt;
            T_Real const Ey = 0.0_prt;
            T_Real const Ez = 0.0_prt;

            // Quantities required to evaluate the full Thomas-BMT precession vector (before phase space kick).
            T_Real const Pnorm = sqrt(1_prt - 2_prt * pt * m_ibeta + pt*pt);
            T_Real const iPnorm = 1_prt/Pnorm;
            T_Real const gamma = gamma_ref * (1_prt - pt*m_beta);
            T_Real Ps = sqrt(Pnorm*Pnorm - px*px - py*py);
            T_Real ux = px * iPnorm;
            T_Real uy = py * iPnorm;
            T_Real uz = Ps * iPnorm;

            // Design-orbit curvature [1/m]
            amrex::ParticleReal const h = m_kx;

            // Generator for half of the effective (equivalent) arc length
            amrex::ParticleReal const half_ds = 0.5_prt * m_ds;

            // Evaluation of the full Thomas-BMT precession vector.
            tbmt_precession_vector(x,ux,uy,uz,gamma,h,gyro_anomaly,Bx,By,Bz,Ex,Ey,Ez,lambdax,lambday,lambdaz);
            lambdax *= half_ds;
            lambday *= half_ds;
            lambdaz *= half_ds;

            // push the spin vector using the generator just determined (half spin-kick)
            rotate_spin(lambdax, lambday, lambdaz, sx, sy, sz);

            // phase space push
            (*this)(x, y, t, px, py, pt, idcpu, refpart);

            // Quantities required to evaluate the full Thomas-BMT precession vector (after phase space kick).
            Ps = sqrt(Pnorm*Pnorm - px*px - py*py);
            ux = px * iPnorm;
            uy = py * iPnorm;
            uz = Ps * iPnorm;

            // Evaluation of the full Thomas-BMT precession vector.
            tbmt_precession_vector(x,ux,uy,uz,gamma,h,gyro_anomaly,Bx,By,Bz,Ex,Ey,Ez,lambdax,lambday,lambdaz);
            lambdax *= half_ds;
            lambday *= half_ds;
            lambdaz *= half_ds;
```

src/elements/Buncher.H
```diff
@@ -90,6 +92,9 @@ namespace impactx::elements
 
             m_neg_kV = -m_k * m_V;
             m_kV_r2bg2 = -m_neg_kV / (2.0_prt * betgam2);
+
+            m_beta = refpart.beta();
+            m_ibeta = 1.0_prt / m_beta;
```
ax3l: In all elements of this PR: I am not sure if `m_ibeta` is used often enough to justify caching it. It is usually used once per particle, doing the division inline based on `m_beta` is probably totally ok.

src/elements/NonlinearLens.H
```diff
@@ -184,6 +189,117 @@ namespace impactx::elements
...
+            croot = re1 - croot;
+            croot = amrex::sqrt(croot);
+
+            // compute carcsin = arcsin(zeta)
+            Complex carcsin = im1 * zeta + croot;
+            carcsin = -im1 * amrex::log(carcsin);
+
+            // compute complex function F'(zeta)
+            Complex dF = zeta / powi<2>(croot);
+            dF = dF + carcsin / powi<3>(croot);
```
ax3l: Computed twice now for spin and ps update in the same kernel. I will add a helper function to avoid calculating this compute intensive part twice.

ax3l: done

src/elements/ShortRF.H
```diff
@@ -101,6 +103,8 @@ namespace impactx::elements
             amrex::ParticleReal const pti_ref = ptf_ref + m_V_cos_phi;
             m_bgf = std::sqrt(powi<2>(ptf_ref) - 1.0_prt);
             m_bgi = std::sqrt(powi<2>(pti_ref) - 1.0_prt);
+            m_gamf = -ptf_ref;
+            m_gami = -pti_ref;
```
ax3l: `m_gamf` & `m_gami`: Those two are not worth caching, sign flips are very cheap to do per particle.

`ptf_ref` is directly accessible via `refpart.pt` in kernel. `pti_ref` can cheaply be recomputed from `refpart.pt` and the cached `m_V_cos_phi` in-kernel.

ax3l:
@cemitch99 a general question on the implementation of half-kicks here:
Excluding `ShortRF`, for the elements here (`Kicker`, `Buncher`, `NonlinearLens`, `ThinDipole`) the fields depend only on coordinates the kick leaves unchanged (`x`, `y`, plus `t` for `Buncher`). Like `TaperedPL`, we could also achieve an exactly reversible kick via midpoint push, that is second order and exactly reversible, too.

midpoint is 1 `tbmt_precession_vector` + 1 `rotate_spin` instead of 2 of each. `rotate_spin` carries a `sincos` and a `sqrt`, so it's the expensive half.
In #1609 I also saw a 2x smaller error constant (midpoint vs. trapezoidal rule behavior) for the midpoint scheme.

Is there a reason to prefer the half-kick structure that I'm missing?

src/elements/ThinDipole.H
ax3l: will push a commit for this.

src/elements/Buncher.H
ax3l: will push to fix

src/elements/ShortRF.H
ax3l: will push to fix

ax3l:
In case midpush sounds good to you, I prepared a PR for this branch that you can apply directly:
https://github.com/cemitch99/impactx/pull/4

If you merge the linked PR to your fork then the PR here will automatically update.

cemitch99:
@ax3l Ok, ready for re-review.

ax3l (approved):
Looks great, thank you! 🎉

PR #1609, opened 2026-08-12
Multipole: 2nd-Order Accurate Spin Push

ax3l:
## Fix s-Symmetric Spin Push for Thin Multipole

Apply the spin rotation once, with the Thomas-BMT generator evaluated at the transverse momentum halfway through the kick. This makes the map second-order accurate in the kick strength and an exact inverse of the reversed element, matching the structure used by `TaperedPL`.

The fields depend only on `(x,y)`, which the kick leaves unchanged, and `pt` is unchanged over the kick, so only the transverse momenta need to be averaged across it.

Measured on a single particle at `x = 1e-3` through a thin quadrupole multipole, forward + `reverse()` + forward:

```
kick             spin rotation    before      after
K=1   (dpx=1e-3)  -5.54e-3 rad    2.770e-09   2.301e-21
K=100 (dpx=0.1)   -0.555 rad      2.791e-03   1.116e-16
```

The residual of the previous one-sided kick scaled as `lambda * dpx^2`, confirmed (locally) across six orders of magnitude and matching the analytic estimate of `2.78e-3` at `K=100`.

## Test

Parametrize `test_Multipole` over the existing octupole and a _strong_ low-order kick (multipole=2, K_normal=300, K_skew=100).

The octupole deflects by only ~1e-13 at this beam size, so it exercises the spin roundtrip in the small-angle limit only. The strong quadrupole rotates the spin by a large angle, where the structure of the spin push decides whether forward + reverse closes.

Against a one-sided spin kick, `[strong_quadrupole-spin]` fails with `||delta spin||` up to `0.73` on a unit vector, while `[strong_quadrupole-nospin]` passes: the phase space map is unaffected.

### DIFF
--- a/src/elements/Multipole.H
+++ b/src/elements/Multipole.H
@@ -191,10 +191,8 @@ namespace impactx::elements
 
         /** This operator pushes the particle spin.
          *
-         * The push is s-symmetric: half spin kick, phase space kick, half
-         * spin kick, with the second half evaluated at the updated momenta.
-         * The map is second-order accurate in the kick strength and an exact
-         * inverse of the reversed element.
+         * The push is s-symmetric: the map is second-order accurate in the
+         * kick strength and an exact inverse of the reversed element.
          *
          * @param x particle position in x
          * @param y particle position in y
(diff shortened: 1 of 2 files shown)


ax3l:
I compared this to a midpoint scheme now, and its about 28% faster (locally), with 2x faster convergence in accuracy (mid-point vs. trapezoidal rule).
A nice side effect is that the midpoint rule is also bit-exact reversible (half-kick is also exact, but "only" to machine precision).

Will switch this PR to midpoint now.

cemitch99 (commented):
Ok, looks good.

PR #1612, opened 2026-08-14
`setup.py`: Find Free-Threaded Python (`Python_FIND_ABI`)

ax3l:
Disabled by default, but our cibuildwheels includes builds for it now
https://github.com/BLAST-ImpactX/impactx-wheels/pull/27
https://cmake.org/cmake/help/latest/module/FindPython.html#variable:Python_FIND_ABI


cedoss-lbl (approved):
Looks like it is correct to me and a narrow scope to not cause unintended issues.  From quickly looking at the context seems like it solves a silent CMake issue, but I am not an expert in this codebase.  Won't hold it up the PR, though :)

ax3l:
Thanks, I admit I did mean to tag @cemitch99 Chris :D Nice to read you, hope you are doing well!

PR #1616, opened 2026-08-17
Diagnostics: Guard Degenerate Beam Moments

ax3l:
Explicitly handle edge cases like flat and cold beams in our beam moments calculation.

Before this PR, these cases caused invalid floating point operations, which showed up in diagnostics and complicate debugging (other) bugs.

<details>

Several reduced beam characteristics were computed without checking that they exist. With `amrex.fpe_trap_invalid` enabled, a beam that is degenerate but otherwise perfectly healthy aborted the run inside its own diagnostics.

### Square roots

`emittance_xd` and `emittance_yd` took the square root of a dispersion-corrected discriminant without the guard their undispersed counterparts `emittance_x`/`_y`/`_t` already had, and the standard deviations took the square root of a variance straight from the moment sums. Both quantities are non-negative mathematically, but finite precision can put them marginally below zero, where `std::sqrt` raises `FE_INVALID`.

A new `sqrt_clamped` clamps that rounding-level case to zero. It passes a NaN through rather than clamping it, so a beam poisoned by a non-finite coordinate stays visible in the output instead of being reported as a zero width.

### Twiss functions

`alpha` and `beta` are ratios to an rms emittance and were divided by it unconditionally. A single particle, a cold plane, or a plane without energy spread has no emittance there — and `emittance_t` is in fact *forced* to zero by the guard immediately above — so the division was a plain `0/0` for any such beam. They now report NaN, which is what an undefined Twiss function is.

Both `reduced_beam_characteristics` overloads (particle container and covariance matrix) are fixed.

### Note on where a NaN detonates

A quiet NaN is silent through arithmetic and `std::sqrt`; it raises `FE_INVALID` only at an *ordered* comparison, which GCC emits as a signaling `comisd`. On GPU builds the device never traps at all. That is why a NaN born anywhere in tracking surfaces as a SIGFPE inside `reduced_beam_characteristics`, on every rank at once, after the all-reduce. `sqrt_clamped` deliberately does not swallow that.

### Tests

`tests/python/test_beam_moments_degenerate.py`: a single particle, a beam without energy spread, and the envelope counterpart of the latter. Verified to fail on `development` before the fix.

Follow-up #1617 stacks on this branch and handles beams with no particles at all.

</details>

### DIFF
--- a/src/diagnostics/ReducedBeamCharacteristics.cpp
+++ b/src/diagnostics/ReducedBeamCharacteristics.cpp
@@ -29,6 +29,7 @@
 #include <AMReX_TypeList.H>             // for TypeMultiplier
 
 #include <array>
+#include <cmath>
 #include <limits>
 #include <vector>
 
@@ -37,6 +38,29 @@ namespace impactx::diagnostics
 {
 namespace
 {
+    /** Square root of a quantity that is mathematically non-negative
+     *
+     * Variances and squared emittances are non-negative by construction, but the
+     * finite-precision moment sums they are recovered from can put them marginally
+     * below zero. std::sqrt of such a value raises FE_INVALID, which aborts the run
+     * under amrex.fpe_trap_invalid, so clamp the (rounding-level) negative case to
+     * zero instead.
+     *
+     * A NaN input is passed through rather than clamped: it means the beam moments
+     * are already poisoned, which should stay visible in the output.
+     *
+     * @param x a quantity that is non-negative up to rounding
+     * @returns sqrt(x), or zero if x is negative
+     */
+    AMREX_FORCE_INLINE
+    amrex::ParticleReal
+    sqrt_clamped (amrex::ParticleReal x)
+    {
+        using namespace amrex::literals; // for _prt
+
+        return (x < 0.0_prt) ? 0.0_prt : std::sqrt(x);
+    }
+
     //! constant coordinate shifts subtracted before forming the beam moments
     struct Shifts
     {
@@ -119,6 +143,8 @@ namespace
         BL_PROFILE("impactx::diagnostics::reduced_beam_characteristics(pc)");
 
         using namespace amrex::literals; // for _prt
+
+        auto const nan = std::numeric_limits<amrex::ParticleReal>::quiet_NaN();
 
         // preparing to access reference particle data: RefPart
         RefPart const ref_part = pc.GetRefParticle();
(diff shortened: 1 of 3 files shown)


src/diagnostics/ReducedBeamCharacteristics.cpp
```diff
@@ -334,19 +360,24 @@ namespace
...
-        amrex::ParticleReal const alpha_x = - xpx_d / emittance_xd;
-        amrex::ParticleReal const alpha_y = - ypy_d / emittance_yd;
-        amrex::ParticleReal const alpha_t = - tpt / emittance_t;
+        amrex::ParticleReal const emittance_yd = sqrt_clamped(y_msd*py_msd-ypy_d*ypy_d);
+        /* Courant-Snyder (Twiss) beta-function and alpha
+         *
+         * Both are ratios to an rms emittance, and are undefined where that emittance
+         * vanishes: a single particle, a cold plane, or a longitudinal plane without
+         * energy spread. Report NaN there rather than dividing by zero, which raises
+         * FE_DIVBYZERO (or FE_INVALID, for 0/0) under amrex.fpe_trap_zero/_invalid.
```
ax3l:
```suggestion
         * energy spread.
```

src/diagnostics/ReducedBeamCharacteristics.cpp
```diff
@@ -546,19 +579,24 @@ namespace
...
```
ax3l:
```suggestion
         * energy spread.
```

tests/python/test_beam_moments_degenerate.py
```diff
@@ -0,0 +1,116 @@
...
+#
+# Authors: Axel Huebl
+# License: BSD-3-Clause-LBNL
+#
+# -*- coding: utf-8 -*-
+
+import math
+
+import impactx
+from impactx import ImpactX, distribution
```
ax3l:
```suggestion
from impactx import ImpactX, create_envelope, distribution
```

tests/python/test_beam_moments_degenerate.py
```diff
@@ -0,0 +1,116 @@
...
+    sim = ImpactX()
+    sim.particle_shape = 2
+    sim.space_charge = False
+    sim.slice_step_diagnostics = False
+    sim.init_grids()
+
+    ref = sim.beam.ref
+    ref.set_species("electron").set_kin_energy_MeV(2.0e3)
+
+    envelope = impactx.create_envelope(waterbag(lambda_pt=0.0), 1.0e-9)
```
ax3l:
```suggestion
    envelope = create_envelope(waterbag(lambda_pt=0.0), 1.0e-9)
```

tests/python/test_beam_moments_degenerate.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022-2026 The ImpactX Community
+#
+# Authors: Axel Huebl
```
ax3l:
```suggestion
# Authors: Axel Huebl, Chad Mitchell
```

src/diagnostics/ReducedBeamCharacteristics.cpp
```diff
@@ -37,6 +38,29 @@ namespace impactx::diagnostics
...
+     * @returns sqrt(x), or zero if x is negative
+     */
+    AMREX_FORCE_INLINE
+    amrex::ParticleReal
+    sqrt_clamped (amrex::ParticleReal x)
+    {
+        using namespace amrex::literals; // for _prt
+
+        return (x < 0.0_prt) ? 0.0_prt : std::sqrt(x);
+    }
```
cemitch99, line 62: Great; I'm surprised that there is not already a C++ or AMReX function that is equivalent to this, that could be used everywhere.

cemitch99 (approved):
Looks good.

PR #1617, opened 2026-08-17
Diagnostics: Beam Moments Without a Beam

ax3l:
The beam moments calculation can be triggered before the simulation.

For use cases where the beam is only initialized in the first element as a `Source`, the run on an empty beam should produce clean, non-failing output that shows the beam moments as undefined, similar to #1616.

<details>

Simulations that load their beam from a `source` element hold no particles when the initial diagnostics run, and the reference particle is only restored from that file later on. Both broke the step-0 diagnostics:

* every beam moment normalized by a total weight of zero, and
* `beta*gamma = sqrt(gamma^2 - 1)` was evaluated for a reference particle of `gamma = 0` — in `reduced_beam_characteristics` *and* in `write_ref`, which runs first, so `diags/ref_particle` crashed before the moments were ever reached.

Under `amrex.fpe_trap_invalid` the run aborted before it started; without the traps it wrote a mix of NaN, zero and reduction identity elements that reads like a measurement.

### What changed

`reduced_beam_characteristics` normalizes by one instead of by zero, which carries the (all-zero) sums through the derived quantities without an exception, and reports the results that carry no information as NaN. `write_ref` reports `beta`, `gamma` and `beta*gamma` of an uninitialized reference particle as NaN.

A beam of zero total weight is not necessarily empty: `bunch_charge=0` massless test particles are a supported workflow (`examples/multipole/run_sextupole_kick.py`, `tests/python/test_chracc_Ez0.py`). Their minima and maxima are unweighted and stay defined. Telling that case apart costs no extra communication — the shift-sampling all-reduce above already establishes, globally, whether any rank owns a particle.

The emptiness test is an equality on purpose: `==` is quiet on a NaN, so a beam poisoned by a non-finite particle coordinate is not absorbed here but keeps tripping the FPE traps where they are enabled.

For a beam that does have particles and weight, `w_norm == w_sum` and every value is bit-identical to before.

### Tests

Added to `tests/python/test_beam_moments_degenerate.py`: an empty beam, an empty beam without a reference particle, and a beam of zero charge. Plus `test_beam_moments_do_not_raise_fpe`, which runs all the degenerate cases in a subprocess under `amrex.fpe_trap_invalid`/`_zero` — `tests/python/conftest.py` sets `amrex.signal_handling=0`, so the traps are otherwise never active in the suite, and a trap kills the process outright (exit 8 via `MPI_Abort`), which no in-process test could survive.

Verified to fail on `development` before the fix.

</details>

### DIFF
--- a/src/diagnostics/DiagnosticOutput.cpp
+++ b/src/diagnostics/DiagnosticOutput.cpp
@@ -114,10 +114,7 @@ namespace
         using namespace amrex::literals; // for _prt
 
         /* A reference particle that carries no energy yet, e.g. before a Source
-         * element loads it from file, has gamma = 0. Neither beta = sqrt(1 - 1/gamma^2)
-         * nor beta*gamma = sqrt(gamma^2 - 1) is a real number there, and evaluating
-         * them raises FE_INVALID (and FE_DIVBYZERO) under amrex.fpe_trap_invalid.
-         * Report the undefined quantities as NaN instead.
+         * element loads it from file, has gamma = 0 and thus no valid beta.
          */
         auto const nan = std::numeric_limits<amrex::ParticleReal>::quiet_NaN();
         bool const ref_is_initialized = ref_part.gamma() >= 1.0_prt;
--- a/src/diagnostics/DiagnosticOutput.cpp
+++ b/src/diagnostics/DiagnosticOutput.cpp
@@ -111,10 +111,21 @@ namespace
         int step
     )
     {
+        using namespace amrex::literals; // for _prt
+
+        /* A reference particle that carries no energy yet, e.g. before a Source
+         * element loads it from file, has gamma = 0. Neither beta = sqrt(1 - 1/gamma^2)
+         * nor beta*gamma = sqrt(gamma^2 - 1) is a real number there, and evaluating
+         * them raises FE_INVALID (and FE_DIVBYZERO) under amrex.fpe_trap_invalid.
+         * Report the undefined quantities as NaN instead.
+         */
+        auto const nan = std::numeric_limits<amrex::ParticleReal>::quiet_NaN();
+        bool const ref_is_initialized = ref_part.gamma() >= 1.0_prt;
+
         amrex::ParticleReal const s = ref_part.s;
-        amrex::ParticleReal const beta = ref_part.beta();
-        amrex::ParticleReal const gamma = ref_part.gamma();
-        amrex::ParticleReal const beta_gamma = ref_part.beta_gamma();
+        amrex::ParticleReal const beta = ref_is_initialized ? ref_part.beta() : nan;
+        amrex::ParticleReal const gamma = ref_is_initialized ? ref_part.gamma() : nan;
+        amrex::ParticleReal const beta_gamma = ref_is_initialized ? ref_part.beta_gamma() : nan;
         amrex::ParticleReal const x = ref_part.x;
         amrex::ParticleReal const y = ref_part.y;
         amrex::ParticleReal const z = ref_part.z;
--- a/src/diagnostics/ReducedBeamCharacteristics.cpp
+++ b/src/diagnostics/ReducedBeamCharacteristics.cpp
@@ -31,6 +31,7 @@
 #include <array>
 #include <cmath>
 #include <limits>
+#include <string>
 #include <vector>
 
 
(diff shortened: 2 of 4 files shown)


cemitch99:
In Python test:

stderr:
  Erroneous arithmetic operation
  addr2line: '/tmp/pytest-of-runner/pytest-0/test_beam_moments_do_not_raise0/amrex': No such file

ax3l:
I fixed the failing `test_beam_moments_do_not_raise_fpe` test, but this one is so hacky that we will not add it here. Was just good for local verification, will clean and remove now.

src/diagnostics/DiagnosticOutput.cpp
```diff
@@ -111,10 +111,21 @@ namespace
...
     )
     {
+        using namespace amrex::literals; // for _prt
+
+        /* A reference particle that carries no energy yet, e.g. before a Source
+         * element loads it from file, has gamma = 0. Neither beta = sqrt(1 - 1/gamma^2)
+         * nor beta*gamma = sqrt(gamma^2 - 1) is a real number there, and evaluating
+         * them raises FE_INVALID (and FE_DIVBYZERO) under amrex.fpe_trap_invalid.
+         * Report the undefined quantities as NaN instead.
+         */
```
ax3l:
```suggestion
         * element loads it from file, has gamma = 0 and thus no valid beta.
         */
```

src/diagnostics/ReducedBeamCharacteristics.cpp
```diff
@@ -150,8 +151,15 @@ namespace
         RefPart const ref_part = pc.GetRefParticle();
         // reference particle charge in C
         amrex::ParticleReal const q_C = ref_part.charge;
+        /* A reference particle that carries no energy yet, e.g. before a Source
+         * element loads it from file, has gamma = 0, for which beta*gamma =
+         * sqrt(gamma^2 - 1) is not a real number. Keep it out of the arithmetic
+         * below - such a beam reports "no beam" moments at the end anyway - so
+         * that we do not raise FE_INVALID under amrex.fpe_trap_invalid.
+         */
```
ax3l, line 156:
```suggestion
        /* A reference particle that carries no energy yet, e.g. before a Source
         * element loads it from file, has gamma = 0 and thus no valid beta.
         */
```

src/diagnostics/ReducedBeamCharacteristics.cpp
```diff
@@ -202,7 +211,10 @@ namespace
             int src_rank = found ? amrex::ParallelDescriptor::MyProc()
                                  : amrex::ParallelDescriptor::NProcs();
             amrex::ParallelAllReduce::Min(src_rank, amrex::ParallelDescriptor::Communicator());
-            if (src_rank < amrex::ParallelDescriptor::NProcs()) {
+            // this also tells us, globally and without a further reduction,
+            // whether the beam holds any particle at all
```
ax3l:
```suggestion
            // Does the beam hold any particle at all:
```

src/diagnostics/ReducedBeamCharacteristics.cpp
```diff
@@ -272,16 +284,33 @@ namespace
         // parallel-axis theorem. The shift keeps every sum at the O(rms^2) scale,
         // so the central moments below stay well-conditioned even in single precision.
         amrex::ParticleReal const w_sum = values_sum[0];
+
+        /* Every moment below is normalized by the beam's total weight, which is zero
+         * for a beam that holds no particles at all - e.g. the initial diagnostic of
+         * a simulation whose beam a Source element loads further down the lattice -
+         * and also for one of zero charge, as used for massless "test" particles.
+         * Both would turn each normalization into 0/0, which raises FE_INVALID under
+         * amrex.fpe_trap_invalid. Normalize by one instead, which carries the
```
ax3l:
```suggestion
         * Both would turn each normalization into invalid divisions by zero.
         * Normalize by one instead, which carries the
```

src/diagnostics/ReducedBeamCharacteristics.cpp
```diff
@@ -272,16 +284,33 @@ namespace
...
+         * (all-zero) sums through the derived quantities without an exception, and
+         * report the undefined results as NaN at the end of this function.
+         *
+         * The comparison is an equality on purpose: it is quiet for a NaN w_sum,
+         * so a beam poisoned by a non-finite particle coordinate is not silently
+         * absorbed here, but keeps tripping the FPE traps where they are enabled.
```
ax3l:
```suggestion
         * absorbed here, but keeps tripping FPE diagnostics when enabled.
```

src/diagnostics/ReducedBeamCharacteristics.cpp
```diff
@@ -517,6 +546,39 @@ namespace
...
+            if (has_particles)
+            {
+                constexpr std::array coordinates = {"x", "y", "t", "px", "py", "pt"};
+                for (std::size_t c = 0; c < coordinates.size(); ++c)
+                {
+                    std::string const coordinate = coordinates[c];
+                    data["min_" + coordinate] = values_min[c];
+                    data["max_" + coordinate] = values_max[c];
+                    // deprecated spellings of the same two entries
+                    data[coordinate + "_min"] = values_min[c];
```
ax3l: This looks more complicated than it needs to be. The logic of skipping could go into the
```C++
for (auto & moment : data) {
   moment.second = nan;
}
```
Look by matching prefix/suffix/exact instead of recovering from above variables.

cemitch99 (approved):

PR #1618, opened 2026-08-17
Diagnostics: Fix `min`/`max` Momentum Columns

ax3l:
The `reduced_beam_characteristics` ASCII output repeated `mean_px` in the `min_px` and `max_px` column headers, and likewise for `py` and `pt`:

```cpp
<< rbc.at("mean_px") << " " << rbc.at("mean_px") << " " << rbc.at("mean_px") << " "
```

<details>

The momentum extrema are exactly what one reads to spot a runaway particle, so they were the least useful values to lose.

The reduced beam characteristics themselves always held the extrema; only this writer picked the wrong keys out of them. `ImpactXParticleContainer.beam_moments()` and the openPMD beam monitor were never affected, and the file's header row was already correct — only the data columns under it were wrong.

Long-standing; not introduced by #1557, which last touched these lines.

### Tests

`tests/python/test_reduced_beam_characteristics.py` tracks a beam through a drift and checks the written columns against `beam_moments()`, and that `min < mean < max`. Verified to fail on `development` before the fix.

Independent of #1616 and #1617.

</details>


cemitch99 (approved):
Good catch.  This highlights the issue that few of our automated make direct use of the rbc output, so issues like this can be missed.

ax3l:
I admittedly would like to reduce the ASCII output more. They were a quick hack before we had proper APIs, are slow and brittle.

With the APIs, users can quickly create (slow) ASCII when needed.

PR #1619, opened 2026-08-18
Docs: Boost on Perlmutter (synmadx)

ax3l:
Now that we optionally support `synmadx`, which needs (header-only) Boost, add Boost to the Perlmutter (NERSC) dependency install scripts and profiles, following what WarpX does on the same machine.

<details>

### Changes

- `install_cpu_dependencies.sh` / `install_gpu_dependencies.sh`: install Boost 1.82.0 into `${SW_DIR}/boost-1.82.0` (same version as WarpX installs on Perlmutter)
- `perlmutter_cpu_impactx.profile.example` / `perlmutter_gpu_impactx.profile.example`: export the Boost prefix via `CMAKE_PREFIX_PATH`
- `perlmutter.rst`: add `-DImpactX_SYNMADX=ON` to the documented Python builds, which is where the parser is exposed as `impactx.synmadx`

### Notes on the two deviations from WarpX

WarpX needs compiled Boost.Math for the PICSAR QED tables; ImpactX only needs the header-only libraries (Spirit, Phoenix, Fusion, Optional, Math constants, Algorithm/String), which `find_package(Boost 1.70.0 CONFIG REQUIRED)` + `Boost::headers` consume. So:

1. We install with `--with-libraries=headers` instead of `--with-libraries=math`. This is Boost's [`libs/headers`](https://github.com/boostorg/headers) pseudo-library, whose `install` target installs the headers *and* `BoostConfig.cmake`.
2. The profiles export only `CMAKE_PREFIX_PATH`, no `LD_LIBRARY_PATH` — there is no runtime library to find.

As in WarpX, the Boost sources are unpacked in `$HOME` rather than in the tmpfs build directory, since the unpacked tree is too large for `/tmp`.

### Verification

The install recipe was run end-to-end locally (not on Perlmutter):

- `./bootstrap.sh --with-libraries=headers --prefix=… && ./b2 install` produces `include/boost/` plus `lib/cmake/Boost-1.82.0/BoostConfig.cmake` and `lib/cmake/boost_headers-1.82.0/`, and no compiled libraries
- with only that prefix on `CMAKE_PREFIX_PATH`, `find_package(Boost 1.70.0 CONFIG REQUIRED)` resolves it (`Boost_VERSION = 1.82.0`)
- a translation unit with the exact header set of `src/synmadx/synergia/lattice/mx_parse.cc`, including a live Spirit Qi grammar, compiles and links against `Boost::headers` at C++20

</details>

### DIFF
--- a/docs/source/install/hpc/perlmutter-nersc/install_cpu_dependencies.sh
+++ b/docs/source/install/hpc/perlmutter-nersc/install_cpu_dependencies.sh
@@ -51,6 +51,19 @@ curl -Lo ${build_dir}/ccache.tar.xz https://github.com/ccache/ccache/releases/do
 tar -xf ${build_dir}/ccache.tar.xz -C ${build_dir}
 mv ${build_dir}/ccache-4.10.2-linux-x86_64 ${SW_DIR}/ccache-4.10.2
 
+# Boost (header-only libraries, for the synmadx MAD-X lattice parser)
+#   unpacked in $HOME: the Boost sources are too large for the tmpfs build dir
+rm -rf $HOME/src/boost-temp
+mkdir -p $HOME/src/boost-temp
+curl -Lo $HOME/src/boost-temp/boost.tar.gz https://archives.boost.io/release/1.82.0/source/boost_1_82_0.tar.gz
+tar -xzf $HOME/src/boost-temp/boost.tar.gz -C $HOME/src/boost-temp
+(
+  cd $HOME/src/boost-temp/boost_1_82_0
+  ./bootstrap.sh --with-libraries=headers --prefix=${SW_DIR}/boost-1.82.0
+  ./b2 install -j 16
+)
+rm -rf $HOME/src/boost-temp
+
 # c-blosc (I/O compression)
 if [ -d $HOME/src/c-blosc ]
 then
--- a/docs/source/install/hpc/perlmutter-nersc/install_cpu_dependencies.sh
+++ b/docs/source/install/hpc/perlmutter-nersc/install_cpu_dependencies.sh
@@ -43,6 +43,9 @@ python3 -m pip uninstall -qqq -y mpi4py 2>/dev/null || true
 # General extra dependencies ##################################################
 #
 
+# build parallelism
+PARALLEL=16
+
 # tmpfs build directory: avoids issues often seen with $HOME and is faster
 build_dir=$(mktemp -d)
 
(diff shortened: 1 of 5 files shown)


docs/source/install/hpc/perlmutter-nersc/install_cpu_dependencies.sh
```diff
@@ -51,6 +51,19 @@ curl -Lo ${build_dir}/ccache.tar.xz https://github.com/ccache/ccache/releases/do
 tar -xf ${build_dir}/ccache.tar.xz -C ${build_dir}
 mv ${build_dir}/ccache-4.10.2-linux-x86_64 ${SW_DIR}/ccache-4.10.2
 
+# Boost (header-only libraries, for the synmadx MAD-X lattice parser)
+#   unpacked in $HOME: the Boost sources are too large for the tmpfs build dir
```
ax3l: This is wrong, use our build dir.

docs/source/install/hpc/perlmutter-nersc/install_cpu_dependencies.sh
```diff
@@ -51,6 +51,19 @@ curl -Lo ${build_dir}/ccache.tar.xz https://github.com/ccache/ccache/releases/do
...
+rm -rf $HOME/src/boost-temp
+mkdir -p $HOME/src/boost-temp
+curl -Lo $HOME/src/boost-temp/boost.tar.gz https://archives.boost.io/release/1.82.0/source/boost_1_82_0.tar.gz
+tar -xzf $HOME/src/boost-temp/boost.tar.gz -C $HOME/src/boost-temp
+(
+  cd $HOME/src/boost-temp/boost_1_82_0
+  ./bootstrap.sh --with-libraries=headers --prefix=${SW_DIR}/boost-1.82.0
+  ./b2 install -j 16
```
ax3l: Add `-std=c++20` as in https://github.com/BLAST-WarpX/warpx/blob/d48df61a1613d24909438579c9e669df8ff1ee57/Tools/machines/perlmutter-nersc/install_gpu_dependencies.sh#L60-L68

Let's also add a `PARALLEL` control var as in
https://github.com/BLAST-WarpX/warpx/blob/d48df61a1613d24909438579c9e669df8ff1ee57/Tools/machines/perlmutter-nersc/install_gpu_dependencies.sh#L47-L48

docs/source/install/hpc/perlmutter-nersc/install_gpu_dependencies.sh
```diff
@@ -51,6 +51,19 @@ curl -Lo ${build_dir}/ccache.tar.xz https://github.com/ccache/ccache/releases/do
...
+# Boost (header-only libraries, for the synmadx MAD-X lattice parser)
+#   unpacked in $HOME: the Boost sources are too large for the tmpfs build dir
+rm -rf $HOME/src/boost-temp
+mkdir -p $HOME/src/boost-temp
+curl -Lo $HOME/src/boost-temp/boost.tar.gz https://archives.boost.io/release/1.82.0/source/boost_1_82_0.tar.gz
+tar -xzf $HOME/src/boost-temp/boost.tar.gz -C $HOME/src/boost-temp
+(
+  cd $HOME/src/boost-temp/boost_1_82_0
+  ./bootstrap.sh --with-libraries=headers --prefix=${SW_DIR}/boost-1.82.0
+  ./b2 install -j 16
```
ax3l: 2x: see above in cpu version for changes

docs/source/install/hpc/perlmutter-nersc/install_cpu_dependencies.sh
ax3l: Fixed in c3a3a262b — the tarball and the unpacked tree now live in `${build_dir}` and are removed right after the install, like the other dependencies.

docs/source/install/hpc/perlmutter-nersc/install_cpu_dependencies.sh
ax3l: Both done in c3a3a262b: `./b2 cxxflags="-std=c++20" install -j ${PARALLEL}`, with `PARALLEL=16` above `build_dir`.

I also routed the c-blosc `--parallel 16` and ADIOS2 `-j 16` through `${PARALLEL}`, so it is an actual control variable rather than one that governs a single line. Happy to drop those two hunks if you would rather keep the diff to Boost only.

One note for the record: since `--with-libraries=headers` compiles nothing, `cxxflags` has no effect here — `b2` only copies headers and writes `BoostConfig.cmake`. Kept for parity with the WarpX script.

docs/source/install/hpc/perlmutter-nersc/install_gpu_dependencies.sh
ax3l: Applied identically in the GPU script in c3a3a262b — the Boost block is byte-identical between the CPU and GPU versions, and `PARALLEL` is wired up the same way in both.

egstern (commented):
looks OK.

cemitch99 (approved):
Looks good to me.

PR #1623, opened 2026-08-20
Source Element: Choose Which Step to Load

### LINKED ISSUE #1284: Source Element: Choose Which Iteration to Load
https://github.com/BLAST-ImpactX/impactx/blob/26.01/src/elements/Source.cpp#L46

Request by @egstern: let the user choose which openPMD iteration to load. Useful when loading from a monitor that has multiple outputs, eg., in a ring.

ax3l:
The `source` element read the last openPMD iteration of a series.
In a ring, a beam monitor might write one step out per turn, so restarting from a specific turn was not possible.

This adds two options that select which step to load, of which at most one is set:

- `load_step`: the ImpactX step at which the beam monitor wrote the beam, which is stored as the openPMD iteration in the file
- `load_step_index`: the position of the step in the file, `0` is the first step and `-1` the last, counting back from it as in Python

If neither is set, the last step in the file is loaded, as before.

```python
# an ImpactX step stored in the file
push(beam, elements.Source("openPMD", "diags/openPMD/m1.h5", load_step=3))

# the second to last step in the file
push(beam, elements.Source("openPMD", "diags/openPMD/m1.h5", load_step_index=-2))
```

```ini
source.type = source
source.distribution = openPMD
source.openpmd_path = "../ring/diags/openPMD/m1.h5"
source.load_step = 3
# or, by position in the file:
#   source.load_step_index = -2
```

The step numbers stored in a series are usually not consecutive, because the global step counter also advances in the elements between two monitors.
Use `load_step_index` when they are not known, e.g. to continue from the last turn that a ring wrote.

A step that is not in the file aborts with the steps that are in it, e.g.

```
Source: load_step 7 not found in ../solenoid/diags/openPMD/m1.h5 (2 available steps: 1, 3)
```

The same list is printed for an index that reaches past either end of the file.
It is elided in the middle above 100 steps, so that a ring that wrote one step per turn does not print thousands of numbers:

```
Source: load_step 99999 not found in diags/openPMD/m1.h5 (102 available steps: 1, 3, 4, 6, ..., 151, 153)
```

Setting both options, or a negative `load_step` (which is `load_step_index` instead), is an error as well.
This is checked on construction and again when the step is read, because both options can also be set on an existing element.

Both options are keyword arguments in front of `name`, so `name` moves in the positional order of the constructor and needs to be passed as a keyword argument, as everywhere else in ImpactX.

An empty series is now caught as well, instead of dereferencing the end of an empty container.

Selecting a step works for every iteration encoding that the `beam_monitor` writes.
ImpactX requires openPMD-api 0.17.0 or newer, and as of 0.17 a `v` (variable based) series reads back in random access from BP4 and BP5 files, which is what the `source` element uses.
The `beam_monitor` documentation still carries a note from before that, which is fixed in a separate PR.

Tested in the new `tests/python/test_source_load_step.py`: a monitor series with two steps is read back by step number, by position from both ends, and by the default; the error paths, the elided step list of a series with more than 100 steps, and the `repr`/`to_dict` round-trip of both options are checked as well.

Closes #1284

## To Do

- [x] draft 🤖 generated with [Claude Code](https://claude.com/claude-code)
- [x] self-review
- [x] bot-review
- [x] finalize

### DIFF
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -2072,21 +2072,31 @@ When ImpactX needs to sort particles spatially, it will redistribute them over M
 
 .. pp:param:: <source_name>.load_step
     :type: ``integer``
-    :default: ``-1``
+    :default: the last step in the file
 
-    Which step to load from the openPMD series.
+    Which step (iteration) to load from the openPMD series, selected by step number.
 
-    A value :math:`\geq 0` is the ImpactX step at which the ``beam_monitor`` wrote the beam, which is
-    stored as the openPMD iteration in the file.
+    This is the ImpactX step at which the ``beam_monitor`` wrote the beam, which is stored as the
+    openPMD iteration in the file.
     These step numbers are usually not consecutive, because the global step counter also advances in
-    the elements between two monitors: list them with, e.g., ``openpmd-ls`` before selecting one.
-    A step that is not in the file is an error, which lists the available steps.
+    the elements between two monitors: list them with, e.g., `openpmd-ls <https://openpmd-api.readthedocs.io/en/0.17.1/utilities/cli.html>`__ before selecting one.
+    A step that is not in the file is an error, which lists the steps that are in it.
 
-    ``-1`` reads the last step in the file and more negative values count back from it, e.g. ``-2``
-    reads the second to last step.
+    Set at most one of ``load_step`` and ``load_step_index``.
 
-    Selecting a step reads the series in random access, which the ``v`` (variable based) iteration
-    encoding of the ``beam_monitor`` does not support.
+.. pp:param:: <source_name>.load_step_index
+    :type: ``integer``
+    :default: the last step in the file
+
+    Which step (iteration) to load from the openPMD series, selected by position in the file.
+
+    ``0`` is the first step in the file and ``-1`` is the last, counting back from it as in Python,
+    e.g. ``-2`` is the second to last step.
+    Use this when the step numbers in the file are not known, e.g. to continue from the last turn
+    that a ring wrote.
+    An index that reaches past either end of the file is an error, which lists the steps in it.
+
+    Set at most one of ``load_step`` and ``load_step_index``.
 
 
 ``spin_map``
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -2072,7 +2072,7 @@ When ImpactX needs to sort particles spatially, it will redistribute them over M
 
 .. pp:param:: <source_name>.load_step
     :type: ``integer``
-    :default: the last step in the file
+    :default: unset (if neither option is set, the last step in the file is loaded)
 
     Which step (iteration) to load from the openPMD series, selected by step number.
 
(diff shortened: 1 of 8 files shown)


egstern:
could you make is so that you can select by iteration number (as stored in the iterations list) or index number, indexed into the iteration list?

docs/source/usage/parameters.rst
```diff
@@ -2070,6 +2070,24 @@ When ImpactX needs to sort particles spatially, it will redistribute them over M
     The existing reference particle in that case needs to be manually configured before the tracking loop.
     This option acts during particle tracking, it has no effect in envelope or reference-particle-only tracking.
 
+.. pp:param:: <source_name>.load_step
+    :type: ``integer``
+    :default: ``-1``
+
+    Which step to load from the openPMD series.
```
```suggestion
    Which step (iteration) to load from the openPMD series.
```

docs/source/usage/parameters.rst
```diff
@@ -2070,6 +2070,24 @@ When ImpactX needs to sort particles spatially, it will redistribute them over M
...
+
+    A value :math:`\geq 0` is the ImpactX step at which the ``beam_monitor`` wrote the beam, which is
+    stored as the openPMD iteration in the file.
+    These step numbers are usually not consecutive, because the global step counter also advances in
+    the elements between two monitors: list them with, e.g., ``openpmd-ls`` before selecting one.
```
```suggestion
    the elements between two monitors: list them with, e.g., `openpmd-ls <https://openpmd-api.readthedocs.io/en/0.17.1/utilities/cli.html>`__ before selecting one.
```

docs/source/usage/parameters.rst
```diff
@@ -2070,6 +2070,24 @@ When ImpactX needs to sort particles spatially, it will redistribute them over M
...
+    A step that is not in the file is an error, which lists the available steps.
+
+    ``-1`` reads the last step in the file and more negative values count back from it, e.g. ``-2``
+    reads the second to last step.
```
ax3l: To clarify: if we write `100, 200, 300, 400`, would `-2` read back `300` or error?

I think this is all in terms of `steps` (iteration), so `-1` would match 400 and `-101` would be needed to load `300`?

docs/source/usage/parameters.rst
```diff
@@ -2070,6 +2070,24 @@ When ImpactX needs to sort particles spatially, it will redistribute them over M
...
+
+    Selecting a step reads the series in random access, which the ``v`` (variable based) iteration
+    encoding of the ``beam_monitor`` does not support.
```
ax3l: variable based: To double check if still true with openPMD-api 0.17.1, probably not, improve implementation and remove this comment.

docs/source/usage/python.rst
```diff
@@ -2013,8 +2013,23 @@ For an element with ``nslice`` > 1, the pushes and maps refer to a single ``ds/n
...
    :param name: an optional name for the element
 
+   .. note::
+
+      A ``load_step`` :math:`\geq 0` is the ImpactX step at which the :py:class:`impactx.elements.BeamMonitor`
+      wrote the beam, which is stored as the openPMD iteration in the file.
+      These step numbers are usually not consecutive, because the global step counter also advances in
+      the elements between two monitors: list them with, e.g., ``openpmd-ls`` before selecting one.
+      A step that is not in the file is an error, which lists the available steps.
+
```
ax3l: Needs to mirror modifications / clarifications we do in `parameters.rst`

src/python/elements.cpp
```diff
@@ -2522,6 +2526,14 @@ void init_elements(py::module& m)
             [](Source & src, bool load_ref_particle) { src.m_load_ref_particle = load_ref_particle; },
             "Restore the reference particle from the species metadata of the openPMD file (particle tracking only)."
         )
+        .def_property("load_step",
+            [](Source & src) { return src.m_load_step; },
+            [](Source & src, int load_step) { src.m_load_step = load_step; },
+            "Which step to load from the openPMD series: a value >= 0 is the ImpactX step at "
```
```suggestion
            "Which step (iteration) to load from the openPMD series: a value >= 0 is the ImpactX step at "
```

ax3l:
@egstern good idea, functionality added :+1:

egstern (approved):
Looks good as far as I can tell and the documentation is good too!

PR #1624, opened 2026-08-21
Perlmutter CUDA 13.2 & Consistent Purge of JIT Caches

ax3l:
HPC install scripts: Ensure existing PIP and JIT Python caches do not survive system upgrades and pull in stale dependencies, e.g., on old CUDA toolkits.

<details>

## X-ref

- WarpX: https://github.com/BLAST-WarpX/warpx/pull/7183 (same change, all `Tools/machines/` scripts)

## Summary

Often, after HPC system upgrades, pip caches have wheels in them that still depend on old system libs, old system modules or old CUDA runtimes that are then incompatible when combined together in Python runs/imports.

This adds a cache purge to every `install*` script in `docs/source/install/hpc/`, early in the "Remove old dependencies" section, before any dependency is installed.

## Details

User-facing:
- every HPC `install*` script now runs `python3 -m pip cache purge` and additionally wipes the GPU kernel/JIT caches `~/.cupy/kernel_cache`, `~/.nv/ComputeCache`, `~/.cache/numba` and `~/.triton`, which `pip cache purge` does not cover but which go equally stale over a CUDA/ROCm runtime upgrade

Internal:
- the purge moved up from the venv section into the "Remove old dependencies" section; the later, then redundant `pip cache purge` calls are removed, so each script purges exactly once
- the purge is guarded with `|| true` so that a disabled cache or a pip too old for the `cache` subcommand does not abort the script under `set -e`

## Testing

`bash -n` passes on all four scripts.

</details>

### DIFF
--- a/docs/source/install/hpc/lumi-csc/install_cpu_dependencies.sh
+++ b/docs/source/install/hpc/lumi-csc/install_cpu_dependencies.sh
@@ -31,6 +31,10 @@ python3 -m pip uninstall -qq -y pyimpactx
 python3 -m pip uninstall -qq -y impactx
 python3 -m pip uninstall -qqq -y mpi4py 2>/dev/null || true
 
+# clean out caches, e.g., depending on old system modules
+python3 -m pip cache purge || true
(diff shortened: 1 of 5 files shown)


egstern (approved):
Good cleanup

PR #1625, opened 2026-08-22
[WIP] Python: Run Multiple Simulations Per Process

ax3l:
Running a second ImpactX simulation in the same process (parameter scans, optimization loops, notebooks) aborted in `init_grids`:

```
Attempting to use an MPI routine (internal_Comm_dup) before initializing or after finalizing MPICH
```

unless the user initialized MPI before ImpactX, e.g., by importing `mpi4py.MPI`, as our optimization examples do.

## MPI lifetime

`amrex::Initialize` calls `MPI_Init` if MPI is not yet initialized, and then `amrex::Finalize` (called from `ImpactX::finalize`) also calls `MPI_Finalize`. MPI cannot be initialized again after that, so the first simulation was the last one of its process.

ImpactX now initializes MPI itself when nobody else did, via `ablastr::parallelization::mpi_init` (same helper as WarpX: picks the MPI thread level and carries the OLCFDEV-1655 workaround), and finalizes it at the end of the process. The two modes we support are unchanged from a user's point of view:

- **externally controlled MPI** (`main()`, mpi4py, ...): detected via `MPI_Initialized`, ImpactX does not touch its lifetime;
- **ImpactX/AMReX-owned MPI**: now spans the whole process instead of a single `amrex::Initialize`/`Finalize` cycle.

## Inputs leaked into the next simulation

`amrex::Initialize` only initializes ParmParse if a command line is passed (`argc >= 1`, `AMReX.cpp`); with ImpactX' `argc = 0`, `ParmParse::Initialize` never ran, so `ParmParse::Finalize` was never registered and the parameter table was never cleared. A second simulation silently inherited the inputs of the first one:

```
run1(space charge ON ): sim.space_charge = '3D'   sig_x = 1.19128186e-04
run2(space charge OFF): sim.space_charge = '3D'   sig_x = 1.19128186e-04   # never requested space charge
```

We now pass a program name if we have no command line. (An alternative fix would be in AMReX: treat `argc == 0` like `argc == 1` when `build_parm_parse` is set.)

## AMReX ownership

`ImpactX::finalize` called `amrex::Finalize` even when pyAMReX or another code in the same process owns AMReX. It now finalizes AMReX only if ImpactX initialized it.

## Third commit: MPI thread level of the executable

`src/main.cpp` called plain `MPI_Init`, so the standalone executable ran at `MPI_THREAD_SINGLE` even though `ImpactX_MPI_THREAD_MULTIPLE` is `ON` by default (async I/O). It now uses the same ABLASTR helpers as the Python path, and reports `MPI initialized with thread support level 3` in this build.

Not added on purpose: `ablastr::parallelization::check_mpi_thread_level()`, which WarpX calls. It records a *medium*-priority warning when the provided level is *stricter* than the requested one, and our example CTests run with `impactx.abort_on_warning_threshold=low` - that would abort them, e.g., wherever MPICH provides `MPI_THREAD_MULTIPLE` for a build that only requests `FUNNELED`.

## Second commit: warning inputs before `init_grids`

Setting `sim.abort_on_warning_threshold` or `sim.abort_on_unused_inputs` before `sim.init_grids()` segfaults (also in released versions): both setters call `init_warning_logger`, and the ABLASTR warning manager reads the MPI rank in its constructor. The setters now only touch the warning logger if AMReX is initialized; `init_grids` sets it up from the same inputs. Happy to split this into its own PR.

## Tests

`tests/python/test_lifecycle.py`: two simulations in one process, no input leak, externally initialized AMReX is not finalized, and warning inputs before `init_grids`. The scenarios run in subprocesses that do not import mpi4py, because `conftest.py` initializes MPI and AMReX for the pytest process and would mask the ImpactX-owned mode.

Verified locally, in addition: element GPU data (`RFCavity`, `PolygonAperture`), openPMD `BeamMonitor`, IGF/FFT space charge and 2 MPI ranks across cycles; scan loops that never call `finalize()`; mpi4py imported before or after ImpactX; the standalone executable (serial and `mpiexec -n 2`).

## Known limitation (follow-up, in ABLASTR)

The ABLASTR `WarnManager` is a static singleton that is never reset, so warnings recorded by one simulation are printed again by the next one in the same process, and can trip `abort_on_warning_threshold` there. Fix belongs into WarpX/ABLASTR; a test for it will follow here once that lands.

### DIFF
--- a/src/initialization/InitAMReX.cpp
+++ b/src/initialization/InitAMReX.cpp
@@ -11,6 +11,8 @@
 
 #include "initialization/InitParser.H"
 
+#include <ablastr/parallelization/MPIInitHelpers.H>
+
 #include <AMReX.H>
 #include <AMReX_ParallelDescriptor.H>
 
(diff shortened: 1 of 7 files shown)


src/initialization/InitAMReX.cpp
```diff
@@ -21,28 +23,133 @@
...
+        MPI_Initialized(&is_initialized);
+        if (is_initialized)
+        {
+            // MPI is owned by the user, e.g., by main() or by mpi4py
+            return;
+        }
+
+        ablastr::parallelization::mpi_init(argc, argv);
+
+        std::atexit(finalize_MPI_atexit);
```
ax3l, line 80: Oh boi. Noooo.

Sadly, the MPI standard allows resource leaks and thus mandates, one cannot `MPI_Init()` after `MPI_Finalize()`.

As in #7144, our solution should just be to manage MPI externally of AMReX in `pytest`, i.e., with `mpi4py` having a context alive that we share across pytest functions in the same test process.

PR #1622, opened 2026-08-20
Expanding beam test with acceleration

cemitch99:
This PR adds 2D and 3D tests designed to validate the case of a uniformly-populated beam expanding under its own space charge in the presence of uniform acceleration ($E_z$).

- [x] implement tests
- [x] update documentation
- [x] add to CI

These are two fast tests to help with Issue #1442.

PR #1639, opened 2026-08-31
Envelope Tracking: Strang Splitting

ax3l:
Strang splitting for collective effects SC kicks in envelope tracking.

Follow-up to #1530, which did particle tracking only. `algo.strang_split` now governs both tracking models, so the two always compose their collective kicks to the same order.

Rebased on the merged #1640: the finely sliced envelope element it added to the `expanding-acc` examples is no longer needed, so those two examples share one element again.

<details>

#1530 moved **particle** tracking to a second-order Strang split of the collective kick and the element transport, `M(ds/2) K(ds) M(ds/2)`. Envelope tracking kept the first-order composition `K(ds) M(ds)`, so the two models carried discretization errors of a different order.

That surfaced in the `expanding_beam` acceleration examples added in #1622, which cross-check particle tracking against envelope tracking: `expanding-acc.py.run` and `expanding-2d-acc.py.run` went red on 🐧 OpenMP, 🍏 macOS and 🐧 Tooling once both PRs were on `development`. #1530's own CI runs were cancelled by a superseding push, so it never saw the tests #1622 had just added.

While both sides were first order their errors biased the same way and largely cancelled, which is why the examples passed before #1530. Only the *mismatch* broke them (`sigma_x`, 3D, relative to the envelope value the assertion divides by, 0.79% tolerance):

| | particle | envelope | apart | |
| --- | ---: | ---: | ---: | --- |
| before #1530 | 7.7810e-4 (1st) | 7.8157e-4 (1st) | 0.44% | pass |
| after #1530 | 7.7153e-4 (2nd) | 7.8157e-4 (1st) | **1.28%** | fail |
| #1640 | 7.7153e-4 (2nd) | 7.7528e-4 (1st, nslice=2000) | 0.48% | pass |
| this PR | 7.7153e-4 (2nd) | 7.7495e-4 (2nd) | 0.44% | pass |

The envelope was the under-converged side, not the particles. Both compositions converge to the same limit, the second-order one just gets there far sooner (ChrAcc, `ds = 10 m`, 3D space charge, `sigma_x`):

| nslice | first order | Strang split |
| -----: | ----------: | -----------: |
|    100 | 7.815700e-4 |  7.749494e-4 |
|    200 | 7.782437e-4 |  7.749498e-4 |
|    400 | 7.765928e-4 |  7.749499e-4 |
|    800 | 7.757704e-4 |  7.749499e-4 |
|   1600 | 7.753599e-4 |  7.749499e-4 |
|   6400 | 7.750524e-4 |  7.749499e-4 |

The split reaches its converged value to 7e-7 relative already at `nslice = 100`, and its error drops 4x per doubling. That is what makes #1640's extra slices unnecessary.

### Example reference values

Three envelope examples needed new expected values. The old ones had been tuned to a *compensating* first-order error rather than to physics, and would not have survived a slice refinement under either composition. No tolerance was loosened:

- `analysis_expanding_envelope.py` now asserts the **analytic** solution — this beam expands to exactly twice its initial size over the 6 m drift — which is the same reference the particle version of that example is already checked against. The second-order envelope hits it to 6e-5; the previous reference was 0.95% away from it.
- `analysis_kurth_10nC_periodic_envelope.py` and `analysis_fodo_envelope_sc.py` now assert the converged values of the envelope model. These beams return to within 0.3% / 0.1% of their initial sizes rather than exactly onto them: they are matched to their own self-consistent space charge, which the envelope model represents by the linear field of the equivalent Gaussian. The old references passed only because the first-order error happened to cancel that offset at the slice count the examples use.

### Tests

`tests/python/test_strang_split.py` covered particle tracking only. It gains five envelope cases: second-order convergence, the first-order fallback when `algo.strang_split` is disabled, both compositions converging to the same limit, and the halved-kick path for elements that cannot be sliced. Three of the five fail without the change in this PR.

</details>

### DIFF
--- a/src/tracking/envelope.cpp
+++ b/src/tracking/envelope.cpp
@@ -112,10 +112,15 @@ namespace impactx
                 ablastr::warn_manager::WarnPriority::high
             );
         }
-        if (space_charge == SpaceChargeAlgo::Gauss3D) {
+        // envelope tracking models space charge as the linear field of the rms-equivalent
+        //   uniform beam ellipse (2D) or ellipsoid (3D); the other models are particle-only
+        if (space_charge == SpaceChargeAlgo::Gauss3D ||
+            space_charge == SpaceChargeAlgo::Gauss2p5D ||
+            space_charge == SpaceChargeAlgo::True_2p5D)
+        {
             throw std::runtime_error(
-                "Gauss3D space charge force calculation is only supported with particle tracking. "
-                "For envelope tracking, use: 3D"
+                to_string(space_charge) + " space charge force calculation is only supported "
+                "with particle tracking. For envelope tracking, use: 2D or 3D"
             );
         }
 
@@ -128,9 +133,13 @@ namespace impactx
         AMREX_ALWAYS_ASSERT_WITH_MESSAGE(!csr, "CSR effects are not yet implemented for envelope tracking.");
 
         // whether any collective effect is active: only then is a kick applied per slice
+        //   At zero intensity the kick leaves the covariance matrix unchanged, so it is
+        //   skipped entirely: that is what the warning above announces, and it keeps the
+        //   element transport of such a run unsplit.
         bool const collective_effects =
-            space_charge == SpaceChargeAlgo::True_2D ||
-            space_charge == SpaceChargeAlgo::True_3D;
+            (space_charge == SpaceChargeAlgo::True_2D ||
+             space_charge == SpaceChargeAlgo::True_3D) &&
+            intensity != 0_prt;
 
         // second-order Strang split of the collective kicks, on by default
         //   Disabling it composes kick and transport to first order instead, which is what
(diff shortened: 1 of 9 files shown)


cemitch99 (approved):

src/tracking/envelope.cpp
```diff
@@ -151,6 +203,17 @@ namespace impactx
...
 
+                // zero-length elements receive no kick: it would leave the momenta unchanged
+                bool const kick = collective_effects && slice_ds != amrex::ParticleReal(0);
+
+                // second-order, time-symmetric Strang split of the kick ``K`` and transport ``M``
+                bool const strang = kick && strang_split;
+
+                // which of the two is halved: an element that can be subdivided puts the
+                // transport outside (MKM), any other one halves the kick instead (KMK)
+                bool const split_transport = strang && elements::can_slice(element_variant);
```
ax3l, line 223: This whole chunk here I can deduplicate with the particle tracking loop, but would do so in a follow-up PR to keep this one here focused.

PR #1644, opened 2026-09-01
`PipeAperture`: Make `aperture_x`/`aperture_y` Settable

ax3l:
The `PipeAperture` mixin exposed only getters, so `aperture_x`/`aperture_y` were read-only on every element with a pipe aperture (`drift.aperture_x = 1.0` raised `AttributeError`), unlike the `Alignment` mixin, whose `dx`/`dy`/`rotation` are read/write.
Adds `set_aperture_x()`/`set_aperture_y()` encapsulating the inverted storage, keeping the constructor convention that zero or less removes the restriction, and extends `test_element_serialization.py` to check settable mixin properties across all element types.

- [x] 🤖 Generated with [Claude Code](https://claude.com/claude-code)
- [x] self-reviewed

### DIFF
--- a/src/elements/mixin/pipeaperture.H
+++ b/src/elements/mixin/pipeaperture.H
@@ -37,10 +37,8 @@ namespace impactx::elements::mixin
             amrex::ParticleReal aperture_y
         )
         {
-            using namespace amrex::literals;  // for _prt
-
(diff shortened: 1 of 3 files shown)


egstern (approved):
Looks good. Thanks.

PR #1645, opened 2026-09-01
`Aperture`: Zero `aperture_x`/`aperture_y` Disables the Boundary

ax3l:
The `Aperture` constructor rejected a half-aperture of zero or less, while the Python property setter divided unconditionally, so `ap.aperture_x = 0.0` silently stored an infinite inverse aperture and a `transmit` aperture then lost the whole beam.

The `PipeAperture` mixin already treats zero or less as "no aperture restriction in that plane". This uses and documents the same convention for the standalone `Aperture`. `set_aperture_x()`/`set_aperture_y()` carry it along with the inverted storage, the matching getters read a disabled plane back as zero, and the constructor and the bindings go through them. A rectangular aperture with only `aperture_y` set is thus a horizontal zero-width slit (an elliptical one degenerates into the same slab), for `action="absorb"` the absorbing domain is unbounded in that direction instead, which the docs call out.

Behavior change: the constructor used to raise on a half-aperture of zero or less and now disables that plane.

Adds `tests/python/test_aperture_disabled.py`, which tracks a beam through the disabled, half-disabled and absorbing cases — a get/set roundtrip alone cannot cover the setter bug, since `1/inf` reads back as zero just like a disabled plane does. Extends `test_element_serialization.py` with that property roundtrip.

Follow-up to #1644.

### DIFF
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -341,6 +341,11 @@ This requires these additional parameters:
 
     Horizontal / vertical half-aperture (elliptical or rectangular).
 
+    A value of zero or less disables the boundary in that plane, so a rectangular
+    aperture with only ``aperture_y`` set acts as a horizontal slit.
+    Note that with ``action = absorb`` a disabled plane makes the absorbing domain
+    unbounded in that direction.
+
 .. pp:param:: <aperture_name>.repeat_x/y
     :link_aliases: <aperture_name>.repeat_x <aperture_name>.repeat_y
     :type: ``float``
--- a/src/elements/Aperture.H
+++ b/src/elements/Aperture.H
@@ -62,10 +62,18 @@ namespace impactx::elements
         /** A thin collimator element that applies a transverse aperture boundary.
          *  Particles outside the boundary are considered lost.
          *
+         *  A half-aperture of zero or less disables the boundary in that plane,
+         *  the same convention as \see mixin::PipeAperture. A rectangular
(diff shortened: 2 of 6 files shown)


src/elements/Aperture.H
```diff
@@ -270,6 +261,66 @@ namespace impactx::elements
...
+         * @throws std::invalid_argument if aperture_x is zero or less
+         */
+        AMREX_GPU_HOST AMREX_FORCE_INLINE
+        void set_aperture_x (amrex::ParticleReal aperture_x)
+        {
+            using namespace amrex::literals; // for _prt
+
+            if (aperture_x <= 0_prt) {
+                throw std::invalid_argument("Aperture: aperture_x must be > 0!");
+            }
```
ax3l: The `PipeAperture` element mixin class (`aperture_x` and `aperture_y` arguments) treats 0 as a disabled aperture. Using and documenting the same convention here would be consistent.

docs/source/usage/parameters.rst
```diff
@@ -341,6 +341,11 @@ This requires these additional parameters:
 
     Horizontal / vertical half-aperture (elliptical or rectangular).
 
+    A value of zero or less disables the boundary in that plane, so a rectangular
+    aperture with only ``aperture_y`` set acts as a horizontal slit.
```
ax3l, line 345:
```suggestion
    aperture with only ``aperture_y`` set acts as a "horizontal slit with zero width".
```

ax3l:
Will merge this fix into #1652.

PR #1648, opened 2026-09-02
Thin Elements: Transverse Beam Pipe Aperture

ax3l:
Every thick element except `Programmable` carries the `PipeAperture` mixin, but no _thin_ one did.

This adds `aperture_x`/`aperture_y` to `Buncher`, `DipEdge`, `Kicker`, `LinearMap`, `Multipole`, `NonlinearLens`, `QuadEdge`, `ShortRF`, `TaperedPL` and `ThinDipole`, in Python, the inputs file and the docs.

The coordinate-transformation and bookkeeping elements `PlaneXYRot`, `PRot` and `SpinMap` deliberately do not get an aperture: they represent no physical magnet or cavity, and `SpinMap` does not touch the phase space at all.

Adds `tests/python/test_thin_element_aperture.py`, which checks that a tight aperture scrapes and that the default of zero does not.

### DIFF
--- a/tests/python/test_thin_element_aperture.py
+++ b/tests/python/test_thin_element_aperture.py
@@ -15,7 +15,7 @@ aperture, just like the finite-length (thick) elements do.
 import numpy as np
 import pytest
 
-from impactx import ImpactX, Map3x6, Map6x6, Vector3, distribution, elements
+from impactx import ImpactX, Map6x6, distribution, elements
 
 # beam sizes, chosen so that a 1 sigma aperture cuts a sizable fraction
 LAMBDA_X = 1.0e-3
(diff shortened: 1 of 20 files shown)


tests/python/test_thin_element_aperture.py
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022-2026 The ImpactX Community
+#
+# Authors: Axel Huebl
```
ax3l:
```suggestion
# Authors: Axel Huebl, Chad Mitchell
```

ax3l:
The performance regression is expected, since we have to load & store the idcpu now. Still better to fuse than as separate element for those quick elements.

We can optimize a bit, e.g., only conditionally writing and not also reading, and recover some in a follow up PR like #1649

cemitch99 (commented):
Several of these elements, including PlaneXYRot, PRot, and (depending on interpretation) LinearMap and SpinMap are "bookkeeping" elements that do not represent a physical magnet or cavity.  It's not clear to me that we should support aperture restrictions in these elements.  For example, PlaneXYRot is just a rotation of coordinates.  (Also, in cases like PlaneXYRot, there is ambiguity about the meaning of the x and y half-width values, unless these value are equal.)

ax3l:
Hm, good point. I intentionally added `LinearMap` because it could be a user-configured linear element that still could have an aperture associated with them.

For `SpinMap` we skip aperture support indeed, since it does not change phase space. We generally apply `PipeAperture` after the element push, using the same shift as the `Alignment` frame applies to the element.

For the momentum-kick thin elements (ShortRF, Buncher, Kicker, Multipole,  ThinDipole, NonlinearLens, TaperedPL) the aperture could technically be applied only once (_if_ the aperture is the same) when we actually change positions in another element (e.g. Drift), but supporting this everywhere for convenience, safety and composability is probably totally ok as well.

I agree that the transformation & placeholder types `PlaneXYRot`, `PRot`, `Marker` could def skip aperture support.

cemitch99 (approved):
Thanks!

PR #1649, opened 2026-09-02
Guard the idcpu Write-Back on an Active Aperture

ax3l:
In SIMD builds the idcpu write-back is gated only at compile time on `has_pipe_aperture_v`, so every element with the `PipeAperture` mixin stores idcpu per particle even when no aperture is set — the only runtime guard sits inside `apply_aperture` and merely skips the mask.
This adds `PipeAperture::aperture_active()` and a `store_idcpu()` helper: unconditional store when the push method writes idcpu through a non-const argument, otherwise conditional on an aperture actually being set (which should also let the compiler sink the load into the branch).

## Benchmarks

Median wall time per `Drift.push` (lower is better), comparing the PR head (`e783948`) with its parent (`236d384`):

| Backend | Case | Parent (ms/push) | PR (ms/push) | Latency change |
|---|---|---:|---:|---:|
| SIMD (4-wide) | no spin, inactive aperture | 9.539 | 7.953 | **-16.62%** |
| SIMD (4-wide) | spin, inactive aperture | 9.499 | 7.959 | **-16.22%** |
| SIMD (4-wide) | no spin, active aperture | 10.393 | 10.425 | +0.31% |
| SIMD (4-wide) | spin, active aperture | 10.475 | 10.469 | -0.05% |
| Non-SIMD | no spin, inactive aperture | 8.004 | 8.038 | +0.42% |
| Non-SIMD | spin, inactive aperture | 8.002 | 8.031 | +0.36% |
| Non-SIMD | no spin, active aperture | 8.323 | 8.295 | -0.34% |
| Non-SIMD | spin, active aperture | 8.314 | 8.370 | +0.67% |
| CUDA | no spin, inactive aperture | 2.1739 | 2.1729 | -0.048% |
| CUDA | spin, inactive aperture | 2.1745 | 2.1745 | -0.002% |
| CUDA | no spin, active aperture | 2.1744 | 2.1742 | -0.008% |
| CUDA | spin, active aperture | 2.1751 | 2.1751 | -0.003% |

The inactive-aperture SIMD path is 16.2–16.6% faster (19–20% higher throughput). Active-aperture SIMD, non-SIMD, and CUDA results are unchanged within measurement noise.

Test setup: 5,000,000 particles in double-precision Release builds. CPU measurements used one pinned Intel i9-12900H P-core with one OpenMP thread; CPU samples contained 20 pushes after five warm-up pushes. CUDA measurements used an NVIDIA RTX A2000 8 GB (compute capability 8.6), explicit device synchronization around every sample, 30 warm-up pushes, and 100 pushes per sample. Each result is the median of 9 CPU or 11 CUDA samples; CPU PR values average two stable runs.

### DIFF
--- a/src/elements/mixin/beamoptic.H
+++ b/src/elements/mixin/beamoptic.H
@@ -179,6 +179,50 @@ namespace detail
         amrex::ignore_unused(val, ptr, i);
     }
 
+    /** Store the particle index back to memory, if anything can have changed it
+     *
+     * Unlike the phase space slots, idcpu is only ever modified when a particle
+     * is lost. If the push method itself does not write it, the only writer left
+     * is \see PipeAperture::apply_aperture - which is a no-op unless an aperture
+     * is actually set on this element. That is a runtime property, so guard the
+     * write-back with it: an unset aperture then costs neither the store nor,
+     * once the compiler sinks it into the branch, the load.
+     *
+     * @tparam P_Method the push method of the element, to inspect argument constness
+     * @tparam N the argument position of idcpu in the push method
+     * @tparam T_HasAperture whether the element carries the \see PipeAperture mixin
+     * @tparam T_Element This can be a \see Drift, \see Quad, \see Sbend, etc.
+     * @tparam IndexType int or SIMD index
+     * @tparam ValType the type of the value to store
+     * @param val the value to store
+     * @param ptr pointer to the SoA data
+     * @param i index or SIMD index
+     * @param element the beamline element that was pushed through
+     */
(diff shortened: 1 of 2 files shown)


ax3l:
This is good on its own merit.

As a follow-up PR, I think we can be even better by not even reading idcpu: we can return a Boolean (or vector bool) from a kernel if a particle survives and if it does not, write conditionally back...? But could be tricky just to flip the one but for invalid, given that we have to move memory banks... Well, the whole read-modify-write could be conditional then.

cemitch99 (approved):

PR #1652, opened 2026-09-03
`PipeAperture`: Bound Each Plane Independently

ax3l:
For our mixin class per element `PipeAperture`, setting only one of `aperture_x`/`aperture_y` disabled the beam pipe entirely.
This silently discarded the half-aperture that was actually set, e.g.
```py
Drift(ds=0.1, aperture_x=1e-4, aperture_y=0)
```
lost no particles at all.

With this change, each plane is bounded independently, so both collimator types are expressible:

* a **jaw** (slit) collimator bounds a single plane - the example above is now infinitely open in y and bounded at ±`1e-4` in x,
* an **iris** (the usual beam pipe) bounds both planes, as before.

For a single bounded plane the `rectangular` and `elliptical` shapes degenerate to the same slab. `PolygonAperture` is unaffected, as it is defined by vertices rather than half-apertures.

This is a behavior change for elements setting exactly one plane, from "no aperture" to "the plane you set cuts". Nothing in the examples or tests sets a single plane, and the old convention was documented nowhere.

## Also folds in #1645

This supersedes #1645, so that the `Aperture` collimator and the `PipeAperture` beam pipe land on the same convention in one go and the docs can state it once:

* `Aperture` no longer throws on a non-positive half-aperture, it disables that plane, matching `PipeAperture`.
* Small Python fix: the `aperture_x`/`aperture_y` properties and `to_dict()` went through a raw `1 / m_inv_aperture_*`, which divides by zero once a plane is disabled. They now use the `aperture_x()`/`aperture_y()` accessors and report `0` for a disabled plane, so `to_dict()`/`from_dicts()` round-trips again.

Verified that the two now agree exactly: a beam pipe, a `rectangular` collimator and an `elliptical` collimator with `aperture_x=1e-4, aperture_y=0` all transmit the same 805 of 10000 particles (the expected Gaussian slab fraction of ~8%).

## Docs

The convention is stated once for the inputs file and once for the Python API, covering the collimator and the beam pipe together. Only what is specific to the collimator stays in its own block: how a disabled plane interacts with `action = absorb`.

## Tests

* `test_single_plane_aperture_still_cuts` in `tests/python/test_thin_element_aperture.py`, parametrized over plane x {thin, thick}: the bounded plane cuts while the other stays unbounded.
* `tests/python/test_aperture_disabled.py` and the serialization coverage from #1645.

### DIFF
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -347,6 +347,11 @@ This requires these additional parameters:
 
     Horizontal / vertical half-aperture (elliptical or rectangular).
 
+    A value of zero or less disables the boundary in that plane, so a rectangular
+    aperture with only ``aperture_y`` set acts as a horizontal slit.
+    Note that with ``action = absorb`` a disabled plane makes the absorbing domain
+    unbounded in that direction.
+
 .. pp:param:: <aperture_name>.repeat_x/y
     :link_aliases: <aperture_name>.repeat_x <aperture_name>.repeat_y
     :type: ``float``
--- a/docs/source/usage/parameters.rst
+++ b/docs/source/usage/parameters.rst
@@ -326,11 +326,14 @@ Lattice Elements
 
     Indicates the element type for this lattice element. This should be one of the following.
 
-Many elements accept a transverse beam pipe aperture via ``aperture_x`` and ``aperture_y``.
-Each plane is bounded independently: a half-aperture of zero or less (the default) removes the
-constraint in that plane only, while the other plane still cuts.
-Setting only ``aperture_y``, for instance, loses a particle when ``|y| > aperture_y`` at any ``x``.
-The aperture is disabled entirely only if both planes are zero or less.
+Many elements take a transverse aperture via ``aperture_x`` and ``aperture_y``: the ``aperture``
+collimator, and as a beam pipe most other elements.
+They all follow the same convention. Each plane is bounded independently: a half-aperture of zero
+or less removes the constraint in that plane only, while the other plane still cuts.
+Bounding a single plane gives a jaw (slit) collimator, bounding both an iris; with only
+``aperture_y`` set, a particle is lost when ``|y| > aperture_y`` at any ``x``, and the
+``rectangular`` and ``elliptical`` shapes degenerate to the same slab.
+The aperture is disabled entirely only if both planes are zero or less, which is the default.
 
 
 ``aperture``
@@ -347,10 +350,9 @@ This requires these additional parameters:
 
     Horizontal / vertical half-aperture (elliptical or rectangular).
 
-    A value of zero or less disables the boundary in that plane, so a rectangular
-    aperture with only ``aperture_y`` set acts as a horizontal slit.
     Note that with ``action = absorb`` a disabled plane makes the absorbing domain
-    unbounded in that direction.
+    unbounded in that direction, so zeroing both planes absorbs the whole beam,
+    while for ``action = transmit`` the element becomes a no-op.
 
 .. pp:param:: <aperture_name>.repeat_x/y
     :link_aliases: <aperture_name>.repeat_x <aperture_name>.repeat_y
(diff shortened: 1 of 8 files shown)


docs/source/usage/parameters.rst
```diff
@@ -341,6 +350,10 @@ This requires these additional parameters:
 
     Horizontal / vertical half-aperture (elliptical or rectangular).
 
+    Note that with ``action = absorb`` a disabled plane makes the absorbing domain
+    unbounded in that direction, so zeroing both planes absorbs the whole beam,
+    while for ``action = transmit`` the element becomes a no-op.
```
```suggestion
    while for ``action = transmit`` the element has no effect on the beam.
```

docs/source/usage/python.rst
```diff
@@ -2473,12 +2482,18 @@ For an element with ``nslice`` > 1, the pushes and maps refer to a single ``ds/n
...
 
    A thin collimator element, applying a transverse aperture boundary.
 
-   :param aperture_x: horizontal half-aperture (rectangular or elliptical) in m
-   :param aperture_y: vertical half-aperture (rectangular or elliptical) in m
+   A half-aperture of zero or less disables the boundary in that plane, the same
+   convention as every other element's ``aperture_x``/``aperture_y`` above.
+   Note that with ``action="absorb"`` a disabled plane makes the absorbing domain
+   unbounded in that direction, so disabling both planes absorbs the whole beam,
+   while for ``action="transmit"`` the element becomes a no-op.
```
```suggestion
   while for ``action="transmit"`` the element has no effect on the beam.
```

src/elements/Aperture.H
```diff
@@ -62,10 +62,20 @@ namespace impactx::elements
...
          *
+         *  A half-aperture of zero or less disables the boundary in that plane,
+         *  the same convention as \see mixin::PipeAperture: each plane is bounded
+         *  independently, so bounding a single plane gives a jaw (slit) collimator
+         *  and bounding both an iris. With only ``aperture_y`` set, a particle is
+         *  lost when ``|y| > aperture_y`` at any ``x``, and the rectangular and
+         *  elliptical shapes degenerate to the same slab. Note that with
+         *  ``action = absorb`` a disabled plane makes the absorbing domain
+         *  unbounded in that direction: disabling both planes absorbs the whole
+         *  beam, while for ``action = transmit`` the element becomes a no-op.
```
ax3l:
```suggestion
         *  beam, while for ``action = transmit`` the element has no effect on the beam.
```

src/elements/mixin/pipeaperture.H
```diff
@@ -85,16 +95,22 @@ namespace impactx::elements::mixin
...
+         * loop, so an element without an aperture pays nothing per particle.
          *
          * @return true if particles can be lost in this aperture
          */
         AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
         bool aperture_active () const
         {
             using namespace amrex::literals; // for _prt
-            return m_inv_aperture_x > 0_prt && m_inv_aperture_y > 0_prt;
+            return m_inv_aperture_x + m_inv_aperture_y > 0_prt;
```
ax3l, line 113: This is the only actual code change of the `PipeAperture` in this PR.

I also tried `&&` -> `||` but that interacts negatively on CPU with the cost heuristics in GCC and makes elements more costly 😅 The `+` formulation is equivalent and does not show the issue.

src/python/elements.cpp
```diff
@@ -643,14 +642,14 @@ void init_elements(py::module& m)
...
+            [](Aperture & ap) { return ap.aperture_x(); },
+            [](Aperture & ap, amrex::ParticleReal aperture_x) { ap.set_aperture_x(aperture_x); },
+            "maximum horizontal coordinate in m; zero or less removes the horizontal boundary"
         )
         .def_property("aperture_y",
-            [](Aperture & ap) { using namespace amrex::literals; return 1_prt / ap.m_inv_aperture_y; },
-            [](Aperture & ap, amrex::ParticleReal aperture_y) { using namespace amrex::literals; ap.m_inv_aperture_y = 1_prt / aperture_y; },
-            "maximum vertical coordinate"
+            [](Aperture & ap) { return ap.aperture_y(); },
+            [](Aperture & ap, amrex::ParticleReal aperture_y) { ap.set_aperture_y(aperture_y); },
```
ax3l, line 651: This is the 2nd actual code change in this PR: `Aperture` setter and getter are more safe now if user provides <= 0 after construction it will handle it gracefully.

cemitch99 (approved):

Issue #1663, opened 2026-09-16
ExactCFbend linear map return yields nan


cemitch99:
The function returning the linear map for the element `ExactCFbend` currently yields nan for the test case attached.  This does not affect particle tracking, which occurs independently of this evaluation.  However, it does affect envelope tracking (which exploits the linear map return) and calls to `lattice.transfer_map()`.  

[run_cfbend_test.py](https://github.com/user-attachments/files/32302207/run_cfbend_test.py)

Note that the normal and skew quad coefficients both vanish in this example.  _Setting the normal quad coefficient to a nonzero value resolves this issue_.  This suggests that the degenerate edge case with vanishing normal quad coefficient requires special treatment.

PR #1664, opened 2026-09-16
Fix linear map return for ExactCFbend (zero quad)

cemitch99:
This PR addresses Issue #1663 .  The linear map return for the `ExactCFbend` is modified to correctly handle the edge case when either the normal or skew quad coefficients vanishes.

A test compares the linear map obtained with both quad coefficients vanishing against the corresponding linear map for an `Sbend`.

This PR affects only envelope tracking and calls to `lattice.transfer_map()`.  The issue is not relevant for particle tracking.


tests/python/test_cfbend_zero_quad.py
```diff
@@ -0,0 +1,83 @@
...
+    sim.space_charge = False
+    sim.slice_step_diagnostics = False
+
+    # domain decomposition & space charge mesh
+    sim.init_grids()
+
+    #  set reference particle
+    ref = sim.beam.ref
+    kin_energy_MeV = 2.606299137493995  # reference kinetic energy (p = 70 MeV/c)
+    charge_qe = 1.0  # particle charge
```
cemitch99:
```suggestion
```

tests/python/test_cfbend_zero_quad.py
```diff
@@ -0,0 +1,83 @@
...
+        aperture_y=0.0,
+        ds=0.391140372489,
+        dx=0.0,
+        dy=0.0,
+        int_order=2,
+        k_normal=[1.3386467172031062, 0.0, 30.21694, -75.81527],
+        k_skew=[0.0, 0.0, 0.0, 0.0],
+        mapsteps=10,
+        name="m1r",
+        nslice=10,
```
cemitch99:
```suggestion
        nslice=ns,
```

ax3l:
Hi @cemitch99,

It looks like `CFbend` with `k=0` has the same bug, but there it also affects particle tracking:
 - The map returns NaN at `R(3,4)` (`CFbend.H:393`).
 - `compute_constants` has the same `0/0` (`CFbend.H:127`), so tracking one particle gives NaN in y and py.

It's the same two-line fix, so I'd include it here and extend the test to cover both elements.

src/elements/ExactCFbend.H
```diff
@@ -697,34 +697,36 @@ namespace impactx::elements
...
             auto const [sinx, cosx] = amrex::Math::sincos(omega_x * slice_ds);
-            amrex::ParticleReal const sinhx = std::sinh(omega_x * slice_ds);
+            amrex::ParticleReal const sincx = (omega_x != 0.0)? sinx/omega_x : slice_ds;
+            amrex::ParticleReal const sinhcx = (omega_x != 0.0)? std::sinh(omega_x * slice_ds)/omega_x : slice_ds;
             amrex::ParticleReal const coshx = std::cosh(omega_x * slice_ds);
             auto const [siny, cosy] = amrex::Math::sincos(omega_y * slice_ds);
-            amrex::ParticleReal const sinhy = std::sinh(omega_y * slice_ds);
+            amrex::ParticleReal const sincy = (omega_y != 0.0)? siny/omega_y : slice_ds;
+            amrex::ParticleReal const sinhcy = (omega_y != 0.0)? std::sinh(omega_y * slice_ds)/omega_y : slice_ds;
             amrex::ParticleReal const coshy = std::cosh(omega_y * slice_ds);
```
ax3l, line 706: Claude:

Style: lines 701 and 705 are 114 characters (the C++ limit is 100), and the rest of the element uses `0_prt` literals and `cond ? a : b` spacing.

This suggestion keeps `sinhx`/`sinhy` and derives the guarded quotients from them. The `R(1,2)`…`R(4,3)` suggestion further down uses `sinhx`/`sinhy`, so it needs this one applied too.

```suggestion
            // trigonometry
            auto const [sinx, cosx] = amrex::Math::sincos(omega_x * slice_ds);
            amrex::ParticleReal const sinhx = std::sinh(omega_x * slice_ds);
            amrex::ParticleReal const coshx = std::cosh(omega_x * slice_ds);
            auto const [siny, cosy] = amrex::Math::sincos(omega_y * slice_ds);
            amrex::ParticleReal const sinhy = std::sinh(omega_y * slice_ds);
            amrex::ParticleReal const coshy = std::cosh(omega_y * slice_ds);

            // sin(w ds)/w and sinh(w ds)/w, which tend to ds for a vanishing w
            amrex::ParticleReal const sincx = omega_x != 0_prt ? sinx / omega_x : slice_ds;
            amrex::ParticleReal const sinhcx = omega_x != 0_prt ? sinhx / omega_x : slice_ds;
            amrex::ParticleReal const sincy = omega_y != 0_prt ? siny / omega_y : slice_ds;
            amrex::ParticleReal const sinhcy = omega_y != 0_prt ? sinhy / omega_y : slice_ds;
```

src/elements/ExactCFbend.H
```diff
@@ -697,34 +697,36 @@ namespace impactx::elements
...
 
             amrex::ParticleReal const igbrc = 1_prt / (gx * bet * rc);
-            amrex::ParticleReal const iobrc = 1_prt / (omega_x * bet * rc);
-            amrex::ParticleReal const igobr = 1_prt / (gx * omega_x * b2rc2);
+            amrex::ParticleReal const iobrc = 1_prt / (bet * rc);
+            amrex::ParticleReal const igobr = 1_prt / (gx * b2rc2);
```
ax3l, line 710: Claude:

The x-plane guards (`sincx`/`sinhcx`) don't take effect yet: `omega_x == 0` only when `gx == 0`, and then `igbrc` and `igobr` divide by zero, so `R(1,6)`, `R(5,2)` and `R(5,6)` are still NaN.

This is reachable with a gradient bend of zero horizontal focusing, `k_normal=[h, -h*h]` (2.606 MeV proton, `ds=0.391`, `nslice=1`, compared with `expm` of the linearized equations):
- rc = 1.7, 3.3, 12.5 m: `gx` is exactly 0, the three entries are NaN.
- rc = 0.747 m: `gx` ends up ~7e-16, `1 - cos` rounds to 0, so `R(1,6)` and `R(5,2)` come out 0 instead of ∓1.377 and `R(5,6)` is 4.8% off.
- Near the cancellation accuracy degrades: 1e-6 relative error at gx/h² ≈ 1e-10, 1.7% at 1e-14.

`CFbend::transport_map` has the same expressions. A fix needs the small-argument limits (1 − cos(ω ds))/gx → ds²/2 and (sin(ω ds)/ω − ds)/gx → −ds³/6, e.g. a series branch when |gx|·ds² is small. This predates the PR, so a follow-up issue would be fine.

ax3l (commented):
Claude:

Checked locally (DP/OMP/MPI build of 3378501): the new map agrees with `expm` of the linearized combined-function-bend equations to ~3e-16 relative for kn = 0, kn > 0, −h² < kn < 0, the gx < 0 branch, `unit=1` and `ds < 0`, so the refactor leaves the non-degenerate branches unchanged. With `k_normal=[h, 0, k2, k3]` it also matches a finite-difference Jacobian of particle tracking to 4e-8. The ExactCFbend/CFbend reversibility tests and `test_lattice_optics.py` pass.

On the description: the linear map never reads `k_skew`, so the degeneracy is only in the normal quad coefficient; and "not relevant for particle tracking" holds for `ExactCFbend` but not for `CFbend` (see inline).

src/elements/ExactCFbend.H
```diff
@@ -697,34 +697,36 @@ namespace impactx::elements
...
 
             // initialize linear map matrix elements
             Map6x6 R = Map6x6::Identity();
```
ax3l, line 713: Claude:

Also predates this PR, but relevant to the description ("normal or skew quad coefficients"): `transport_map` never reads `k_skew`, so the linear map, and with it envelope tracking and `transfer_map`, drops skew components entirely.

For `k_normal=[1.3386, 0.0]`, `k_skew=[0.0, 0.8]`, a finite-difference Jacobian of particle tracking (`int_order=6`, `mapsteps=200`) shows x–y coupling, e.g. `R(1,3) ≈ 0.060`, `R(2,3) ≈ 0.299`, `R(4,1) ≈ 0.299`, where `transfer_map` has 0 (max deviation 0.30). With `k_skew` all zero, the same Jacobian matches `transfer_map` to 4e-8.

The MAD-X importer sends bends with `K1S` to `ExactCFbend` (`madx_to_impactx.py:574`), so envelope tracking of such lattices is silently uncoupled. Worth a follow-up issue, or at least a note in the docs.

src/elements/ExactCFbend.H
```diff
@@ -697,34 +697,36 @@ namespace impactx::elements
...
-            R(2,1) = gx > 0_prt ? -omega_x * sinx : omega_x * sinhx;
+            R(2,1) = gx > 0_prt ? -omega_x * sinx : omega_x * omega_x * sinhcx;
             R(2,2) = gx > 0_prt ? cosx : coshx;
-            R(2,6) = gx > 0_prt ? -sinx * iobrc : -sinhx * iobrc;
+            R(2,6) = gx > 0_prt ? -sincx * iobrc : -sinhcx * iobrc;
             R(3,3) = gy > 0_prt ? cosy : coshy;
-            R(3,4) = gy > 0_prt ? siny / omega_y : sinhy / omega_y;
-            R(4,3) = gy > 0_prt ? -omega_y * siny : omega_y * sinhy;
+            R(3,4) = gy > 0_prt ? sincy : sinhcy ;
+            R(4,3) = gy > 0_prt ? -omega_y * siny : omega_y * omega_y * sinhcy;
```
ax3l, line 723: Claude:

`R(2,1)` and `R(4,3)` didn't need rewriting: the original `omega_x * sinhx` is already `0 * sinh(0) = 0` at `omega_x == 0`, so `omega_x * omega_x * sinhcx` only changes rounding in the hyperbolic branch for every other strength. This restores the original expressions (requires the `sinhx`/`sinhy` suggestion above) and drops the space before `;` in `R(1,2)` and `R(3,4)`.

```suggestion
            R(1,2) = gx > 0_prt ? sincx : sinhcx;
            R(1,6) = gx > 0_prt ? -(1_prt - cosx) * igbrc : -(1_prt - coshx) * igbrc;
            R(2,1) = gx > 0_prt ? -omega_x * sinx : omega_x * sinhx;
            R(2,2) = gx > 0_prt ? cosx : coshx;
            R(2,6) = gx > 0_prt ? -sincx * iobrc : -sinhcx * iobrc;
            R(3,3) = gy > 0_prt ? cosy : coshy;
            R(3,4) = gy > 0_prt ? sincy : sinhcy;
            R(4,3) = gy > 0_prt ? -omega_y * siny : omega_y * sinhy;
```

tests/python/test_cfbend_zero_quad.py
```diff
@@ -0,0 +1,82 @@
...
+    # return the linear map
+    R_cfbend = sim.lattice.transfer_map(ref).to_numpy()
+    np.savetxt("matrix_cfbend.txt", R_cfbend, delimiter=" ")
+
+    sim.lattice.clear()
+    sim.lattice.extend([sbend])
+
+    # return the linear map
+    R_sbend = sim.lattice.transfer_map(ref).to_numpy()
+    np.savetxt("matrix_sbend.txt", R_cfbend, delimiter=" ")
```
ax3l, line 63: Claude:

Copy-paste bug: this saves `R_cfbend` again, so `matrix_sbend.txt` comes out byte-identical to `matrix_cfbend.txt`. `conftest.py` runs each test in a temporary directory, so the files are discarded anyway; I'd drop both `np.savetxt` calls.

tests/python/test_cfbend_zero_quad.py
```diff
@@ -0,0 +1,82 @@
...
+    print()
+    print("  Linear map for sbend:  ")
+    print(R_sbend)
+
+    assert np.allclose(
+        R_cfbend,
+        R_sbend,
+        rtol=rtol,
+        atol=atol,
+    )
```
ax3l, line 79: Claude:

This assertion fails in CI on *GCC w/ SIMD w/o MPI w/ Python* (built with `-march=x86-64-v3`), although the printed matrices look identical. `atol=1e-14` is absolute while `R(5,6) ≈ 67`; in my local DP build the max deviation is already 8.0e-15. It would also fail in single precision, since unlike the neighboring tests it doesn't adjust tolerances for `Config.precision`.

`tests/python/test_lattice_optics.py` already has `test_transport_map_zero_strength` for quads and a `_tolerances()` helper, and a `RefPart` alone is enough (no `ImpactX()`, `init_grids()` or `finalize()`). Suggest moving the test there:

```python
def test_cfbend_transport_map_zero_quad():
    """A combined-function bend with zero quadrupole strength must transport like an Sbend."""
    ref = RefPart()
    ref.set_species("proton").set_kin_energy_MeV(2.606299137493995)
    rtol, atol = _tolerances()
    rc, ds = 0.747023084693581, 0.391140372489
    el = elements.ExactCFbend(
        ds=ds, k_normal=[1.0 / rc, 0.0, 30.21694, -75.81527], k_skew=[0.0] * 4
    )
    R = el.transfer_map(ref).to_numpy()
    R_sbend = elements.Sbend(ds=ds, rc=rc).transfer_map(ref).to_numpy()
    assert np.all(np.isfinite(R))
    assert np.allclose(R, R_sbend, rtol=rtol, atol=atol)
```

This passes locally (DP), also with `nslice=10` through a lattice. It can be parametrized over `CFbend(ds=ds, rc=rc, k=0.0)` once that element gets the same guard.

tests/python/test_cfbend_zero_quad.py
ax3l:
```suggestion
    np.savetxt("matrix_sbend.txt", R_sbend, delimiter=" ")
```

src/elements/ExactCFbend.H
cemitch99: @ax3l There is a suggested change here, but it cannot be applied.

tests/python/test_cfbend_zero_quad.py
```diff
@@ -0,0 +1,82 @@
...
+        rotation=0.0,
+        unit=0,
+    )
+    sbend = elements.Sbend(ds=0.391140372489, rc=0.747023084693581)
+
+    sim.lattice.extend([cfbend])
+
+    # return the linear map
+    R_cfbend = sim.lattice.transfer_map(ref).to_numpy()
+    np.savetxt("matrix_cfbend.txt", R_cfbend, delimiter=" ")
```
cemitch99, line 56:
```suggestion
```

tests/python/test_cfbend_zero_quad.py
cemitch99, line 63:
```suggestion
```

src/elements/ExactCFbend.H
ax3l: Ah, it's a bit flaky in the GUI. Can apply manually if you like, should read:
```C++
            // trigonometry
            auto const [sinx, cosx] = amrex::Math::sincos(omega_x * slice_ds);
            amrex::ParticleReal const sinhx = std::sinh(omega_x * slice_ds);
            amrex::ParticleReal const coshx = std::cosh(omega_x * slice_ds);
            auto const [siny, cosy] = amrex::Math::sincos(omega_y * slice_ds);
            amrex::ParticleReal const sinhy = std::sinh(omega_y * slice_ds);
            amrex::ParticleReal const coshy = std::cosh(omega_y * slice_ds);
            // sin(w ds)/w and sinh(w ds)/w, which tend to ds for a vanishing w
            amrex::ParticleReal const sincx = omega_x != 0_prt ? sinx / omega_x : slice_ds;
            amrex::ParticleReal const sinhcx = omega_x != 0_prt ? sinhx / omega_x : slice_ds;
            amrex::ParticleReal const sincy = omega_y != 0_prt ? siny / omega_y : slice_ds;
            amrex::ParticleReal const sinhcy = omega_y != 0_prt ? sinhy / omega_y : slice_ds;
```

tests/python/test_cfbend_zero_quad.py
ax3l: removed and moved into `test_lattice_optics.py`

tests/python/test_cfbend_zero_quad.py
ax3l: removed and moved into `test_lattice_optics.py`

PR #1665, opened 2026-09-17
Doc: Perlmutter (NERSC) HDF5 & Py

ax3l:
Modules went missing, upgrading to new defaults for parallel HDF5 & Python.

Note: This also needs a re-install of our dependencies with the provided script, because the base modules changed.


kngott:
Looks correct to me.

kngott (approved):

